PR #48, opened 2024-06-04
fix: issue where docker compose was not providing DB creds to api server

### LINKED ISSUE #47: docker compose doesnt pass DB creds to backend api
The API recently made a major change to how DB credentials are pulled from the environment and docker compose was never updated to reflect this which has lead to `docker compose up` returning errors from the api

talentedmrjones:
Just needed to refactor credential delivery in compose.env
(only compose.env-example is included in the repo. compose.env is ignored for security reasons)
fixes #47


ATNoblis (approved):
:shipit:

PR #52, opened 2024-06-07
chore: cleanup unnecessary files

### LINKED ISSUE #51: chore: original scripts can be archived
A new branch was created `original-script-archive` as a way to keep the original JS and python scripts. These can now be deleted from main

talentedmrjones:
The scripts are archived in the `original-script-archive` branch and are not needed on main. 
`www/` is no longer necessary because we have all frontend code in `ui/`
closes #51


talentedmrjones:
CodeQL python check is only failing because the python script was deleted. That check is now no longer required, and should not appear again in future PRs

ATNoblis (approved):
:shipit:

PR #57, opened 2024-06-25
feat: added tables and basic user management functionality

### LINKED ISSUE #56: add table and upsert in api for user management
Design a data model that supports the storage of personas and user journeys

try it simple first: create users table to include email, job code, and use an array for the fisma systems the user has access to

When a user makes any request, because authentication has already happened with verified access, we should upsert to create or update with last request time and job code

talentedmrjones:
closes #56


ATNoblis (approved):
:shipit:

PR #65, opened 2024-07-15
feat: added /whoami as endpoint to return encoded jwt of logged in user

### LINKED ISSUE #61: feat: api should provide endpoint for retreiving user profile
`/<url tbd>` should return a JWT that represents current logged in user

This will essentially proxy the encoded JWT as provided by verified access header `x-amzn-ava-user-context`

also provided an endpoint to get the validation key

talentedmrjones:
closes #61


ATNoblis (approved):
:shipit:

PR #67, opened 2024-07-16
feat: added users type to graph and related queries

### LINKED ISSUE #66: api provides queries to list users and get user by id
Add users type to graphql schema and provide associated resolvers, controllers, and model methods

talentedmrjones:
closes #66


ATNoblis (approved):
:shipit:

PR #74, opened 2024-07-22
refactor: removed outputs

talentedmrjones:
Outputs are not used programmatically and only serve to expose information in logs


ATNoblis (approved):
:shipit:

PR #86, opened 2024-07-25
refactor: response struct now has methods for error and response codes

### LINKED ISSUE #76: response objects should have methods for error and status codes
Currently the graph resolvers are each testing for error and setting Code and Message fields which leads to a lot of repetitious code. To make this more DRY move the error testing and code setting to methods on the response struct

talentedmrjones:
We love DRY code here.
closes #76


ATNoblis (approved):
:shipit:

PR #90, opened 2024-08-01
refactor: createUser to saveUser, and added update function to model

### LINKED ISSUE #84: api createUser should be saveUser
For the sake of consistency, all write operations should be `save<Resource>` with a nullable ID. When ID != null it updates, else it saves

talentedmrjones:
closes #84


ATNoblis (approved):
:shipit:

PR #91, opened 2024-08-08
fix: removed once.Do which was only caching first key

talentedmrjones:
Fixes an issue where keys after the first were never cached, and since the return pulls from the cache it returned nil which caused the token validator to error


ATNoblis (approved):
:shipit:

PR #96, opened 2024-08-21
refactor: adding public alb and cloudfront

talentedmrjones:
CloudFront -> public ALB (configured with OIDC authentication via Okta) will replace verified access.


ATNoblis (approved):
:shipit:

ATNoblis (approved):
LGTM

PR #106, opened 2024-08-28
refactor: replace /whoami with /users/current

talentedmrjones:
`/whoami` was returning a JWT from which the email was then used to call `/users/<email>`
The issues with this were that
- because OIDC auth is handled by the ALB, which acts as a "Backend for Frontend" token handler, the client should only be concerned with the token handler session cookie and never receive a JWT
- /whoami is not RESTful in that it was not returning a resource

`/users/current` DOES return a resource, the object of a single user which is identified by way of the token which is forwarded to the api by the token handler mapped to the session cookie


ATNoblis (approved):
🚢

PR #115, opened 2024-09-06
fix: resolves an issue where /users/current wasnt matching correctly

talentedmrjones:
/users/current was being matched to /users/{userid:} leading the UI into an infinite redirect loop to /login


ATNoblis (approved):
:shipit:

PR #117, opened 2024-09-06
refactor: controllers now obfuscate db errors, and return proper response codes

### LINKED ISSUE #102: api should return consistent wrapper in every response
```
{
  "error": <error>
  "data": <data>
  "metadata": {
    "page": 1,
    "limit": 20,
    "total_items": 100,
    "total_pages": 5
  }
}
```

### LINKED ISSUE #112: controller should wrap and obfuscate DB errors
currently some DB errors are passed straight through to the client which is a security concern. All DB errors should be caught and wrapped by the controller and sent to the client as generic 500s

talentedmrjones:
closes #112 
closes #102

### DIFF
--- a/backend/cmd/api/internal/controller/controller.go
+++ b/backend/cmd/api/internal/controller/controller.go
@@ -2,40 +2,71 @@ package controller
 
 import (
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
+
+	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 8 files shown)


talentedmrjones:
this is a good start but I want to refactor further so that the controller knows nothing of the DB driver. The model should wrap all errors returned so the controller can just check against something like `type.(model.DbError)`

ATNoblis (approved):
🚢

PR #121, opened 2024-09-09
refactor: decouple error handling between controller and db driver

### LINKED ISSUE #118: controller should be decoupled from db driver
As it is now, the controller is checking against `pgx` errors which couples the controller to the db driver which should be entirely isolated by the model. Refactor the model to wrap all pgx errors, and refactor the controller to check against model error types

talentedmrjones:
closes #118 

still need to resolve an issue where an invalid UUID is passed for user/<userid> and controller returns "unknown error". This should be a 404


ATNoblis (approved):
:shipit:

PR #122, opened 2024-09-12
feat: added ability to list, add, and delete users assigned fismasystems

talentedmrjones:
This introduces the `assignedfismasystems` as a distinct resource apart from `fismasystems`

The following routes and functionality are included:

`GET /users/{userid}/assignedfismasystems` returns a list of fismasystem ids
`POST /users/{userid}/assignedfismasystems` creates a record of the assignment using the following body
```
{"fismasystemid": <int>}
```
`DELETE /users/{userid}/assignedfismasystems/{fismasystemid}` deletes the assignment


ATNoblis (approved):
BIG Changes... took me a hot minute but LGTM

PR #145, opened 2024-11-04
feat: imeplemented test db population

### LINKED ISSUE #144: populate test database with dummy data

talentedmrjones:
Implements mechanism to populate a test database with dummy data. If `DB_POPULATE` env var is set to a path, migrations will load that file and execute the SQL contained therein.

The backend test was also modified to include a volume to allow the docker compose container access to the test data sql file

closes #144

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -68,7 +68,7 @@ jobs:
       # TODO: convert this placeholder to a series of tests via BATS
       - name: Curl
         run: |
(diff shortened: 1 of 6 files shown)


ATNoblis (approved):
:shipit:

.github/workflows/backend.yml
```diff
@@ -59,13 +59,16 @@ jobs:
           DB_NAME: ztmf
           DB_USER: admin
           DB_PASS: localDev 
+          DB_POPULATE: "/src/backend/_test_data.sql"
```
ATNoblis, line 62: Is this of any risk to show

talentedmrjones: nope its dummy test data that does not contain any production data

PR #148, opened 2024-11-14
feat: admins have ability to add and edit questions

### LINKED ISSUE #129: admins need ability to add, and edit questions
Need to think through the workflow and UI for this since editing questions is really editing questions, related functions, and function options.

talentedmrjones:
## Added
- controller.SaveQuestion
- controller.GetQuestionByID
- model.FindQuestionByID
- (model.Question).Save()

## Changed
- refactored previous handling of FindQuestions to be clearly FindQuestionsByFismaSystem since they serve 2 different purposes. FindQuestions is for admin management, and the latter is for users listing questions with joined data such as function and pillar

closes #129

### DIFF
--- a/backend/cmd/api/internal/controller/questions.go
+++ b/backend/cmd/api/internal/controller/questions.go
@@ -29,6 +29,19 @@ func ListQuestions(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, questions, err)
 }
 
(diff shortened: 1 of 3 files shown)


eschwein-usds (approved):
Looks good to me.
Additions of new functions look well constructed, and there is better SQL construction and results handling than before.

PR #154, opened 2024-12-04
feat: api logs user events

### LINKED ISSUE #132: api should log user events
User A change ___ from __ to ___

talentedmrjones:
## Added
events table via migration
all inserts, updates, and deletes are recorded as events

# Changed
Refactored entire model package to pass squirrel builders to `query` and `queryRow`, centralizing the builder.ToSQL method, which allows event sourcing to be centralized. Refactored majority of model methods to use the new signatures for `query` and `queryRow` which greatly simplified those functions and reduced repitition

closes #132

### DIFF
--- a/.github/workflows/orchestration-dev.yml
+++ b/.github/workflows/orchestration-dev.yml
@@ -50,7 +50,7 @@ jobs:
       - analysis
       - diff
       - backend
-    if: github.event.pull_request.draft == false && ${{ !failure() && !cancelled() }}
+    if: ${{ github.event.pull_request.draft == false && !failure() && !cancelled() }}
     uses: ./.github/workflows/infrastructure.yml
     with:
       environment: DEV
--- a/.github/workflows/orchestration-dev.yml
+++ b/.github/workflows/orchestration-dev.yml
@@ -8,14 +8,17 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
 
 jobs:
   analysis:
+    if: github.event.pull_request.draft == false
     name: Analysis
     uses: ./.github/workflows/analysis.yml
     secrets: inherit
 
   diff:
+    if: github.event.pull_request.draft == false
     name: Check for Changes
     runs-on: ubuntu-latest
     outputs:
@@ -35,7 +38,7 @@ jobs:
     needs: 
       - analysis
       - diff
-    if: contains(needs.diff.outputs.backend, 'backend/')
+    if: github.event.pull_request.draft == false && contains(needs.diff.outputs.backend, 'backend/')
     uses: ./.github/workflows/backend.yml
     with:
       environment: DEV
@@ -47,7 +50,7 @@ jobs:
       - analysis
       - diff
       - backend
-    if: ${{ !failure() && !cancelled() }}
+    if: github.event.pull_request.draft == false && ${{ !failure() && !cancelled() }}
     uses: ./.github/workflows/infrastructure.yml
     with:
       environment: DEV
--- a/backend/cmd/api/internal/auth/middleware.go
+++ b/backend/cmd/api/internal/auth/middleware.go
@@ -1,7 +1,6 @@
 package auth
 
 import (
-	"context"
 	"log"
 	"net/http"
 	"strings"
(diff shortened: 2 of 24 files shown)


.github/workflows/orchestration-dev.yml
```diff
@@ -8,14 +8,17 @@ on:
       - opened
       - reopened
       - synchronize
+      - ready_for_review
 
 jobs:
   analysis:
+    if: github.event.pull_request.draft == false
```
talentedmrjones, line 15: we don't want draft PRs deployed to dev for the high risk that they will break because they aren't ready

backend/cmd/api/internal/auth/middleware.go
```diff
@@ -45,7 +44,7 @@ func Middleware(next http.Handler) http.Handler {
 				return
 			}
 
-			ctx := context.WithValue(r.Context(), userCtxKey, user)
+			ctx := model.UserToContext(r.Context(), user)
```
talentedmrjones, line 47: moved this to `package model` to avoid a circular import cycle where auth imports model imports auth

backend/cmd/api/internal/model/model.go
```diff
@@ -5,43 +5,85 @@ package model
...
-var sqlBuilder = squirrel.StatementBuilder.PlaceholderFormat(squirrel.Dollar)
+var stmntBuilder = squirrel.StatementBuilder.PlaceholderFormat(squirrel.Dollar)
+
+// SqlBuilder allows methods to receive different types like squirrel.InsertBuilder, squirrel.UpdateBuilder, etc. that all implement the ToSql method
+type SqlBuilder interface {
+	ToSql() (string, []interface{}, error)
+}
+
+// query is a proxy to *pgx.Conn.Query and wrapper around pgx.CollectRows, enabling the centralizing of event tracking
+func query[T any](ctx context.Context, sqlb SqlBuilder, fn pgx.RowToFunc[T]) ([]T, error) {
```
talentedmrjones, line 25: `query` and `queryRow` are perhaps the crux and most crucial piece of this refactor. Every other change across the codebase is just to conform to the change in signature.

Previously, all model functions and struct methods were duplicating efforts when it came to converting the sql builder into a prepared sql statement and bound arguments, and duplicating efforts related to error handling as well.

This change centralizes those steps so that model functions and struct methods need only pass the sql builder and the appropriate pgx.RowToFunc function

backend/cmd/api/internal/model/questions.go
```diff
@@ -15,46 +14,34 @@ type Question struct {
...
+			Suffix("RETURNING " + strings.Join(questionsColumns, ", "))
 	}
 
-	row, err := queryRow(ctx, sql, boundArgs...)
-	if err != nil {
-		return trapError(err)
-	}
-
-	err = row.Scan(&q.QuestionID, &q.Question, &q.NotesPrompt, &q.Order, &q.PillarID)
+	return queryRow(ctx, sqlb, pgx.RowToStructByNameLax[Question])
```
talentedmrjones, line 43: Most `queryRow` will be given `pgx.RowToStructByName` or the `Lax` version.
most `query` will be given `pgx.RowToAddrOfStructByName`

In rare cases a custom function is passed to account for specific details of the calling function that the generalized functions cannot address

backend/cmd/api/internal/model/questions.go
```diff
@@ -63,69 +50,39 @@ func (q *Question) Save(ctx context.Context) error {
...
-	rows, err := query(ctx, sql, boundArgs...)
-
-	if err != nil {
-		log.Println(err)
-		return nil, err
-	}
+		OrderBy("pillars.ordr, questions.ordr ASC")
 
-	return pgx.CollectRows(rows, func(row pgx.CollectableRow) (*Question, error) {
+	return query(ctx, sqlb, func(row pgx.CollectableRow) (*Question, error) {
```
talentedmrjones, line 80: One of the rare examples of where a custom function is passed to account for result fields that don't align to the struct exactly

ATNoblis (approved):
:shipit:

PR #156, opened 2024-12-17
refactor: UUID validation supports values without dashes

### LINKED ISSUE #155: fix: loosen validation for fismauid

talentedmrjones:
## Added
Nothing new

## Changed
model validations `isValidUUID` now tests for the absence of dashes. If none were found it matches against a no dash version. This is because _some_ records of fisma systems do not include dashes in the UUID value

closes #155


ATNoblis (approved):
:shipit:

PR #157, opened 2024-12-31
fix: added security headers

talentedmrjones:
## Added

Added a policy to the CloudFront distribution to ensure that the following response headers are set with secure values:
```
Strict-Transport-Security
Frame-Options
Content-Security-Policy
Referrer-Policy
Permissions-Policy
```


TranScripted (approved):
:shipit:

eschwein-usds (approved):
Looks good to me

PR #160, opened 2025-01-02
fix: content-security-policy headers allow design.cms.gov

talentedmrjones:
## Changed

modified content security policy headers to allow web assets from design.cms.gov


TranScripted (approved):
:shipit:

PR #161, opened 2025-01-06
fix: refactored secrets and error handling, exiting on password error

talentedmrjones:
## Added
nothing new

## Changed
Refactored db and secrets packages to move refreshing into the scope of secrets.
Refactored db conn to trap errors
Refactored model errors to unwrap and exit on password errors. This is not ideal but it will kill the container, force ECS to respawn and from there pick up the latest secret. 
Added logging in an effort to gain more visibility into why this continues to happen despite best attempts and caching and rotation


TranScripted (approved):
:shipit:

PR #175, opened 2025-01-28
fix: increase minimum TLS version to 1.2 in CloudFront distribution

talentedmrjones:
## Added
nothing new

## Changed
Set minimum TLS protocol for CloudFront


ATNoblis (approved):
:shipit:

PR #187, opened 2025-03-11
fix: validate to check whether datacall deadline has passed

### LINKED ISSUE #185: prevent updates to question answers when current date is past datacall deadline

talentedmrjones:
## Added 

- validate method to model/score struct to check if related datacall deadline has passed

## Changed
none

closes #185


ATNoblis (approved):
🚢

PR #188, opened 2025-03-12
implemented soft delete feature for users resource

### LINKED ISSUE #184: Delete a User
An Admin should be able to delete a User, say when they leave the agency.

talentedmrjones:
## Added
- migration to add `delete` column to users table

## Changed
- model/users FindUsers() now returns only users where delete=false
- auth now returns http 401 unauthorized when deleted user tries to log in

closes #184

### DIFF
--- a/backend/cmd/api/internal/migrations/008_users_soft_delete.go
+++ b/backend/cmd/api/internal/migrations/008_users_soft_delete.go
@@ -1,0 +1,9 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		// add soft delete capability
+		"users table role type",
+		`ALTER TABLE IF EXISTS public.users ADD COLUMN deleted boolean NOT NULL DEFAULT FALSE;`,
+		`ALTER TABLE IF EXISTS public.users DROP COLUMN IF EXISTS deleted;`)
+}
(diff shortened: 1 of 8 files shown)


backend/cmd/api/internal/migrations/008_users_soft_delete.go
```diff
@@ -0,0 +1,9 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		// add soft delete capability
+		"users table role type",
+		`ALTER TABLE IF EXISTS public.users ADD COLUMN deleted boolean NOT NULL DEFAULT FALSE;`,
+		`ALTER TABLE IF EXISTS public.users DROP COLUMN IF EXISTS deleted;`)
```
eschwein-usds, line 8: I'm a little confused why the DROP happens after the ADD.  Should you drop before adding the one we want?  Or, is the order in the function parameters not reflective of what order they happen?

eschwein-usds (approved):
Answer my question in the comments of backend/cmd/api/internal/migrations/008_users_soft_delete.go and you are good to go.

backend/cmd/api/internal/migrations/008_users_soft_delete.go
talentedmrjones: All the migrations follow that pattern, the second statement is there in case the migration needs to be rolled back

PR #191, opened 2025-03-17
feat: ability to specify group of email recipients

### LINKED ISSUE #182: need ability specify which group of users gets emailed
When sending emails, we need to be able to specify which group of users gets those emails:
- ISSO
- ISSM
- DCC (Data Call Contacts)

talentedmrjones:
## Added
- migration to add `group` column to massemails table
- ability to specify which group of recipents gets the email being sent

## Changed
- changed route from email/ to massemails/ to remain consistent with route to resource mapping
- renamed controller/emails to controller/massemails
- renamed model/emails to model/massemails
- moved functionality of model/datacallcontacts to model/massemails
- made mail/send recieve `[]string` instead of having it look up recipients 
- made request POST to support returning data with response
- removed test mode env var

closes #182


eschwein-usds (dismissed):
Looks good to me.

eschwein-usds (approved):
Still looks good to me

PR #192, opened 2025-03-18
feat: ability to filter list of users via query string parameters

### LINKED ISSUE #189: ability to filter list of users
Currently `GET /api/v1/users` always lists all users where delete=false

It is necessary to display users according to values of deleted, role, and potentially email (`like '%email%'`)

talentedmrjones:
## Added
- model/users `FindUsersInput` type

## Changed
- `controller.ListUsers` now hydrates `FindUsersInput` and passes to `model.FindUsers`
- `model.FindUsers` now filters based on all fields of `FindUsersInput` that are present in the request

## Usage
To filter list of users, append the following query string parameters in any order in any combination:
`?role=string&email=string&fullname=string&deleted=bool`

`role` and `deleted` will be used as an exact match
`email` and `fullname` will be used as a loose match e.g.: `email=cms.hss.gov` will return all users with that as part of their email, `fullname=bob` will return all users with "bob" anywhere in their name

closes #189

### DIFF
--- a/.github/workflows/orchestration-dev.yml
+++ b/.github/workflows/orchestration-dev.yml
@@ -12,7 +12,6 @@ on:
 
 jobs:
   analysis:
-    if: github.event.pull_request.draft == false
     name: Analysis
     uses: ./.github/workflows/analysis.yml
     secrets: inherit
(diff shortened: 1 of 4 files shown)


eschwein-usds (approved):
I approve

PR #205, opened 2025-04-09
feat: scores include functionoption

### LINKED ISSUE #194: add function option score to get score route

talentedmrjones:
## Added
- `input` type in the model which is used for embedding into `find<resource>Input` structs

## Changed
- extended the `model/FindScores` and `FindScoresInput` to allow including the `functionoption` sub resource

## Usage
```
/api/v1/scores?datacallid=<int>&fismasystemid=<int>&include=functionoption
```

Where `include` is optional and currently only supports `functionoption` as a value

closes #194


TranScripted (approved):
:shipit:

PR #207, opened 2025-04-10
fix: resolved issue where save returned error cannot find field FunctionOption

talentedmrjones:
## Added
non

## Changed
- fixed issue with model function to covert DB record to Scores struct, ignoring FunctionOption field when not present


TranScripted (approved):
🚢

PR #210, opened 2025-04-16
infra: added s3 interface endpoint

talentedmrjones:
## Added
- s3 interface endpoint

This solves an issue where docker image pulls download layers from S3 and were traversing the NAT gateway


eschwein-usds (approved):
looks good to me

PR #211, opened 2025-04-16
fix: added unique email index to users

### LINKED ISSUE #201: api should prevent duplicate emails regardless of case

talentedmrjones:
## Added
- migration to remove the unique constraint which is case-sensitive, and replace with a unique index bound to the lower() function for case-insensitive uniqueness

# Changed
DB schema as mentioned above

closes #201


eschwein-usds (approved):
Looks good to me

PR #287, opened 2026-04-14
Add S3-triggered cert rotation Lambda (DEV)

Whardison1911:
**Summary**
-Adds an S3-triggered Lambda for dev.ztmf.cms.gov that validates cert.pem/key.pem/chain.pem.

-Re-imports a fixed ACM cert ARN, backs up the bundle to Secrets Manager, archives files to _processed/<env>/<timestamp>/_

-Sends Slack success/failure notifications (placeholder was created until we obtain the webhook url).

**Includes Terraform to deploy Lambda, IAM, and S3 notifications**


Whardison1911:
Here are the steps that I took to remediate:

**Removed the  terraform.tfstate, terraform.tfstate.backup, bootstrap, and lambda.zip from git history entirely by utilizing the BFG Repo-Cleaner to scrub them from history.**

- This allowed me to rewrite the branch history so the file look like they were never committed at all.

**Add all four files to .gitnore before re-pushing**

- This will prevent Git from tracking the files again in future commits.

**Force pushed the cleaned branch**

- Since rewriting changes commit hashes, Git sees it as a different branch history.  A normal push doesn't work. The bad files disappear from the PR history and GitHub reflects the new commit chain due to the force push.

PR #288, opened 2026-04-14
Increase notes limit, add validation, allow ADMIN post-deadline saves

danielbowne:
## Summary

- Adds migration to increase `scores.notes` from VARCHAR(1000) to VARCHAR(2000) so users have more room to describe their system posture per question
- Adds server-side notes length validation in `Score.validate()` using `utf8.RuneCountInString()` to correctly count characters (not bytes) and return a clear 400 error instead of a cryptic DB constraint failure
- ADMINs can now save scores after the datacall deadline has passed - the deadline check in `validate()` is skipped for ADMIN role so they can correct responses after close
- Updates OpenAPI spec with `maxLength: 2000` on the notes field
- Adds Emberfall E2E test for notes exceeding 2000 characters

The companion frontend PR is at CMS-Enterprise/ztmf-ui (same branch name) and handles the UI enforcement, character counter, and role-based editing restrictions.

**Note:** This will auto-deploy to dev when CI passes.

## Test plan

- [ ] CI pipeline passes (unit tests + Emberfall E2E)
- [ ] Verify on dev after deploy: POST score with notes > 2000 chars returns 400
- [ ] Verify on dev: ADMIN can save scores after deadline passes
- [ ] Verify on dev: non-ADMIN gets 403 when attempting to save past deadline


jsos3-cms (approved):
LGTM, going to add an approve.

I'm not a fan of [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#:~:text=In%20computer%20programming%2C%20a%20magic,classified%20as%20a%20magic%20number.) and this 2k limit may feel "magic" by that respect. Is it possible to define it once in the code, and have future modifications, lets say up to 4k characters, not need as much of a change?

Alternatively, if we could bookmark this PR for the next time we need to do a character change, that could be easy too.

danielbowne:
> LGTM, going to add an approve.
> 
> I'm not a fan of [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)#:~:text=In%20computer%20programming%2C%20a%20magic,classified%20as%20a%20magic%20number.) and this 2k limit may feel "magic" by that respect. Is it possible to define it once in the code, and have future modifications, lets say up to 4k characters, not need as much of a change?
> 
> Alternatively, if we could bookmark this PR for the next time we need to do a character change, that could be easy too.

Good point on the magic number. This was a design decision made by the previous devs - the existing codebase uses inline values throughout the model layer (VARCHAR lengths, validation limits, etc.). We're keeping with the existing patterns here since the focus of this PR is fixing a bug and adding validation, not refactoring conventions.

Also worth noting that even with a Go constant we'd still have at least two hardcoded spots - the OpenAPI spec (openapi.yaml) doesn't support variables so maxLength has to be a literal value, and the migration SQL is a point-in-time schema change so we can't use constants there either. So 2 out of 3 locations would still need manual updates regardless.

This PR is a good bookmark for where the limit lives if we ever need to change it.

PR #292, opened 2026-04-22
lambda-cert-rotation-draft

Whardison1911:
@danielbowne please review the drafted PR

For "backend/cmd/lambda-cert-rotation/internal/certvalidator/validator_test.go" this is test-only code that generates throwaway self-signed certificates/keys in memory for units tests.  It doesn't ship in the Lambda runtime and it doesnt contain any real secrets.

Just wanted to state this if Snyk throws any false positives for private key material (even in _test.go files).


jsos3-cms (approved):
Great work cleaning this up :) Love it.

Is this cert rotation lambda only relevant for ZTMF? Could this be its own module/repo, and we import it here? Just a thought. For now, let's keep it moving.

danielbowne:
> Great work cleaning this up :) Love it.
> 
> Is this cert rotation lambda only relevant for ZTMF? Could this be its own module/repo, and we import it here? Just a thought. For now, let's keep it moving.

Yes, normally I would do this as its own repo, but due to difficulty of repos at CMS, wasn't sure how we want to handle.

PR #297, opened 2026-04-29
508 compliance: restore user + reactivate system endpoints

### LINKED ISSUE #281: 508 Compliance: Add confirmation dialogs and restore capability for all destructive actions
## Context

508/accessibility compliance requires that destructive actions have confirmation prompts and a way to recover without direct database access. Currently several delete/soft-delete operations are missing one or both.

## Current State Audit

| Action | Endpoint | Delete Type | Confirmation? | Restore? |
|--------|----------|-------------|---------------|----------|
| Delete User | `DELETE /users/{id}` | Soft (`deleted=true`) | **NO** | **NO** |
| Decommission System | `DELETE /fismasystems/{id}` | Soft (decommissioned flags) | Yes (two-step) | **NO** |
| Unassign System from User | `DELETE /users/{id}/assignedfismasystems/{id}` | **Hard delete** | **NO** | N/A (re-assign manually) |

### 1. Missing confirmation dialogs
- **Delete User** (`ztmf-ui: UserTable.tsx:357-388`): Clicking delete immediately removes the user with no "Are you sure?" prompt
- **Unassign System** (`ztmf-ui: AssignSystemModal.tsx:146-186`): Unchecking a system in the Autocomplete immediately fires the DELETE — no confirmation

### 2. No restore/undo capability
- **Deleted users**: `users.deleted = true` in DB, no API or UI to set it back to `false`. Only fix is manual SQL.
- **Decommissioned systems**: Confirmation dialog explicitly states "This action cannot be undone through the UI." No reactivate endpoint or UI exists. (See ztmf-ui#318 for the reactivation discussion.)
- **Unassigned systems**: Hard delete from `users_fismasystems` junction table — data is gone, must re-assign manually. This one is less critical since re-assignment is straightforward through existing UI.

#### Restore Deleted Users
- Add `PUT /api/v1/users/{userid}/restore` endpoint (admin-only)
- Sets `deleted = false`
- Logs event for audit trail

#### Reactivate Decommissioned Systems
- Add `PUT /api/v1/fismasystems/{fismasystemid}/reactivate` endpoint (admin-only)
- Sets `decommissioned = false`, clears decommission metadata
- Optionally add `reactivated_by`, `reactivated_date`, `reactivation_notes` columns for audit trail
- See ztmf-ui#318 for design options (Option A: simple clear vs Option B: with audit trail)

#### Update OpenAPI spec
- Document new restore/reactivate endpoints in `backend/openapi.yaml`

#### Add missing confirmation dialogs
- **Delete User**: Add ConfirmDialog before DELETE call — "Are you sure you want to delete {user name}? This will remove their access to ZTMF."
- **Unassign System**: Add ConfirmDialog when unchecking a system — "Are you sure you want to unassign {system name} from {user name}?"

#### Add restore/reactivate UI
- **Deleted users**: Add toggle or filter to show deleted users in UserTable, with a "Restore" action button (admin-only)
- **Decommissioned systems**: Enable reactivation from the system detail page or decommissioned systems list (admin-only). Update confirmation text to remove "cannot be undone" language.

## Acceptance Criteria

- [ ] All destructive actions have an "Are you sure?" confirmation dialog
- [ ] Soft-deleted users can be restored via API and UI (admin-only)
- [ ] Decommissioned systems can be reactivated via API and UI (admin-only)
- [ ] Audit trail captures who restored/reactivated and when
- [ ] OpenAPI spec updated for new endpoints
- [ ] Emberfall E2E tests for restore and reactivate endpoints

## Related Issues

- ztmf-ui#318 — Reactivate/un-decommission systems (design options)
- Backend decommission: PR #241
- Frontend decommission: ztmf-ui#314, ztmf-ui#323

danielbowne:
## Summary

Stage 1 of issue #281 (508 compliance for destructive actions). Adds two admin-only endpoints so soft-deleted users and decommissioned FISMA systems can be recovered through the API instead of direct database access.

- `PUT /api/v1/users/{userid}/restore` clears the deleted flag and returns the updated user (HTTP 200).
- `PUT /api/v1/fismasystems/{fismasystemid}/reactivate` clears the decommissioned flag, stamps `reactivated_by`/`reactivated_date`/`reactivation_notes`, and preserves the prior `decommissioned_*` audit columns so the decommission record remains queryable.

Both endpoints reject READONLY_ADMIN and ISSO with 403, return 404 for unknown ids, and return 400 with `InvalidInputError` when the target is already in the requested state.

Folded in dependabot PR #296: `jackc/pgx/v5` 5.9.1 -> 5.9.2 (resolves GHSA-j88v-2chj-qfwx, SQL injection via dollar-quoted placeholder confusion in the simple protocol). One PR, one approval.

Fixes #281
Closes #296

## Implementation notes

- Migration 0025 adds `reactivated_by` (UUID FK to `users` with `ON DELETE SET NULL`), `reactivated_date` (TIMESTAMPTZ), and `reactivation_notes` (TEXT) to `fismasystems` plus a partial index on `reactivated_by`.
- Both model functions use `SELECT FOR UPDATE` inside a transaction so the not-found vs already-active decision is atomic. Audit events are recorded explicitly because the transactional path bypasses the `queryRow` auto-event hook.
- New `respondOK` helper in `controller.go` writes 200 with body for PUT-as-action endpoints. The existing `respond` function still returns 204 for in-place PUT updates and is left untouched to avoid changing the contract on `SaveUser`, `SaveFismaSystem`, etc.
- OpenAPI spec extended with both paths and three new fields on the `FismaSystem` schema.
- Fixture additions in `_test_data_empire.sql`: a pre-deleted user (Captain Needa) and a pre-decommissioned fismasystem (Star Destroyer Tyrant, id 1004) for unit-test and dev smoke coverage. Captain Needa uses a v4-conforming UUID so the existing `isValidUUID` regex accepts it as a path parameter.

## Stage 2 (separate branch)

The audit in #281 listed four ztmf-ui surfaces that depend on these endpoints. Stage 2 covers them on the next branch:

1. Confirmation dialog before Delete User.
2. Confirmation dialog before Unassign System.
3. Show-deleted-users toggle and Restore action in `UserTable`.
4. Reactivate UI on decommissioned systems (and remove the "cannot be undone" copy in the existing decommission dialog).

## Test plan

- [x] `make dev-setup` resolves cleanly with the new migration applied; `\d fismasystems` shows the three new columns and the FK
- [x] `go build ./...` clean against all backend binaries
- [x] `make test-e2e` passes 84/0/0 against the isolated `compose-test.yml` stack (port 8090). Includes the new positive flows and READONLY_ADMIN/ISSO 403 negatives for both endpoints
- [x] Manual smoke from `localhost:3000`:
  - Restore Captain Needa: 200 with `deleted: false`
  - Replay: 400 `{"deleted": "user is already active"}`
  - Bogus UUID: 404
  - Reactivate fismasystem 1004: 200 with `decommissioned: false`, `reactivated_*` populated, decommission audit preserved
  - Replay: 400 `{"decommissioned": "system is already active"}`
  - Bogus id: 404
  - READONLY_ADMIN and ISSO tokens: 403 on all four PUTs
  - `events` table contains rows for both updates against the admin's UserID
- [ ] CI green on the PR (analysis/snyk/lint/backend smoke)
- [ ] Dev auto-deploy completes; spot-check from the dev environment before stage 2 merges

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -28,6 +28,11 @@ INSERT INTO public.users VALUES (DEFAULT, 'Readonly.Admin@nowhere.xyz', 'Readonl
 -- Fixed UUID so we can assign to fismasystems for CFACTS access testing.
 INSERT INTO public.users VALUES ('66666666-6666-6666-6666-666666666666', 'Isso.User@nowhere.xyz', 'ISSO Test User', 'ISSO', DEFAULT) ON CONFLICT DO NOTHING;
 
+-- Pre-deleted user fixture for RestoreUser tests.
+-- UUID is v4-conforming (4 at position 14, 8 at position 19) so it satisfies
+-- isValidUUID's strict regex when used as a path param.
+INSERT INTO public.users VALUES ('77777777-7777-4777-8777-777777777777', 'Captain.Needa@executor.empire', 'Captain Needa', 'ISSO', TRUE) ON CONFLICT DO NOTHING;
+
 -- Test Pillars (using production pillar names for testing consistency)
 INSERT INTO public.pillars VALUES (1, 'Devices', 0) ON CONFLICT DO NOTHING;
 INSERT INTO public.pillars VALUES (2, 'Applications', 0) ON CONFLICT DO NOTHING;
@@ -100,6 +105,27 @@ INSERT INTO public.fismasystems (fismasystemid, fismauid, fismaacronym, fismanam
     NULL,
     NULL,
     NULL
+) ON CONFLICT DO NOTHING;
+
+-- Pre-decommissioned fixture for ReactivateFismaSystem tests
+INSERT INTO public.fismasystems (fismasystemid, fismauid, fismaacronym, fismaname, fismasubsystem, component, groupacronym, groupname, divisionname, datacenterenvironment, datacallcontact, issoemail, sdl_sync_enabled, decommissioned, decommissioned_date, decommissioned_by, decommissioned_notes) VALUES (
+    1004,
+    'BC1B3100-1980-4D5E-AB8C-D1FE0BB00808',
+    'SD-TYR',
+    'Star Destroyer Tyrant',
+    'Imperial Class Destroyer',
+    'IMPNAVY-(FLEET)',
+    'STARCOM',
+    'Imperial Starfleet Command',
+    'Naval Operations Division',
+    'Imperial-Fleet',
+    'admiral.ozzel@executor.empire',
+    'Captain.Needa@executor.empire',
+    FALSE,
+    TRUE,
+    '1980-05-21 00:00:00+00',
+    '11111111-1111-1111-1111-111111111111',
+    'Decommissioned to provide a reactivation test target'
 ) ON CONFLICT DO NOTHING;
 
 -- User-System Assignments (Officers assigned to their systems)
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -95,6 +95,24 @@ func TestSaveUser_ISSOForbidden(t *testing.T) {
 
 // --- DeleteUser ---
 
+func TestRestoreUser_ReadonlyAdminForbidden(t *testing.T) {
+	r := httptest.NewRequest("PUT", "/api/v1/users/11111111-1111-1111-1111-111111111111/restore", nil)
+	r = withUser(r, readonlyAdmin)
+	w := httptest.NewRecorder()
+
+	RestoreUser(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
+func TestRestoreUser_ISSOForbidden(t *testing.T) {
+	r := httptest.NewRequest("PUT", "/api/v1/users/11111111-1111-1111-1111-111111111111/restore", nil)
+	r = withUser(r, issoUser)
+	w := httptest.NewRecorder()
+
+	RestoreUser(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
 func TestDeleteUser_ReadonlyAdminForbidden(t *testing.T) {
 	r := httptest.NewRequest("DELETE", "/api/v1/users/11111111-1111-1111-1111-111111111111", nil)
 	r = withUser(r, readonlyAdmin)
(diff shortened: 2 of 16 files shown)


MackOverflow (dismissed):
Looks good Danny, two things my review flagged (not blockers, but things to consider):

---

**1. On the `TestFismaSystemReactivationFields` test** ([`fismasystems_test.go:648-650`](backend/internal/model/fismasystems_test.go))

> Asserting column positions by hardcoded index (`fismaSystemColumns[17]`, `[18]`, `[19]`) is a little brittle — if anyone inserts a column earlier in the slice these assertions flip to the wrong names without a clear failure message. Could use `slices.Contains` or just check the last 3 elements to make it more resilient to future column additions.

---

**2. On `RestoreUser` vs `ReactivateFismaSystem` scan style** ([`users.go:716-719`](backend/internal/model/users.go))

> Minor inconsistency: `RestoreUser` manually scans the RETURNING columns into struct fields, but `ReactivateFismaSystem` uses `pgx.CollectOneRow` + `pgx.RowToStructByName`. Not a bug, but if new fields get added to `User` later it's easy to forget to update the manual scan. Probably worth making them consistent?

MackOverflow (commented):
@danielbowne Removing a user is functioning as exepcted, but I am getting an error when I try to add. The new user appears in the list, but does not last past a refresh.

Image

Other notes (none of them blockers): 


---

**1. On the `TestFismaSystemReactivationFields` test** ([`fismasystems_test.go:648-650`](backend/internal/model/fismasystems_test.go))

> Asserting column positions by hardcoded index (`fismaSystemColumns[17]`, `[18]`, `[19]`) is a little brittle — if anyone inserts a column earlier in the slice these assertions flip to the wrong names without a clear failure message. Could use `slices.Contains` or just check the last 3 elements to make it more resilient to future column additions.

---

**2. On `RestoreUser` vs `ReactivateFismaSystem` scan style** ([`users.go:716-719`](backend/internal/model/users.go))

> Minor inconsistency: `RestoreUser` manually scans the RETURNING columns into struct fields, but `ReactivateFismaSystem` uses `pgx.CollectOneRow` + `pgx.RowToStructByName`. Not a bug, but if new fields get added to `User` later it's easy to forget to update the manual scan. Probably worth making them consistent?

danielbowne:
Quick context on the "error when adding a user" Mack flagged:

The flow was deleting an existing user (Allie) and then trying to add her back by typing her email into the Add User form. That is not actually the supported recovery path in this PR — the way to bring a soft-deleted user back is to flip the new **Show Deleted** toggle on the toolbar and click the **Restore** action on her row. The toggle is the discoverability surface for deleted users; Add User is intended for net-new emails only.

The reason the typed-email path errors instead of just succeeding: `users.email` has a case-insensitive unique index, and soft-deleting a user does not free up that email. Re-inserting trips the constraint. That behavior is unchanged in this PR.

That said, Mack's flow uncovered a real backend error-handling bug that is now fixed in commit `afe1cff`:

- `trapError` in `backend/internal/model/errors.go` was using `errors.Unwrap(e)` to extract `*pgconn.PgError`, which returns nil when pgx hands the error back unwrapped. The type switch then missed the PgError case entirely and trapError returned a generic "unknown error", causing the controller to send a 500 with `{"error": "server error"}` for what should have been a 400 with a clear unique-violation message.
- Switched to `errors.As`, which walks the wrap chain and matches whether the error is bare or wrapped. Added a defensive bounds check on the detail-message split.
- Net effect: duplicate-email INSERT now returns `400 {"error": "not unique : (email@example.com) already exists."}`. Foreign-key violations (`23503`) get the same fix and now correctly return 400. Same trapError function feeds every model write in the codebase, so this fixes the silent-500 path for any existing or future endpoint that hits a unique or FK constraint.

Pre-existing bug, not introduced by this PR — but Mack's testing surfaced it and the fix is small + tightly related, so folding it in here rather than opening a separate PR.

Test plan for the new commit: POST `/api/v1/users` with an email that already exists in dev returns 400 with the formatted `ErrNotUnique` message instead of 500. POST with a new email still returns 201. Existing tests untouched.

MackOverflow (approved):
approved

Issue #212, opened 2025-04-17
Include Answer Text in Download


eschwein-usds:
In the download-able spreadsheet, the score is listed as a number.  there has been a request for the text of the answer to also be included.  

I would list it as a separate column, so that the score can be used in team things.

jsos3-cms:
I believe this is already how the download feature functions today. @eschwein-usds - can you download the new spreadsheet and confirm that the columns are as expected?

costacalvin:
Following up, believe this ticket can be closed due to https://github.com/CMS-Enterprise/ztmf/pull/298. @a-dipietro @jsos3-cms

a-dipietro:
@costacalvin did we confirm this applies to the downloaded data as well as the UI? If yes, okay to close. Thanks!

costacalvin:
@a-dipietro Can confirm this is included in the downloaded data: 

[FY2023ImperialSecurityReview.xlsx](https://github.com/user-attachments/files/29470404/FY2023ImperialSecurityReview.xlsx)

PR #298, opened 2026-04-29
Export: replace tier label with answer text and add Maturity Tier column

danielbowne:
## Summary

Fixes the datacall xlsx export so the "Answer" column carries the actual descriptive text the ADO selected, not just the tier label that duplicates the numeric Score. Adds a separate "Maturity Tier" column so non-experts who do not have the 0-4 score legend memorized can still see `initial`/`advanced`/`optimal` next to each row.

Issue: #212

## What changed in the spreadsheet

Old layout (9 columns):

| A | B | C | D | E | F | G | H | I |
|---|---|---|---|---|---|---|---|---|
| Fisma Acronym | Data Center Environment | Pillar | Function | Function Description | Question | **Answer** _(tier label, e.g. `initial`)_ | **Score** | ADO Answer Details |

New layout (10 columns -- same first 6, **G/H/I/J** updated):

| A | B | C | D | E | F | G | H | I | J |
|---|---|---|---|---|---|---|---|---|---|
| Fisma Acronym | Data Center Environment | Pillar | Function | Function Description | Question | **Answer** _(descriptive text)_ | **Maturity Tier** _(initial/advanced/etc)_ | **Score** | ADO Answer Details |

## Sample output (fake data, no real systems)

| Fisma Acronym | Data Center Environment | Pillar | Function | Question | **Answer** | **Maturity Tier** | **Score** | ADO Answer Details |
|---|---|---|---|---|---|---|---|---|
| EXAMPLE-1 | CMS-Cloud-AWS | Networks | NetworkSegmentation | How does the system define the network architecture? | Application has intra-account, resource level and intra-VPC segmentation; uses AWS PrivateLink for service traffic. | optimal | 4 | Architecture documented in design ADR; reviewed by cloud security in last quarterly audit. |
| EXAMPLE-1 | CMS-Cloud-AWS | Devices | Device-ThreatProtection | How does the system deploy threat protection capabilities for all devices? | Endpoint protection deployed via gold image, with anomaly detection and centralized log review. | advanced | 3 | TrendMicro DSM enforced; observability through NewRelic agent. |
| EXAMPLE-1 | CMS-Cloud-AWS | Data | DataAccess | How does the system implement data access controls? | Custom-built static role mapping for application access; identity provider used only for human users. | traditional | 1 | Static row-level security in custom code; no automated entitlement reviews. |

Reading the row: the human reads "Maturity Tier = advanced" and immediately understands what "Score = 3" means without going back to a score legend. The "Answer" column carries the full sentence the ADO read on the questionnaire when picking that option.

## Adjacent fixes folded in

While in this code path I noticed three small issues; including them here so the export endpoint is solid end-to-end rather than splitting into multiple PRs.

1. **Pre-existing panic on empty answers.** `GetDatacallExport` derived the filename from `answers[0].DataCall`, which dereferences index 0 on an empty slice and panicked the request. Now falls back to `datacall-<id>` when there are no scored answers; the response is a valid header-row-only xlsx instead of a 500.
2. **Content-Disposition filename was unquoted.** Datacall names containing spaces or punctuation could produce a malformed header per RFC 6266. Quoted now.
3. **`excelize.File.Write` error was discarded.** Headers are already on the wire so we cannot change the status code, but the error is now logged with the datacall id so a mid-stream client disconnect leaves a server-side signal.

## Test plan

- [x] `make test-e2e` -- 85/0/0 against the isolated `compose-test.yml` stack (was 84 before; new export-endpoint test added)
- [x] `go build ./...` clean across all backend binaries
- [x] Manual smoke against local dev with synced dev data:
  - `GET /api/v1/datacalls/<id>/export?fsids=<sysid>` returns 200, valid xlsx file
  - Column G contains full descriptive answer text for every row
  - Column H contains the tier label (`traditional`/`initial`/`advanced`/`optimal`) matching column I (Score)
  - Column I numeric, column J carries `scores.notes` (ADO answer details)
  - Empty-answers path returns 200 with header-row-only xlsx and `datacall-<id>.xlsx` filename (no panic)
  - Content-Disposition header is `attachment; filename="<name>.xlsx"` (quoted)
- [ ] CI green on PR (analysis + backend smoke)
- [ ] Dev auto-deploy completes; spot-check from dev environment

## Frontend impact

None. `ztmf-ui/src/views/FismaTable/FismaTable.tsx::saveSystemAnswers` downloads the response as a binary blob (`responseType: 'blob'`) and never parses columns. The new column shape is purely a server-side concern.

## OpenAPI impact

None. `/datacalls/{datacallid}/export` is already declared as `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` binary with no per-column schema.

### DIFF
--- a/backend/cmd/api/internal/controller/datacalls.go
+++ b/backend/cmd/api/internal/controller/datacalls.go
@@ -69,9 +69,12 @@ func GetDatacallExport(w http.ResponseWriter, r *http.Request) {
 	if len(answers) > 0 {
 		filename = strings.ReplaceAll(answers[0].DataCall, " ", "")
 	}
-	// Quote the filename per RFC 6266 so datacall names containing spaces or
-	// punctuation produce a well-formed Content-Disposition header.
-	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="%s.xlsx"`, filename))
+	// Filename is left unquoted because the frontend (FismaTable.saveSystemAnswers)
+	// parses the header by splitting on `filename=` and uses the resulting value
+	// directly as the anchor's download attribute. Chrome sanitizes filesystem-
+	// unsafe characters in that attribute -- including the double-quote -- which
+	// turns a quoted filename into _name.xlsx_ and breaks the .xlsx association.
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s.xlsx", filename))
 	w.Header().Set("Content-Type", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
 	// Headers are already on the wire by this point, so a write error cannot
 	// be surfaced as a 5xx -- the client will see a truncated download. Log
--- a/backend/cmd/api/internal/controller/datacalls.go
+++ b/backend/cmd/api/internal/controller/datacalls.go
@@ -62,9 +62,23 @@ func GetDatacallExport(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s.xlsx", strings.ReplaceAll(answers[0].DataCall, " ", "")))
+	// Fall back to the datacall id when no answers exist so the export still
+	// returns a valid, named xlsx (header row only) instead of panicking on
+	// answers[0].DataCall.
+	filename := fmt.Sprintf("datacall-%d", findAnswersInput.DataCallID)
+	if len(answers) > 0 {
+		filename = strings.ReplaceAll(answers[0].DataCall, " ", "")
+	}
+	// Quote the filename per RFC 6266 so datacall names containing spaces or
+	// punctuation produce a well-formed Content-Disposition header.
+	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="%s.xlsx"`, filename))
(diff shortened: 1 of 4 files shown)


backend/cmd/api/internal/controller/datacalls.go
eschwein-usds, line 1: Something is not quite right with the filename.  When I (using Chrome 147.0.7727.137 (Official Build) (arm64)), selected 1 system and downloaded the file, the filename is surrounded by "_", which means the file cannot be opened by double clicking as the extension is wrong. see screenshot.

Image

eschwein-usds (changes requested):
Mostly looks good, and the contents of the file are now correct.  But if you could fix the file name that would be appreciated.

backend/cmd/api/internal/controller/datacalls.go
danielbowne: Good catch @eschwein-usds and thanks! I had quoted the filename in the header to make it RFC 6266 compliant, but that fix broke the actual download - Chrome treated the double-quotes as filesystem-unsafe characters and converted them to underscores when saving, which is why the .xlsx association was lost. Pushed a fix that drops the quotes; downloads should now save as plain "FY2026Q1.xlsx" again.  CI is deploying now!

eschwein-usds (approved):
File name Looks good now!

Issue #283, opened 2026-04-06
Automate TLS certificate import with S3 trigger, validation, and Slack alerts


danielbowne:
## Problem

DigiCert TLS certificates are manually imported into AWS ACM via CLI. In January 2026, a cert was imported without the intermediate CA chain, causing 502 errors on CloudFront and breaking login for dev. A secondary issue was two certs with the same domain coexisting in ACM, causing Terraform to silently pick the wrong one.

The DigiCert cert workflow is a CMS requirement and will not change. What we can automate is the import, validation, and notification steps that happen after the cert files are received.

## Solution

Build an S3-triggered Lambda that validates cert files before importing to ACM, with Slack notifications on success or failure.

### Workflow

1. Team member receives new cert files from DigiCert (server cert, private key, intermediate CA chain)
2. Upload 3 files to S3 bucket:
   ```
   aws s3 cp cert.pem  s3://ztmf-cert-rotation-dev/dev/cert.pem
   aws s3 cp key.pem   s3://ztmf-cert-rotation-dev/dev/key.pem
   aws s3 cp chain.pem s3://ztmf-cert-rotation-dev/dev/chain.pem
   ```
3. Lambda triggers automatically, validates the cert bundle, imports to ACM, backs up to Secrets Manager, sends Slack notification
4. CloudFront and ALB pick up the updated cert automatically (same ARN, ACM handles distribution)
5. No Terraform changes needed

### Architecture

```text
S3 Bucket (ztmf-cert-rotation-<env>)
  dev/cert.pem, key.pem, chain.pem
  impl/cert.pem, key.pem, chain.pem
  prod/cert.pem, key.pem, chain.pem
         |
         v  (S3 ObjectCreated, filter *.pem)
         |
  Lambda (ztmf-cert-rotation-<env>)
         |
         +-- 1. Check all 3 files present for env prefix
         |      (if not, exit cleanly, no noise)
         +-- 2. Validate cert bundle
         |      - chain.pem has intermediate CA (prevents Jan 2026 incident)
         |      - Server cert chains to intermediate
         |      - Private key matches certificate
         |      - Domain matches expected env
         |      - Cert is not expired
         +-- 3. Import to ACM (re-import over existing ARN)
         +-- 4. Backup to Secrets Manager
         +-- 5. Archive files to processed/<env>/<timestamp>/
         +-- 6. Slack notification (success or failure)
         |
    +----+----+
    |         |
    v         v
  ACM     Secrets Manager
[...]

PR #300, opened 2026-05-01
Activate cert-rotation Lambda + align ztmf_api log retention

danielbowne:
## Summary

Turns on the S3-triggered cert-rotation Lambda that shipped in #292. The Lambda code, Terraform resources, IAM, DLQ, and CloudWatch alarms have all been on `main` since the merge but were gated by `enable_cert_rotation_lambda = false` in tfvars. This PR flips the toggle and wires the per-environment ACM ARN through SSM Parameter Store rather than committing it to tfvars.

Also aligns the `ztmf_api` CloudWatch log group retention with the CMS Cloud automation that resets it monthly, so Terraform stops drifting on every apply.

## What changed

**`infrastructure/lambda-cert-rotation.tf`**
- Adds `data "aws_ssm_parameter" "cert_rotation_acm_arn"` reading `/ztmf/${var.environment}/cert-rotation/acm-arn`, gated by the same `enable_cert_rotation_lambda` flag.
- Removes the `cert_rotation_acm_certificate_arn` Terraform variable. ARNs and account IDs no longer live in the repo.
- Wraps the SSM value with `nonsensitive()` because an ACM ARN is not secret material; CloudFront and the internal ALB already publish the same ARN.
- Adds a `precondition` on `aws_lambda_function.cert_rotation` that re-establishes the ARN-shape regex guard the deleted variable used to enforce, so a typo in the SSM parameter is caught at plan time.

**`infrastructure/ecs.tf`**
- Sets `aws_cloudwatch_log_group.ztmf_api.retention_in_days = 731`. CloudTrail shows a CMS Cloud automation Lambda named `loggroups-retention-policy-lambda` resets every log group to 731 days on the 1st of each month. Without this, Terraform fights the policy and prints `731 -> 0` drift on every plan.

**`infrastructure/tfvars/dev.tfvars` and `infrastructure/tfvars/prod.tfvars`**
- `enable_cert_rotation_lambda = true` in both.
- Removed `cert_rotation_acm_certificate_arn` line; sourced from SSM now.
- Fixed `cert_rotation_domain` in `prod.tfvars` from `prod.ztmf.cms.gov` to `ztmf.cms.gov`. The new multi-SAN cert covers `ztmf.cms.gov`, `impl.ztmf.cms.gov`, and `dev.ztmf.cms.gov`; there is no `prod.` subdomain in DNS.

## Pre-merge runbook

1. **Per account, populate the SSM parameter** with the ACM ARN holding the new multi-SAN cert:
   ```bash
   aws ssm put-parameter --profile ztmf-<env> --region us-east-1 \
     --name /ztmf/<env>/cert-rotation/acm-arn \
     --type String \
     --value arn:aws:acm:us-east-1:<account>:certificate/<uuid>
   ```
   Without this, `terraform plan` will fail on the data source lookup.

2. **Confirm the watched S3 prefix is empty.** The bucket `ztmf-cert-rotation-<env>` is created by Terraform during this apply. If anything leftover is at `<env>/*.pem` from earlier manual testing, the next upload will pair with stale files and fail the freshness check.

3. **Run `terraform plan` against each env** and confirm the only adds are cert-rotation resources plus the log retention update.

## Behavior notes

- Dev runs the Lambda in `DRY_RUN = true` mode (validates only, never imports to ACM). Operators testing the full rotation path on dev have to flip the env variable on the function manually or accept that dev is validation-only. Comment in `internal/config/config.go` explains the rationale.
- The Lambda re-imports over the configured ACM ARN, which is why a stable ARN per environment is the right pattern.
- Future ACM ARN changes require updating the SSM parameter and running `terraform apply`; the value is baked into the Lambda env at plan time.

## Test plan

- [x] SSM parameter set in both `ztmf-dev` and `ztmf-prod` accounts before merge.
- [x] `terraform plan -var-file=tfvars/dev.tfvars` shows the cert-rotation resources adding cleanly and `ztmf_api` retention updating to 731.
- [x] `terraform plan -var-file=tfvars/prod.tfvars` likewise.
- [x] After merge, dev orchestration applies; verify Lambda is active in dev account, watching the bucket.
- [x] Drop the multi-SAN cert bundle into `s3://ztmf-cert-rotation-dev/dev/{cert,key,chain}.pem`; confirm Slack receives a `[DRY RUN]` success message.
- [x] Repeat for prod after the prod ACM ARN is imported and SSM is seeded.

## Refs

- #292 (Lambda implementation, merged 2026-04-28)
- #283 (original incident motivating the Lambda, closed)

### DIFF
--- a/backend/cmd/lambda-cert-rotation/README.md
+++ b/backend/cmd/lambda-cert-rotation/README.md
@@ -16,9 +16,15 @@ Uploads are expected at:
 
 ## Event behavior (S3 notifications)
 
-S3 can emit one event per uploaded object, so uploading all 3 files quickly may trigger **3 separate Lambda runs**. This is expected.
+S3 emits one `ObjectCreated` event per uploaded object, so uploading all 3 files triggers 3 Lambda invocations. **Only the invocation triggered by `chain.pem` performs the rotation work.** The `cert.pem` and `key.pem` invocations exit silently. This guarantees:
 
-The ACM `ImportCertificate` call is idempotent for a fixed certificate ARN, but you may see duplicate Slack success messages if all 3 events observe "all files present" around the same time.
+- Exactly one Slack notification per real rotation, regardless of upload order.
+- No concurrent races on validation, ACM import, Secrets Manager backup, or archival.
+- The `chain.pem` upload is the canonical commit point of a new bundle.
+
+**Required upload order:** upload `chain.pem` last. If `chain.pem` is uploaded before the other two files, the bundle-completeness check exits cleanly (other two missing) and no rotation runs. Re-upload `chain.pem` after the other two files are in place to fire the rotation.
+
+The ACM `ImportCertificate` call remains idempotent for a fixed certificate ARN, so re-running the rotation by re-uploading `chain.pem` against an already-current bundle is safe.
 
 ## Bundle freshness window
 
@@ -30,7 +36,7 @@ This rule exists to defend against a specific failure mode: a prior rotation tha
 
 - Always upload all three files within the same `aws s3 cp` session. Sequential uploads are fine; the window is wide enough for normal retry scripts.
 - If you need to replace a single file (typo, wrong version), delete and re-upload **all three** rather than only the one you want to change.
-- If the Slack alert says the bundle "spans X across LastModified timestamps", delete every `*.pem` at the prefix and re-upload from scratch:
+- If the Slack alert says the bundle "spans X across LastModified timestamps", delete every `*.pem` at the prefix and re-upload from scratch. Upload `chain.pem` last, since it is the trigger for the rotation:
 
   ```bash
   aws s3 rm "s3://ztmf-cert-rotation-<env>/<env>/" --recursive --exclude "*" --include "*.pem"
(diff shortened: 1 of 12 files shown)


MackOverflow (dismissed):
This looks pretty good Danny, some really good fixes here, and everyhting is responding as it should. One note:

 `data "aws_ssm_parameter" "ztmf_acm_arn"` is unconditional — this is a hard prereq for all terraform operations now**

Previously `data "aws_acm_certificate" "ztmf"` resolved dynamically from ACM. Now `data.aws_ssm_parameter.ztmf_acm_arn` is always evaluated — it's consumed by CloudFront and ALB which are always-on resources. Any `terraform plan` or `terraform apply` will fail with a data source error if the SSM parameter hasn't been seeded yet. The pre-merge runbook documents this, but it means:
- A net-new environment stands up with terraform in a broken state until an operator manually runs `aws ssm put-parameter`
- CI pipelines (if any) running plan against a scratch account need the parameter pre-populated

This is an acceptable trade-off for pinning over non-deterministic cert lookups, but the runbook should be promoted to a `README` or `CONTRIBUTING` note, not just the PR description.

MackOverflow (approved):
Looks great Danny!!

PR #305, opened 2026-05-06
Add impl environment alongside dev/prod in shared dev account

danielbowne:
## Summary

- Adds a third Terraform environment \`impl\` for business-owner experimental testing
- Lives in the dev AWS account (451245779631), shares dev's CMS-provisioned VPC, isolated from dev/prod via per-resource suffix renaming and account-singleton sharing through data sources
- Net effect on dev/prod: zero AWS resource changes; all HCL touches resolve as state-only \`moved\` migrations
- Adds new orchestration workflow that triggers on push to the long-lived \`impl\` branch and runs Analysis on PRs against \`impl\`

## Strategy

- **Suffix namespace** for resources whose AWS object name can carry an env suffix (RDS cluster, ALB, ECS cluster, security groups, log groups, per-env secrets). \`local.name_suffix\` renders empty for dev/prod (preserving existing AWS object names) and \`-impl\`/\`_impl\` for impl.
- **Count-gate** account-level singletons (ECR repos, OIDC provider, GitHub Actions IAM role, account-shared S3 logs bucket, SMTP/TLS secrets, Slack webhook, CloudFront OAC, HSTS policy, VPC endpoints + endpoint SG). dev/prod manage them; impl reuses them via data source fallbacks abstracted behind locals.
- **\`moved.tf\`** rewrites the address shifts from \`count\` adoption so dev/prod plans show only state-only \`moved\` entries, no destroy/create.
- **Snowflake / CFACTS-Snowflake EventBridge schedules** disabled for impl via the \`state\` attribute; lambdas themselves remain ungated to avoid extensive moved-block churn. Kion rotation held off via \`kion_rotate_schedule_enabled = false\` in tfvars.
- **\`prevent_destroy\` lifecycle** on \`ztmf_va_trust_provider\` and \`ztmf_db_user\` secrets across all envs. These hold operator-seeded values terraform never tracks (OIDC JSON, Aurora master_username); \`terraform destroy\` would orphan them.

## CI

- New \`orchestration-impl.yml\` triggers on push and pull_request against the \`impl\` branch
- Push to impl runs full pipeline (analysis -> diff -> backend -> infrastructure)
- PR to impl runs Analysis only (no deploy; avoids racing concurrent impl PRs for shared state)
- \`backend.yml\` gains a \`PARAMETER_NAME\` assertion before SSM put-parameter so an IMPL environment misconfigured with dev's parameter name fails fast instead of overwriting dev's image pointer

## Out-of-band prereqs (already executed in dev account before this PR)

[...]

PR #307, opened 2026-05-07
Import CMS-managed impl CloudFront, fix RDS secret tag-prefix collision

danielbowne:
## Summary

- Imports the CMS-Cloud-provisioned impl CloudFront distribution (`E1AO1UMW0EY8PK`, ticket [CLDSPT-106119](https://jiraent.cms.gov/plugins/servlet/desk/portal/22/CLDSPT-106119)) into impl's terraform state via an env-gated `import {}` block.
- Fixes a real, dev-impacting bug: dev's `DB_SECRET_ID` was about to be rewritten as a corrupt concatenation of dev's and impl's RDS-managed secret ARNs because the Secrets Manager `tag-value` filter does prefix matching and `cluster:ztmf` is a prefix of `cluster:ztmf-impl`.
- Bumps the GitHub Actions infrastructure-job role duration from 15 min to 1 hour so first impl bootstrap apply has time to complete.

## CloudFront import

`cloudfront:CreateDistribution` is denied in this org by SCP `p-q0yh3mfg`. `UpdateDistribution` is allowed. CMS Cloud created an empty distribution shell; the import block brings it under management and the next impl apply attaches aliases, the ALB origin, cache behaviors, the cert, and WAF via UpdateDistribution.

## RDS secret resolution

Before:
```hcl
db_cred_secret = join("", data.aws_secretsmanager_secrets.rds.arns)
```

The data source filtered by tag-value `arn:...:cluster:ztmf`. AWS Secrets Manager treats that as a prefix, so once impl was provisioned the filter started returning both clusters' secrets and `join("")` produced a corrupt ARN string. Dev's next apply would have written that into the ECS task definition's `DB_SECRET_ID` env var.

After:
```hcl
db_cred_secret = aws_rds_cluster.ztmf.master_user_secret[0].secret_arn
```

The cluster resource is the authoritative source for the auto-generated secret ARN. No Secrets Manager lookup, no prefix-collision exposure.

## Verification

- `terraform validate` passes
- `tflint --chdir=infrastructure/` clean
- `terraform plan -var-file=tfvars/dev.tfvars` against live dev state: **No changes. Your infrastructure matches the configuration.**

## Test plan

- [x] terraform plan against dev shows zero diff after the master_user_secret refactor
- [x] tflint + fmt clean (now also enforced by the pre-push hook from #305)
- [ ] PR auto-deploys to dev via CI; confirm dev ECS service stays healthy after apply
- [ ] After merge, push main into impl branch; impl apply imports the CloudFront distribution and configures aliases/origins/cache behaviors via UpdateDistribution
- [ ] Capture the now-managed `cloudfront_distribution_domain` output for the impl DNS CNAME

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -29,7 +29,11 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v4
         with:
           role-to-assume: ${{ secrets.ROLEARN }}
-          role-duration-seconds: 900
+          # 1 hour: dev/prod applies normally complete in a few minutes, but
+          # first-time impl bootstrap takes 15+ minutes (Aurora cluster create,
+          # ECS service stabilization, CloudFront propagation) and the previous
+          # 900s window expired mid-apply, leaving an inconsistent state.
+          role-duration-seconds: 3600
           aws-region: us-east-1
 
       - name: Setup Terraform
--- a/infrastructure/cloudfront.tf
+++ b/infrastructure/cloudfront.tf
@@ -62,6 +62,23 @@ resource "aws_cloudfront_vpc_origin" "internal_alb" {
   }
 }
 
+// CMS Cloud manages CloudFront creation centrally (cloudfront:CreateDistribution
+// is denied by SCP p-q0yh3mfg in this org). The dev/prod distributions were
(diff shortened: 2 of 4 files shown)


jsos3-cms (approved):
LGTM! Appreciate you putting this together.  Appreciate the comments and contexts inline!

Issue #314, opened 2026-05-14
Database: Remove legacy ADMIN and READONLY_ADMIN role values (follow-up cleanup)


danielbowne:
### Feature Type

Database Component

### Workflow Description

Follow-up cleanup release. Remove the legacy `ADMIN` and `READONLY_ADMIN` values from the role validation map and from every controller helper that still recognizes them. This issue should only land after the multi-OpDiv role swap (CMS-Enterprise/ztmf#312) and predicate flip (CMS-Enterprise/ztmf#313) have soaked in prod for at least one release cycle with zero rows carrying the legacy values.

Specifically:

- Remove `"ADMIN"` and `"READONLY_ADMIN"` from `backend/internal/model/validations.go` `roles` map.
- Drop the legacy values from `User.HasUnscopedRead`, `User.IsAdmin`, and `User.IsReadOnlyAdmin` in `backend/internal/model/users.go`.
- Update the local-dev auto-create middleware (`backend/cmd/api/internal/auth/middleware.go`) so newly-seen JWT identities are stamped with the equivalent new-tier role rather than `ADMIN`.
- Update test data fixtures (`backend/_test_data.sql`, `backend/_test_data_empire.sql`) and Emberfall expectations (`backend/emberfall_tests.yml`) to use the new role names instead of `ADMIN` / `READONLY_ADMIN`.
- Update the OpenAPI spec (`backend/openapi.yaml`) to drop the legacy values from the role enum.
- Validation test (`backend/internal/model/validations_test.go`) flips the legacy-role assertions from valid-to-invalid.
- Optional: add a `CHECK` constraint or database-level guard that rejects any new write with a legacy role value, as defense in depth.

### Acceptance Criteria

- [ ] `isValidRole("ADMIN")` returns false
- [ ] `isValidRole("READONLY_ADMIN")` returns false
- [ ] No production row carries a legacy role value at the moment of merge (pre-flight: `SELECT count(*) FROM users WHERE role IN ('ADMIN','READONLY_ADMIN')` returns 0)
- [ ] All test fixtures use new-tier role names
- [ ] OpenAPI role enum no longer lists `ADMIN` or `READONLY_ADMIN`
- [ ] Local-dev auto-create middleware emits a new-tier role
- [ ] CI / E2E pass with the updated fixtures
- [ ] Down migration restores both legacy values in the validation map and in helpers (in case a rollback to the prior release is needed during the soak window)

### Additional Context

[...]

Issue #313, opened 2026-05-14
Backend: Surface OpDiv and identity_provider fields and add scope predicates


danielbowne:
### Feature Type

General Feature

### Workflow Description

Surface the new OpDiv and identity-provider fields on the API, add the role-aware scope predicates on list and detail endpoints, ship the new `GET /api/v1/opdivs` reference endpoint, and write-gate `opdiv_id` on FISMA system creation. This is the API-layer step in the larger multi-OpDiv migration (epic: CMS-Enterprise/ztmf-misc#198). After this issue, real users (Danny, Allie, etc.) who became `OWNER` in the prior role swap regain their admin endpoint access, OpDiv-scoped admins see only their granted OpDivs, and the frontend can fetch the OpDiv list for selectors.

Specifically:

- Add `IdentityProvider string` and `AssignedOpDivIDs []*int32` to the `User` struct in `backend/internal/model/users.go`. Load both via subquery on per-user detail fetches; list responses can omit `AssignedOpDivIDs` to avoid the extra read cost. The `*User.Save()` insert path defaults `identity_provider` to `okta` when the caller does not provide a value (existing CMS admin-panel flow keeps working unchanged); update path leaves `identity_provider` immutable.
- Add `OpDivID *int32` to the `FismaSystem` struct. Append `opdiv_id` to `fismaSystemColumns`. The insert path honors a caller-provided value and falls back to the CMS subquery (`SELECT opdiv_id FROM opdivs WHERE code = 'CMS' AND active = TRUE LIMIT 1`); update path leaves `opdiv_id` immutable.
- New role helpers on `*User`: `IsOwner`, `IsHHSAdmin`, `IsOpDivAdmin`, `HasUnscopedRead`, `IsAssignedOpDiv`, `CanAccessFismaSystem`. Expand `IsAdmin` / `IsReadOnlyAdmin` / `HasAdminRead` to recognize both the new tier names and the legacy values during transition.
- Add an OpDiv scope predicate to `FindFismaSystems` (list query): unscoped admin tiers see every system; OpDiv-scoped admins see only systems in their granted OpDivs; ISSO/ISSM see only systems they are explicitly assigned to. Fail closed: if a caller is OpDiv-scoped but has no grants in `users_opdivs`, return zero rows rather than falling through to "no filter".
[...]

Issue #312, opened 2026-05-14
Database: Migrate users.role to multi-OpDiv role taxonomy


danielbowne:
### Feature Type

Database Component

### Workflow Description

Migrate the `users.role` column from the legacy single-admin model onto the multi-OpDiv role taxonomy. This is the role-swap step in the larger multi-OpDiv migration (epic: CMS-Enterprise/ztmf-misc#198). Predicates in the API layer are not flipped in this issue - the legacy `ADMIN` and `READONLY_ADMIN` values stay valid in `isValidRole` so every controller that still recognizes them continues to operate exactly as it does today. The follow-up backend issue surfaces the new role tiers in the API.

Specifically:

- Widen `users.role` from `VARCHAR(20)` to `VARCHAR(30)` so the longest new tier name (`OPDIV_READONLY_ADMIN`, 20 characters) plus headroom fits.
- Update the role validation map in `backend/internal/model/validations.go` to accept the new constants alongside the legacy values: `OWNER`, `HHS_ADMIN`, `HHS_READONLY_ADMIN`, `OPDIV_ADMIN`, `OPDIV_READONLY_ADMIN`, in addition to the existing `ADMIN`, `READONLY_ADMIN`, `ISSO`, `ISSM`.
- One-time data migration: every current `ADMIN` row becomes `OWNER`, every current `READONLY_ADMIN` row becomes `HHS_READONLY_ADMIN`. ISSO and ISSM are unchanged. The migration should log the row counts via `RAISE NOTICE` so the post-deploy verification is one log line.
- Down migration reverses the swap (`OWNER → ADMIN`, `HHS_READONLY_ADMIN → READONLY_ADMIN`).

### Acceptance Criteria

- [ ] `users.role` column type is `VARCHAR(30)`
- [ ] `isValidRole` recognizes all five new tiers plus the two legacy values
- [ ] One-time migration swaps every `ADMIN` to `OWNER` and every `READONLY_ADMIN` to `HHS_READONLY_ADMIN`
- [ ] Post-migration verification: `SELECT role, count(*) FROM users WHERE deleted = FALSE GROUP BY role` shows zero rows with legacy role values
- [ ] `RAISE NOTICE` line logs the row counts for each swap (use the PostgreSQL connection's `OnNotice` hook on the application side so the log line surfaces in the API log group, not just the DB log)
- [ ] Down migration reverses both swaps cleanly
[...]

Issue #311, opened 2026-05-14
Database: OpDivs reference table and OpDiv ownership on FISMA systems


danielbowne:
### Feature Type

Database Component

### Workflow Description

Introduce the OpDiv dimension on the database side. This is the data-prep step in the larger multi-OpDiv migration (epic: CMS-Enterprise/ztmf-misc#198). No user-visible behavior change in this issue - the new columns and tables exist and are backfilled, but every existing CMS user, system, and access pattern continues to operate exactly as it does today. Subsequent issues flip the role enum and the API scope predicates.

Specifically:

- Create a new `public.opdivs` reference table seeded with the current HHS Operating Divisions per the authoritative list at https://www.hhs.gov/about/organization/divisions/index.html, plus the HHS parent row. Names match hhs.gov spellings; only CMS uses `&` in its official name.
- Add `fismasystems.opdiv_id` as a foreign key to `opdivs.opdiv_id`. Backfill every existing row to CMS unconditionally (every pre-multi-tenant FISMA system is by definition CMS-owned; CFACTS `component_acronym` was considered as a derivation key but those values are CMS-internal organizational codes (OEDA, CCIIO) rather than OpDiv codes, so they cannot be used). Then enforce `NOT NULL` with an index for scoped list queries.
- Create a new `public.users_opdivs` junction table to track multi-OpDiv membership: `(userid, opdiv_id)` composite primary key, `granted_at` timestamp, `granted_by` user id with `ON DELETE SET NULL` so the audit trail survives a grantor leaving.
- Seed `users_opdivs` with a CMS grant for every existing user so day-1 visibility matches day-0 for everyone currently in the database.

Migrations should be reversible (`up` and `down`), idempotent (`IF NOT EXISTS` / `IF EXISTS` guards), and split into small numbered files so partial rollback is possible. Each migration runs in its own transaction per tern's default behavior; data-only migrations should defensively check for required referent rows (e.g. raise `EXCEPTION` if the seeded CMS row is missing) so a misconfiguration surfaces as a clear failure rather than a silent NULL.

### Acceptance Criteria

- [ ] `public.opdivs` table exists with `opdiv_id`, `code`, `name`, `is_parent`, `active`, `created_at` columns and a case-insensitive unique index on `code` (filtered to `active = TRUE`)
[...]

PR #316, opened 2026-05-18
Backend: multi-OpDiv schema foundation and scope predicates

danielbowne:
## Summary

First step toward HHS-wide multi-OpDiv support tracked under ztmf-misc#198.
Adds an OpDiv dimension to users and FISMA systems without changing CMS
behavior. Day-1 every existing system carries an OpDiv ownership row and
every existing user carries the matching grant, so CMS continues to operate
exactly as it does today.

Bundles three closely-coupled sub-issues so review can reason about the data
model and the predicate updates together:

- #311 - Database: OpDivs reference table and OpDiv ownership on FISMA systems
- #312 - Database: Migrate users.role to multi-OpDiv role taxonomy
- #313 - Backend: Surface OpDiv and identity_provider fields and add scope predicates

Stage D legacy role cleanup (#314) stays on a separate later branch per plan,
so we keep a recovery window if anything downstream still emits the legacy
role names.

## Merge order

This PR must merge **before** any companion work in the `ztmf-ui` frontend
repo. The UI work depends on the new `opdiv_id`, `identity_provider`, and
`assignedopdivids` fields surfaced here, and on the new `/api/v1/opdivs`
endpoint. Merging the frontend first would break dev. Coordinate with the
frontend track so their PR opens against an updated `main` after this one
deploys.

## What changes

**Schema (migrations 0026 - 0036)**

- New `opdivs` reference table seeded with HHS + 11 OpDivs.
- New `fismasystems.opdiv_id` FK (NOT NULL after backfill to CMS).
- New `users.identity_provider` column (NOT NULL after backfill to `okta`).
- New `users_opdivs` junction with CMS grant seeded for every existing user.
- `users.role` widened to `VARCHAR(30)` and swapped to the new taxonomy:
  - former `ADMIN` rows -> `OWNER`
  - former `READONLY_ADMIN` rows -> `HHS_READONLY_ADMIN`
  - ISSO / ISSM unchanged
- Legacy `ADMIN` / `READONLY_ADMIN` values remain valid through Stage D so
  any downstream system still emitting them keeps working during transition.

**Model / controller**

- New role helpers on `*User`: `IsOwner`, `IsHHSAdmin`, `IsOpDivAdmin`,
  `IsReadOnly`, `HasUnscopedRead`, `IsAssignedOpDiv`, `CanAccessFismaSystem`.
- `User.AssignedOpDivIDs` populated from `users_opdivs`.
- `FismaSystem.OpDivID` surfaced on the API; `Save()` defaults to CMS via
  subquery so legacy POSTs without `opdiv_id` continue to work.
- `ListFismaSystems` scopes by role tier. OpDiv-scoped admins use the
  `RestrictToOpDivIDs` flag so an admin with zero grants fails closed rather
  than falling through to an unscoped read.
- `GetFismaSystem` gates via `CanAccessFismaSystem` after fetch so NotFound
  stays NotFound (no existence leak via 403).
- `SaveFismaSystem` write-gate prevents an OPDIV_ADMIN from creating or
  moving a system into an OpDiv they have no grant for.
- Mass emails now route ALL recipients through the unscoped admin tier
  (OWNER / HHS_ADMIN / OPDIV_ADMIN / legacy ADMIN) so the existing
  notification path keeps the same audience semantics.

**API**

- New endpoint: `GET /api/v1/opdivs` (read-only list, all roles).
- `openapi.yaml` updated: OpDiv schema, role enum, `opdiv_id`,
  `identity_provider`, `assignedopdivids`, new endpoint, Stage D
  transition note.

**Tests / data**

- Unit tests for all role helpers (11 roles x 7 helpers) and
  `CanAccessFismaSystem` (16 cases).
- Emberfall smoke for `/opdivs` and the new fields on `/users/current`.
- `_test_data.sql` and `_test_data_empire.sql` updated with explicit column
  lists, EMPIRE OpDiv, sequence reset, and matching `users_opdivs` grants.

**Misc**

- `pgx` `OnNotice` handler so `RAISE NOTICE` from migrations surfaces in
  app logs.
- `.playwright-mcp/` added to `.gitignore`.

## Pre-merge action required

Snowflake DDL must be applied against dev Snowflake BEFORE this PR merges,
or the next scheduled `ztmf-data-sync` Lambda run will fail on missing
columns. Same failure mode as the 2026-05-04 `REACTIVATED_BY` regression.

Apply against `BUS_ZEROTRUST.CORE`:

    ALTER TABLE ZTMF_FISMASYSTEMS ADD COLUMN IF NOT EXISTS OPDIV_ID NUMBER(38,0);
    ALTER TABLE ZTMF_USERS ADD COLUMN IF NOT EXISTS IDENTITY_PROVIDER VARCHAR(50);

Run `DESCRIBE TABLE BUS_ZEROTRUST.CORE.ZTMF_USERS` first to confirm the
table exists; if missing, create from the local `ZTMF_TABLES.sql` template
before the ALTER.

## Snyk policy ignore (CI unblock)

`backend/.snyk` adds a scoped ignore for
`SNYK-GOLANG-GITHUBCOMAPACHETHRIFTLIBGOTHRIFT-16323410`, a high-severity
finding in `apache/thrift` v0.23.0. Important context for the reviewer:

- The package is a transitive dependency only, introduced via
  `snowflakedb/gosnowflake` and `apache/arrow-go`. ZTMF code does not
  import `thrift` directly.
- The same vulnerable version is already deployed in dev and prod today,
  so the finding was not introduced by anything in this PR. The only
  effect of leaving the gate red is that no change can reach dev for
  validation before being merged toward prod.
- Upstream `apache/thrift` v0.24.0 (the actual fix) has not been
  released. Both `arrow-go` and `gosnowflake` still declare a v0.22.0
  dep, so forcing a manual override would drift us out of sync with the
  upstream graph rather than closer to a fix.
- The ignore is scoped to this single Snyk ID with a 90-day expiry
  (2026-08-19). Every other Snyk finding still fails the build. Revisit
  when `apache/thrift` v0.24.0 ships or when `arrow-go` cuts a release
  pinning to it, whichever comes first.

## Out of scope (tracked separately)

- Stage D legacy role removal - #314
- OpDiv scope on scores / datacalls / events endpoints
- Grant management endpoints (admin panel reads / writes for users_opdivs)
- Cross-OpDiv ISSO assignment trigger
- Frontend OpDiv badges and OpDiv-aware admin panel (ztmf-ui follow-up PR)
- PG -> Snowflake Lambda field surfacing

## Test plan

- [x] `make dev-setup` - all migrations apply clean on empty DB
- [x] `make test-unit` - new role helper and access tests pass
- [x] `make test-full` - Emberfall E2E green including new fields
- [ ] Reviewer: apply Snowflake DDL against dev BEFORE merge
- [ ] Reviewer: confirm CMS user `GET /api/v1/fismasystems` returns the
      same set of systems as on `main`
- [ ] Reviewer: confirm `GET /api/v1/users/current` carries
      `identity_provider` and `assignedopdivids`
- [ ] Reviewer: confirm `GET /api/v1/opdivs` returns HHS + 11 OpDivs
- [ ] Open companion ztmf-ui PR ONLY after this merges and deploys to dev


jsos3-cms (dismissed):
Looks good to me! Thanks for getting this out!

Is there ever a case where there may be HHS and another division accidentally set as Parents, that aren't? I could see a far future where this tool serves multiple orgs (HHS, DOI, ETC) , but we're not there yet - How do we want to handle a potential insert of parents and what could that do to the schema? Really appreciate the thorough tests you did here :)

MackOverflow (dismissed):
This is looking awesome Danny! One flag that popped up for me, not sure if this is an issue to think about or not: 
on backend/internal/model/users.go — IsHHSAdmin() method:
This function returns true for both HHS_ADMIN and HHS_READONLY_ADMIN. The name reads like it's asking "is this a full HHS admin?" rather than "is this user in the HHS tier?" That could trip up a future caller who reaches for IsHHSAdmin() expecting a write-access gate and doesn't notice it also covers read-only. Since neither controller is actually calling this yet, might be worth renaming to IsHHSTier() or IsHHSLevel() before it gets more callers — much easier to rename before it spreads. Or if there's a good reason to keep the name, a comment spelling out the "includes read-only" intent would help.

danielbowne:
@MackOverflow good catch, renamed in cd9e8fd. Both `IsHHSAdmin` and `IsOpDivAdmin` had the same shape (each covered the read-only variant too), so I renamed both to `IsHHSTier` / `IsOpDivTier` for consistency. Added doc comments on both spelling out that they are tier-membership checks, not write gates. Write-versus-read gating still goes through `IsAdmin` / `HasAdminRead` as before.

danielbowne:
@jsos3-cms good question. The `opdivs` table is intentionally flat: `is_parent` is a label, not a structural pointer. There is no `parent_opdiv_id` FK, so the auth layer never walks a hierarchy and a fat-fingered second `is_parent = TRUE` row would not break anything that exists today. Horizontal expansion to a second top-level org (DOI, etc.) is just inserting rows.

The real work for true multi-department reuse would be in the roles enum (`HHS_ADMIN` is hardcoded), not the table, and that is a follow-up if it ever lands.

MackOverflow (dismissed):
Looks great Danny!

MackOverflow (approved):
Still looking good Danny. Lets just make sure this is getting documented to be shelf-stable and ready for handoff.

PR #318, opened 2026-05-20
Align CMS scoring with HHS scale (1.0-5.0) and surface tier in API

danielbowne:
## Summary

Aligns CMS scoring with the HHS scorecard scale (1.0 to 5.0) and moves all
maturity tier logic to the backend so the API response is the single source
of truth for both numeric score and tier label. Tracked under ztmf-misc#175.
This PR ships the backend portion only; the matching frontend PR opens
against ztmf-ui on the same branch name and merges in a tight window after
this one deploys to prod.

Once this merges to main and auto-deploys to dev, both prod and dev will
serve scoring numbers side by side and the customer can compare directly.

## Math change

Each stored answer (integer 1-4 per question) shifts by +1 at pillar
aggregation time so Traditional contributes 2.0, Initial 3.0, Advanced 4.0,
Optimal 5.0. A missing answer for a question that was expected in the
system's environment contributes 1.0, so partial completion drags pillar
scores down. Pillar score is the average of those shifted values for the
questions expected in that pillar.

System score is the simple average of pillar scores (equal weighting).
The divisor follows the actual pillar count via Postgres `AVG` so adding
or removing a pillar does not silently shift the math. Real-data shape:
six pillars per system today (Devices, Applications, Networks, Data,
CrossCutting, Identity), confirmed against the production volume.

Tier thresholds match the HHS scorecard:

- Optimal: `>= 4.1`
- Advanced: `>= 3.1`
- Initial: `>= 2.1`
- Traditional: `>= 1.01`
- Not Assessed: everything below 1.01

These boundaries are identical at the pillar level and the system level.

## API surface

`/api/v1/scores/aggregate` keeps the same shape with two additions and one
semantic shift on existing fields:

- `data[].systemscore` now holds an HHS-scale float in 1.0 to 5.0 (was 1.0
  to 4.0)
- `data[].systemtier` is a new string field at the top level
- `data[].pillarscores[].score` is on the same HHS scale
- `data[].pillarscores[].tier` is a new string field on each pillar entry

Tier values are exactly one of `Optimal`, `Advanced`, `Initial`,
`Traditional`, `Not Assessed`. The backend is the only place the predicate
runs.

`openapi.yaml` is updated with the new fields, including enum values, and
the score-field descriptions call out the new scale.

## Universe of returned rows

A (system, datacall) pair appears in `/scores/aggregate` only if at least
one score row exists for it in the `scores` table. That preserves the
legacy aggregate contract: systems that never started a data call do not
appear (the frontend cross-references `/fismasystems` to render those as
"No Score"). Pillar-level Not Assessed is still reachable for partial
systems whose specific pillars have zero answered questions.

The aggregation does not filter on `fismasystems.decommissioned` so
historical aggregates for retired systems remain visible.

## Real-data validation

Verified against the production data volume (260 systems, 28k score rows
across three real data calls):

- Zero tier downgrades across 1,494 pillar records and 249 system records.
- ~60% of records move up exactly one tier under the HHS scale, ~36% stay
  put, ~7% are unanswered (rename only from "No Score" to "Not Assessed").
- No system score lands above 5.0 or below 1.0.
- batCAVE (CMS-Cloud-AWS, decommissioned with the proper feature) returns
  3.26 Advanced as expected.

The audit artifacts that drove this verification live under
`tmp/snowflake-recon/` for the env-cleanup auditor.

## Pre-merge action: legacy datacenter environment cleanup

A small set of systems in production were archived using the pre-real-
decommission flow that overwrote `datacenterenvironment` to the sentinel
string `'DECOMMISSIONED'`. The new aggregation needs to know the real
environment to enumerate the expected question catalog, so those systems
return empty aggregates today.

Fix is data-only and lives at `tmp/fix-legacy-env-sentinel.sql`. It restores
the real env value (derived from score-history evidence or CFACTS hosting
fields, depending on what's available), and fixes two systems whose
`fismauid` was overwritten to the literal string `'Retired'`. Apply in dev,
smoke-test, then apply in prod. The SQL is idempotent and wrapped in a
transaction with a verification select.

Two systems (HITECH, MMS-dep) have no usable source to determine the right
env and are intentionally left alone. They have no score history so they
do not appear in the aggregate either way.

## Test coverage

| Layer | What breaks if you touch X |
|---|---|
| `Tier()` boundary tests | wrong threshold value |
| `aggregatePillarRows()` unit tests | Go-side group/rollup logic, divisor-follows-count |
| `buildPillarScoresSQL()` unit tests | wrong filter shaping for ISSO + specific-system intersect |
| `TestFindScoresAggregateIntegration` | wrong SQL math, missing tier fields, `Tier(score)` disagreement |
| `TestScoreSaveValidationIntegration` | `validate()` notes-length and deadline gates |
| `TestCopyPreviousScoresIntegration` | data-call rollover regression |
| Existing controller auth tests | scoring endpoint auth gates |
| Emberfall scored-system smoke | 5xx in the new SQL, broken JSON serialization |
| Emberfall unscored-system smoke | cardinality regression (cartesian over fismasystems x datacalls) |

`make test-full` runs all of the above as steps 2 through 5.

## Out of scope

- Frontend changes ship in the companion `feature/hhs-scoring-alignment`
  branch on ztmf-ui. The shape contract there is to delete the hardcoded
  threshold helpers in `PillarScoresModal.tsx` and `FismaTable.tsx` and
  read `tier` from the API response. Tooling for the frontend Claude is
  set, just waiting on backend to deploy to dev first.
- Snowflake export shape. Per the customer's recent direction the SDL
  toggle is per-system with CMS as the only opt-in population, so the
  raw `scores` rows that get synced are unaffected by this change.
- Historical scale snapshot. Easy to recompute and the new path is
  reversible by `git revert`; deferred unless someone asks for it.

## Test plan

- [x] `make test-full` passes all 5 stages locally with sourced env
- [x] Live API on local dev DB returns expected aggregates for known
      systems (RFX 4.09 Advanced, MACPro and ECWS at their AWS-scale
      values, batCAVE 3.26 Advanced)
- [ ] Reviewer: apply `tmp/fix-legacy-env-sentinel.sql` in dev, confirm
      the four legacy-archive systems return historical aggregates
- [ ] Reviewer: hit `/api/v1/scores/aggregate?fismasystemid=1001&include_pillars=true`
      against dev after auto-deploy, confirm `systemtier` and per-pillar
      `tier` fields render
- [ ] Reviewer: confirm an unscored fismasystemid returns `{"data":[]}`
- [ ] Open companion ztmf-ui PR ONLY after this merges and deploys to dev
- [ ] Apply `tmp/fix-legacy-env-sentinel.sql` in prod before main merge,
      or as the first action immediately after

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -299,19 +299,22 @@ test-build:
 test-full:
 	@echo "Running comprehensive test suite..."
 	@echo ""
-	@echo "1/4 Building all binaries (API + Lambdas)..."
+	@echo "1/5 Building all binaries (API + Lambdas)..."
 	@cd backend && go build ./cmd/api/...
 	@cd backend && go build ./cmd/lambda/...
 	@cd backend && go build ./cmd/lambda-cfacts-snowflake/...
 	@cd backend && go build ./cmd/lambda-cfacts-s3/...
 	@echo ""
-	@echo "2/4 Running unit tests..."
+	@echo "2/5 Running unit tests..."
 	@cd backend && go test -short ./...
 	@echo ""
-	@echo "3/4 Generating coverage report..."
+	@echo "3/5 Generating coverage report..."
 	@cd backend && go test -cover $$(go list ./... | xargs -I{} sh -c 'test -n "$$(find $$(go list -f "{{.Dir}}" {}) -maxdepth 1 -name "*_test.go" 2>/dev/null)" && echo {}' | tr "\n" " ")
 	@echo ""
-	@echo "4/4 Running Emberfall E2E tests (isolated containers)..."
+	@echo "4/5 Running integration tests (require DB_* env from backend/dev.compose.env)..."
+	@cd backend && go test -run Integration ./... -count=1
+	@echo ""
+	@echo "5/5 Running Emberfall E2E tests (isolated containers)..."
 	@make test-e2e
 	@echo ""
 	@echo "✅ All tests complete"
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -730,9 +730,24 @@ tests:
           data:
             <<: *restoredUserData
 
-  # Scores: HHS-aligned aggregate smoke test (ztmf-misc#175). Math is
-  # asserted in unit tests at internal/model/scores_test.go; this confirms
-  # the endpoint serves a parseable 200 response with the new shape.
+  # Scores: HHS-aligned aggregate smoke battery (ztmf-misc#175).
+  #
+  # Math correctness, response field presence, and Tier(score)==tier
+  # consistency are all pinned in internal/model/scores_integration_test.go,
+  # which runs against the real schema as part of `make test-full`.
+  # Emberfall's body assertion is exact-match (body.text is full-body
+  # equality, body.json equality fails on array-typed responses due to
+  # aquia-inc/emberfall#36), so these HTTP-layer smokes cover what we
+  # can express cleanly: cardinality contract, auth gates, and content
+  # type. Anything that survives unit + integration but breaks at the
+  # HTTP boundary still shows up here.
+  #
+  # The DS-1 seed (fismasystemid=1001) has 24 score rows across datacalls
+  # 1 and 2 in _test_data_empire.sql, guaranteeing a non-empty response.
+
+  # 1. Scored system returns a populated aggregate via the full HTTP
+  #    path. Regression target: 5xx anywhere in the new aggregation SQL,
+  #    JSON serialization of the new tier fields, content negotiation.
   - url: http://localhost:8080/api/v1/scores/aggregate?fismasystemid=1001&include_pillars=true
     method: GET
     headers:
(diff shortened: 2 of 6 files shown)


MackOverflow (commented):
We've gone back and forth on this a ton, but I'm still flagging this issue: 

1. Float boundary precision at 4.1, 3.1, 2.1

The PR explicitly explains the 1.01 floor (a fully-unanswered pillar lands at exactly 1.0 after the shift, so >= 1.01 correctly excludes it). But the same float concern applies to the other thresholds. 4.1, 3.1, and 2.1 cannot be represented exactly in IEEE 754 float64 — they're stored as 4.10000000000000035..., 3.10000000000000053..., etc. If Postgres AVG returns a value that "should be" exactly 4.1 but comes out as 4.0999999... due to floating-point arithmetic, a system would land in Advanced instead of Optimal.

In practice this risk is low — the denominators are small integers (≤6 pillars), so most averages are exact or very close — but it's worth confirming with a known-boundary system in dev smoke. The real-data validation you ran (RFX at 4.09 Advanced) doesn't exercise this edge.

Do we want to surface this to @eschwein-usds  to final approval?

danielbowne:
@MackOverflow good catch on the float boundary precision concern. Real bug, fixed in f0a943a.

Quick recap: IEEE 754 cannot represent 4.1, 3.1, or 2.1 exactly, so a system whose arithmetic landed at a value just below the literal representation (and would display to the user as `"3.10"` via `toFixed(2)`) was getting classified one tier lower than expected. A synthetic probe across 200,000 random 6-pillar combinations found 642 distinct mismatches where the displayed score and the assigned tier disagreed.

The fix moves the comparison into integer space: multiply by 100, round with `math.Round`, then compare against integer thresholds (`>= 410` for Optimal, `>= 310` for Advanced, etc.). That rounding step is the same one the frontend does for display, so the contract is now "whatever the user sees on screen is the number used to assign the tier." If they see `4.10` they get Optimal, full stop.

Re-running the same probe against the fixed predicate produces zero mismatches. New test cases in `TestTier` pin the pathological inputs (`3.0999999999999996`, `4.0999999999999996`) so the regression cannot silently re-introduce itself.

MackOverflow (approved):
Danny, I think this fix negates my question to Elizabeth, this is g2g.

Issue #310, opened 2026-05-13
Ops image workflow fails silently when ECR repo absent on first apply


danielbowne:
## Description

The `ops-image.yml` GitHub Actions workflow that builds and pushes the `ztmf/ops` container (used by the on-demand Fargate DB-access task introduced in #290) fails for any environment where the ECR repo `ztmf/ops` does not yet exist at the time the workflow runs. Because the workflow only re-triggers on changes to `backend/ops/**` or the workflow file itself, the failure persists until someone manually retriggers via `workflow_dispatch`.

This is what happened to PROD on 2026-04-17: the workflow ran when #290 landed on `main`, the DEV matrix leg succeeded, the PROD leg failed at the `ECR - Resolve Ops Repo URL` step because Terraform had not yet created the repo in the prod account. The SSM parameter `ztmf_ops_tag` was therefore never updated past the placeholder value `bootstrap`, and the Fargate ops task in prod could not pull an image (`pull image manifest has been retried 7 time(s): ... :bootstrap: not found`). DB access via `make db-shell-prod` / `make db-forward-prod` / `tmp/sync_prod_to_local.sh` was broken from 2026-04-17 until 2026-05-12, when the workflow was re-dispatched manually for PROD.

## Steps to Reproduce

1. Spin up a new ZTMF environment whose Terraform has not yet been applied (no `ztmf/ops` ECR repo, no `ztmf_ops_tag` SSM parameter yet, or that parameter is set to a placeholder).
2. Merge any commit to `main` that touches `backend/ops/**` or `.github/workflows/ops-image.yml`.
3. The matrix job for the new environment fails at `ECR - Resolve Ops Repo URL`.
4. Later, the Terraform apply creates the ECR repo, but nothing re-triggers the workflow.
5. The SSM parameter remains stale; the Fargate ops task cannot launch.

## Expected Behavior

The workflow either creates the ECR repo on miss before pushing, or fails in a way that re-runs successfully once the repo exists without requiring a code change to retrigger.

## Actual Behavior

Workflow fails at `aws ecr describe-repositories --repository-names ztmf/ops`, exits 254, never retries. SSM parameter stays at the placeholder value. Operators have no signal that DB access is broken until they try to use it weeks later.

## Environment

GitHub Actions, `.github/workflows/ops-image.yml`, matrix strategy across `DEV` and `PROD` environments. Affects any future env (e.g. `IMPL`) on first apply.

## Acceptance Criteria

[...]

PR #319, opened 2026-05-21
Embed last-edited audit fields on Score responses (refs ztmf-ui#310)

danielbowne:
## Summary

Adds `last_edited_at` + `last_edited_by` to every Score returned by the API so the dashboard can show "Last edited <when> by <who>" per question without calling `/events`. Authz follows the resource: if you can read the score, you see the editor. `/events` stays admin-only as the compliance firehose.

The frontend consumer is tracked in [CMS-Enterprise/ztmf-ui#310](https://github.com/CMS-Enterprise/ztmf-ui/issues/310).

## What ships

**Per-resource audit fields**
- New `AuditRef` struct + opt-in `Auditable` interface in `backend/internal/model/audit.go`. Reusable shape future resources adopt without re-litigating authz.
- `Score` gains `LastEditedAt *time.Time` and `LastEditedBy *AuditRef`. Both fields populated together or omitted together; the frontend never has to branch on each independently.
- `FindScores` joins the latest write event laterally + resolves editor via `LEFT JOIN users`. ISSO scope is preserved through the lateral join, verified by integration test.
- `Save` reads back the actual event row via `lookupScoreAudit` rather than synthesizing from `time.Now()` and ctx user. If `recordEvent` failed silently the response will not advertise a phantom editor, and the timestamp matches what subsequent GETs return.

**Read-through edit protection** (commit 2)
- The questionnaire UI PUTs unconditionally on every Next click. Without a guard a user walking past a question already answered by a prior cycle's reviewer would get stamped as the new editor and overwrite history.
- `Save` short-circuits when every answer field matches the existing row. No UPDATE, no event, returns the existing row with its existing audit intact. Defense in depth even if a client adds its own dirty check (the frontend has one in CMS-Enterprise/ztmf-ui).
- `scoresEqualForUpdate` is the pure comparison extracted so the equality rules can be unit-tested without a database.

**Performance**
- Migration `0037eventsauditindex` adds a partial expression index on `events ((payload->>'scoreid')::int, createdat DESC) WHERE resource='public.scores'` so the lateral does not degenerate to a sequential scan as the events table grows. Also adds a generic `(resource, createdat DESC)` index for future Auditable resources.

**OpenAPI**
- New `AuditRef` schema, new `last_edited_at` + `last_edited_by` on `Score`, role enum expanded to the multi-OpDiv taxonomy with a canary comment pointing to the canonical Go source so the spec does not drift again.

**Test data**
- Empire seed gets `datacallid=3` with a 2099 deadline so ISSO writes through the dashboard test path are not blocked by the past-deadline guard. Cross-referenced in `_test_data_empire.sql` and `emberfall_tests.yml`.

## Tests

- Unit: `TestScoreAuditInfo`, `TestDerefString`, `TestScoresEqualForUpdate` (covers identical match, each-field-differs, nil-vs-empty-notes equality, whitespace-not-empty contract, nil-input safety). Plus a static `var _ Auditable = (*Score)(nil)` interface conformance check.
- Integration (DB-coupled): `TestScoreSaveStampsAuditFieldsIntegration`, `TestFindScoresIncludesAuditFieldsIntegration`, `TestFindScoresISSOScopeRetainsAuditFieldsIntegration`, `TestScoreSaveNoOpPreservesPriorEditorIntegration`.
- Emberfall E2E: admin POST + GET + PUT lifecycle with audit assertion, ISSO own-system POST asserts editor identity, ISSO foreign-system POST returns 403.

`make test-full`: 94/94 emberfall + unit + integration all pass.

## Design notes (read-only)

- `/events` stays admin-only by design. Audit visibility on a per-resource basis is the user-facing surface; `/events` is the cross-resource compliance firehose. Different products.
- `recordEvent` continues to log-and-swallow errors. `lookupScoreAudit` is the canonical pattern for any caller that needs to confirm an event was actually written before stamping audit on a response. Documented in `events.go`.
- A small TOCTOU window exists between the SELECT in `scoreUpdateIsNoOp` and the (skipped) UPDATE. Practical consequence: the next GET corrects the audit fields. Wrapping `Save` in a transaction would close the window but rework every model that relies on `queryRow`'s auto-event hook, which is out of scope here.

## Follow-ups (separate issues)

1. Normalize `events.resource` from `public.scores` to `scores` (typo from PR #94, 2024-08-19). Cosmetic; partial index in this PR matches the literal.
2. Roll the same `LastEditedAt` + `LastEditedBy` shape onto FismaSystem, DataCall, User, MassEmail.
3. Resolve `FismaSystem.decommissioned_by` / `reactivated_by` from raw UUIDs into `AuditRef`.
4. Consider denormalized `updated_at` + `updated_by` columns once 4+ resources implement Auditable.

## Test plan

- [x] CI pipeline passes (unit, integration, Emberfall E2E)
- [x] Verify migration `0037` applies cleanly on dev deploy
- [x] Smoke against dev: GET `/api/v1/scores?fismasystemid=X&datacallid=Y` returns `last_edited_at` + `last_edited_by` on every row that has an event
- [x] As ISSO, POST a score to an assigned system; response carries the ISSO's identity in `last_edited_by`
- [x] As ISSO, POST a score to an unassigned system; 403
- [x] Re-PUT an identical score; events table count unchanged, prior editor preserved
- [x] Coordinate with CMS-Enterprise/ztmf-ui#310 frontend PR for the visible dashboard footer

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -90,6 +90,13 @@ INSERT INTO public.pillars VALUES (6, 'Identity', 0) ON CONFLICT DO NOTHING;
 -- Test DataCalls (Imperial Audits)
 INSERT INTO public.datacalls VALUES (1, 'FY2024 Imperial Security Review', '2024-01-01T00:00:00Z', '2024-12-31T23:59:59Z') ON CONFLICT DO NOTHING;
 INSERT INTO public.datacalls VALUES (2, 'FY2025 Death Star Assessment', '2025-01-01T00:00:00Z', '2025-03-31T23:59:59Z') ON CONFLICT DO NOTHING;
+-- Future-deadline cycle used by audit-field smoke tests (ISSO writes need
+-- a non-expired datacall so validate() does not trip the deadline guard).
+-- NOTE: backend/emberfall_tests.yml references datacallid=3 literally in
+-- the audit-fields block; if you reorder or renumber this row, update the
+-- four "datacallid: 3" references and the "?datacallid=3" query string
+-- in that file in lockstep.
+INSERT INTO public.datacalls VALUES (3, 'Audit Fields Smoke Cycle', '2026-01-01T00:00:00Z', '2099-12-31T23:59:59Z') ON CONFLICT DO NOTHING;
 
 -- Test FISMA Systems (Imperial Systems)
 -- Use explicit column names to work with initial schema
--- a/backend/cmd/api/internal/migrations/0037eventsauditindex.go
+++ b/backend/cmd/api/internal/migrations/0037eventsauditindex.go
@@ -1,0 +1,45 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		"add audit-read indexes on events",
+		`
+-- Per-resource audit fields (ztmf-ui#310) read events via a LATERAL
+-- subquery filtered on resource + payload->>'scoreid'. The events
+-- table previously had zero indexes; the lateral degenerates to a
+-- sequential scan per score row at any non-trivial event count.
+--
+-- Two indexes:
+--
+-- 1. events_score_audit_idx: partial expression index sized to the
+--    per-question-on-dashboard hot path. Scoped to resource='public.scores'
+--    so it stays small as other resources adopt the Auditable pattern,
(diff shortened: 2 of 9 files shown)


jsos3-cms (approved):
Good work on this! I like the test cases :)

PR #320, opened 2026-05-22
Parametrize local dev ports per ZTMF_ENV so impl runs alongside main

danielbowne:
## Summary

- Pulls all hardcoded local ports (API, postgres, test API, test DB, vite) and the docker compose project name out of the Makefile and compose-test.yml, routing them through ZTMF_ENV-selected fragments.
- New `Makefile.main` and `Makefile.impl` (checked in, mirror of the `tfvars/*.tfvars` pattern). `Makefile.local` (gitignored) for ad-hoc local tweaks.
- ZTMF_ENV auto-detects from the current git branch: \`impl\` / \`feature/impl-*\` -> impl, else main. Override explicitly with \`make ZTMF_ENV=impl dev-setup\`.

| | main | impl |
|---|---|---|
| API | 8080 | 8081 |
| postgres | 54321 | 54322 |
| test API | 8090 | 8091 |
| test postgres | 54399 | 54398 |
| vite | 5174 | 5175 |
| compose project | ztmf | ztmf-impl |

Frontend side: \`frontend-env\` now writes \`VITE_DEV_PORT=\$(FRONTEND_PORT)\`. \`ztmf-ui-impl/vite.config.ts\` already reads it, so \`yarn dev\` in the impl UI worktree picks up the offset automatically.

## Test plan

- [x] \`make dev-setup\` on \`feature/impl-multi-env-ports\` brings up \`ztmf-impl-{api,postgre}-1\` on 8081/54322 while the main worktree's \`backend-{api,postgre}-1\` keeps running on 8080/54321.
- [x] \`/api/v1/users/current\` returns data from both 8080 and 8081 with the same Empire JWT.
- [x] Pre-push hook ran \`make test-full\` (unit + integration + Emberfall E2E) with impl ports 8091/54398 - all green.
- [x] \`make ZTMF_ENV=main -np\` resolves to the historical defaults, so main worktree behaviour is unchanged.
- [x] Reviewer to spot-check that no other files reference the old literals (\`grep -rE '54321|localhost:8080' backend/ Makefile\`).


MackOverflow (approved):
Again, ack that this is impl: 


1. **Connection leak in `scoreUpdateIsNoOp` and `lookupScoreAudit`** — the missing `defer conn.Close(ctx)` is still worth checking regardless of which branch this is on. If those functions run in impl's dev environment, they'd still leak connections over time.

2. **`Tier()` float fix** — this PR has the better implementation (integer comparison via `math.Round`) vs. PR #318's direct `>= 4.1` float comparison. Worth making sure the fix carries over when impl eventually merges into main.

3. **The open reviewer checklist item** (`grep -rE '54321|localhost:8080' backend/ Makefile`) — still unchecked and still easy to run.

danielbowne:
Thanks for the review. Walking through each point:

> 1. **Connection leak in `scoreUpdateIsNoOp` and `lookupScoreAudit`** - the missing `defer conn.Close(ctx)` is still worth checking regardless of which branch this is on. If those functions run in impl's dev environment, they'd still leak connections over time.

Great catch, this was a real bug. Both functions acquire a `*pgx.Conn` via `db.Conn(ctx)` and never release it, leaking one connection per Score `Save` UPDATE. It existed in `main` too because both branches share the audit-fields code from PR #319. Tracking and fixing on `main` so impl inherits on the next merge:

- Issue: #321
- PR: #322 (adds `defer conn.Close(ctx)` to both functions, plus a small test-fragility fix that surfaced while validating)

> 2. **`Tier()` float fix** - this PR has the better implementation (integer comparison via `math.Round`) vs. PR #318's direct `>= 4.1` float comparison. Worth making sure the fix carries over when impl eventually merges into main.

I verified this with a byte-level diff of the function on `main` versus the head of this branch and the two are identical. `Tier()` was introduced on `main` via PR #318 (commit `72dfefa`) with the integer comparison version already in place. The floating-point version you remember was a draft revision during the conversation on PR #318 that got changed before it merged. Nothing needs to carry over here, both branches have the same code.

Quick way to confirm this in the future, run from a clone of `main`:

```bash
diff <(git show main:backend/internal/model/scores.go | sed -n '/^func Tier/,/^}/p') \
     <(git show origin/feature/impl-multi-env-ports:backend/internal/model/scores.go | sed -n '/^func Tier/,/^}/p')
# empty output = identical
```

> 3. **The open reviewer checklist item** (`grep -rE '54321|localhost:8080' backend/ Makefile`) - still unchecked and still easy to run.

This one is on the reviewer side of the PR template, it is one of the checks the reviewer runs locally before approving so the author knows the audit is complete. To run it yourself for any PR that touches local-dev wiring:

```bash
git fetch origin feature/impl-multi-env-ports
git checkout feature/impl-multi-env-ports
grep -rE '54321|localhost:8080' backend/ Makefile
```

If the grep returns no hits, the port parametrization is complete. If it returns hits, those are stragglers to flag back to the author. Running the reviewer checklist before clicking Approve is the small habit that catches issues like the connection leak in #1 before they merge, very worth building in.

## One thing on branch context

This PR is targeting `impl`, not `main`, which means the diff GitHub shows is everything between `feature/impl-multi-env-ports` and `impl`, not the difference from `main`. The `impl` branch is a long-lived experimental branch and is intentionally well ahead of `main` on the multi-environment work, so the file list looks scary (40+ files) but most of those are already on `impl` and are not actually being changed by this PR.

When reviewing an impl PR, the right mental model is "what does this PR add to `impl`", not "how does this branch differ from `main`". One way to make that obvious in the GitHub UI:

1. On the PR page, open the **Files changed** tab.
2. Confirm the base label in the top-left reads `base: impl <- compare: feature/impl-multi-env-ports`. That confirms you are looking at the diff against `impl`, not against `main`.
3. If anything in that file list does not match what the PR description says it does, that is the signal to dig in. Files that are unchanged from `impl` will not appear in this diff at all.

Locally you can reproduce what the GitHub UI shows with:

```bash
git fetch origin impl feature/impl-multi-env-ports
git diff origin/impl..origin/feature/impl-multi-env-ports -- <path>
```

## Suggestion for the team

To keep this clearer for everyone, I think we should add an `env:impl` (or just `impl`) label and tag every impl-targeted PR with it. Same idea for `env:main` if useful. That way you can filter the PR list by environment and the labels remind reviewers up front which baseline they should be comparing against. Happy to add the label and start applying it if no one objects.

Thanks again for catching the leak, that one would have grown quietly.

PR #323, opened 2026-05-27
Add system_enrichment table and read endpoint (refs #211)

danielbowne:
## Summary

Adds a generic `public.system_enrichment` table that ztmf core owns and a separate enrichment pipeline (private `ztmf-insights` stack) fills with an opaque JSON payload, plus a read-only endpoint that returns it. This is the core-side, **additive** half of decoupling the CMS-specific enrichment pipeline from the open-source data-call app.

The two sides meet at one stable contract: the table. Adding or changing enrichment fields becomes a payload-key change in the pipeline with no ztmf migration. The core app works with or without an enrichment pipeline attached (an unfilled table simply returns 404s).

## Changes

- **Migration 0038** creates `public.system_enrichment (fisma_uuid VARCHAR(255) PK, payload jsonb NOT NULL, synced_at timestamptz)`. `fisma_uuid` has no FK, matching the existing `cfacts_systems` precedent (`fismasystems.fismauid` is `VARCHAR(255)` and not unique); the table is a disposable cache the pipeline rewrites via TRUNCATE+INSERT.
- **`GET /api/v1/systemenrichment/{fisma_uuid}`** returns the payload verbatim. Access control reuses the existing FISMA-system assignment check: admins / read-only admins see any system, ISSOs only assigned ones, missing rows return 404.
- `model.SystemEnrichment.Payload` is `json.RawMessage` so the payload serializes back out as raw JSON, not base64.
- OpenAPI path + `SystemEnrichment` schema, Emberfall E2E coverage, and one empire seed row.

## Scope / deferred

Additive only — **no cfacts files are touched**. Retiring the enrichment-specific (CFACTS) read surface and dropping `cfacts_systems` is deferred until the UI repoints to this endpoint and the old sync is decommissioned, so this is safe to run alongside the current stack. Tracked under the enrichment-extraction epic.

## Deploy note

On merge/deploy to dev, migration 0038 applies on API startup and creates the table. This is intentionally sequenced ahead of the enrichment pipeline's first run against dev.

## Test plan

- [x] `go build` / `go vet` / `staticcheck` clean
- [x] model unit tests pass (incl. payload-serializes-as-JSON guard)
- [x] `make dev-setup` applies 0038; live curl returns 200 (admin), 404 (no row), 200 (assigned ISSO), 403 (unassigned ISSO)
- [x] `make test-full` passes (unit + Emberfall E2E), Failed: 0

Refs #211

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -545,6 +545,16 @@ INSERT INTO public.idm_scoring (idm_name, display_name, score, reasoning) VALUES
     ('Single Factor (Only)', 'Single Factor Only', 1, 'Single factor authentication only - no MFA capability.')
 ON CONFLICT (idm_name) DO NOTHING;
 
+-- Generic system_enrichment row (issue #211), keyed on fismasystems.fismauid for
+-- Shield Gen (1003), which the Emberfall ISSO is assigned to. Gives the
+-- /systemenrichment endpoint E2E coverage via the same access join as CFACTS.
+-- The payload is opaque to ztmf core (owned by the enrichment pipeline).
+INSERT INTO public.system_enrichment (fisma_uuid, payload, synced_at) VALUES (
+    'E1D00198-36D4-4EAB-8C00-501E1D000999',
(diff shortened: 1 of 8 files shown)


MackOverflow (approved):
Looks good to me Danny, no nits to pick!

PR #325, opened 2026-05-28
Fix local dev on Apple Silicon macOS

voidspooks:
## Summary

Three small, independent fixes that together make `make full-stack-up` work on Apple Silicon macOS out of the box. Each one addresses a Mac-specific quirk; none change behavior on Linux.

- **`backend/Dockerfile`** — `FROM scratch` → `FROM gcr.io/distroless/base-debian12`, and drop the hardcoded `x86_64-linux-gnu` libc/ld COPY lines. Distroless ships the right libc/ld/ca-certs for whichever arch the build runs on, so the build is now arch-agnostic and runs native-fast on both arm64 and amd64.
- **`Makefile` (compose-dev.yml + dev.compose.env generators)** — replace `network_mode: host` with explicit `ports:` mappings (`54321:54321` for postgres, `$(API_PORT):$(API_PORT)` for api) and change `DB_ENDPOINT=localhost` → `DB_ENDPOINT=postgre`. Docker Desktop on macOS runs containers inside a Linux VM, where `network_mode: host` binds to the VM, not the Mac, unless the experimental host-networking toggle is enabled. Port mappings behave the same on every platform.
- **`Makefile` (top)** — add `SHELL := /bin/bash`. The default `/bin/sh` on macOS is bash invoked in POSIX mode, where the `echo` builtin prints `-n` literally instead of suppressing the trailing newline. That corrupted every base64-encoded segment in `generate-jwt` and `frontend-env`, producing tokens the API rejected with 401.

## Symptom before this PR (Apple Silicon)

```
=> ERROR [stage-1 4/7] COPY --from=builder /lib/x86_64-linux-gnu/ld-linux-x86-64.so.2 ...
=> ERROR [stage-1 5/7] COPY --from=builder /usr/lib/x86_64-linux-gnu/libc.so.6 ...
failed to compute cache key: ... "/usr/lib/x86_64-linux-gnu/libc.so.6": not found
```

After patching the Dockerfile, the build succeeds but `curl localhost:8080` returns connection-refused (host networking issue), and once port mappings are in place, the JWT from `make frontend-env` decodes to `-n {"alg":"HS256"}\n...` and the API returns 401.

## Test plan

- [x] `make dev-down && rm backend/compose-dev.yml backend/dev.compose.env`
- [x] `make full-stack-up` on Apple Silicon (arm64) — backend container builds natively, both containers come up healthy
- [x] `curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: $VITE_AUTH_TOKEN3" http://localhost:8080/api/v1/users/current` → `200` with the seeded admin user
- [x] Frontend (`vite`) loads in browser and can hit the API
- [x] Verify on Linux that the same flow still works (echo behavior on dash differs but `SHELL := /bin/bash` is widely available; port-mapping change is platform-neutral)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,9 @@
 # ZTMF Development Environment Makefile
 
+# Use bash (not the default /bin/sh, which on macOS is bash in POSIX mode and
+# treats `echo -n` as a literal argument, breaking JWT generation).
+SHELL := /bin/bash
+
 # Single source of truth for the local API port
 API_PORT ?= 8080
 
@@ -79,7 +83,8 @@ backend/compose-dev.yml:
 	@echo "    image: postgres:16.8" >> backend/compose-dev.yml
 	@echo "    env_file:" >> backend/compose-dev.yml
 	@echo "      - dev.compose.env" >> backend/compose-dev.yml
-	@echo "    network_mode: host" >> backend/compose-dev.yml
+	@echo "    ports:" >> backend/compose-dev.yml
+	@echo "      - \"54321:54321\"" >> backend/compose-dev.yml
 	@echo "    volumes:" >> backend/compose-dev.yml
 	@echo "      - postgres_data:/var/lib/postgresql/data" >> backend/compose-dev.yml
 	@echo "    command: -p 54321" >> backend/compose-dev.yml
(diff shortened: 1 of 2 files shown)


danielbowne (approved):
Thanks for the macOS fix, Cameron! Verified the changes on Linux as well:

- `make dev-down`, removed the generated `compose-dev.yml` / `dev.compose.env`, then `make full-stack-up`
- Backend container builds cleanly on the new distroless base
- Both containers come up healthy with the explicit port mappings
- `make generate-jwt` produces a clean token (no `-n` prefix corruption) and `curl /api/v1/users/current` returns HTTP 200 with the seeded admin user

The three Snyk failures are the standard "no `SNYK_TOKEN` on fork PRs" auth issue, not real findings, so we will merge with admin.

PR #326, opened 2026-05-28
Retire CFACTS read surface and drop cfacts_systems table (ztmf-misc#214)

danielbowne:
## Summary

Final step of the ZTMF Insights extraction tracked under ztmf-misc#190. The legacy CFACTS read surface in the ztmf core API is removed and the underlying `cfacts_systems` table is dropped. ztmf-ui#367 has been the sole consumer of this surface and was repointed at the new `system_enrichment` endpoint two weeks ago, with no fallback to the old path. CMS-Enterprise/ztmf#324 removed the write side (lambdas + S3 ingest + Snowflake secrets) last week. This PR closes the loop on the read side.

## Changes

- Add migration `0039cfactsdrop.go` that drops `public.cfacts_systems`. The down migration recreates the table with the cumulative shape from `0015` + `0021` + `0022` (table + two ALTER ADDs) plus the original `COMMENT ON TABLE`, so a rollback restores the table faithfully.
- Remove the `GET /api/v1/cfactssystems` and `GET /api/v1/cfactssystems/{fisma_uuid}` routes from `router.go`, delete `controller/cfacts.go`, delete the `CfactsSystem` model and unit tests.
- Move the per-system access check out of the deleted model file and into `internal/model/fismasystems.go` as `UserCanAccessFismaSystemByUUID`. The query was always against `users_fismasystems INNER JOIN fismasystems` and never touched `cfacts_systems`, so the old name `UserCanAccessCfactsSystem` was misleading; the SQL is byte-identical. The new `systemenrichment` controller is the only caller and is updated to use the renamed function.
- Strip the six `cfacts_systems` INSERTs and stale CFACTS comments from `_test_data_empire.sql`, and delete the unused standalone `_test_data_cfacts_empire.sql` fixture (it was not loaded by any compose file or Makefile target).
- Drop the `cfactsSystemData` anchor and all CFACTS test cases from `emberfall_tests.yml` (eight cases across admin, read-only admin, and ISSO sections). The retained `system_enrichment` cases cover the same access matrix and a new READONLY_ADMIN case was added so coverage of that role's controller branch is preserved.
- Remove the two `/cfactssystems` paths and the `CfactsSystem` schema from `backend/openapi.yaml`.

## Why this is safe to merge

- **Backend not in use.** Both `ztmf-ui` and `ztmf-insights` return zero hits for `cfactssystems` or `cfacts_systems`. The card was repointed at `/api/v1/systemenrichment/{fisma_uuid}` in ztmf-ui#367 and the section is gated on `sdl_sync_enabled`, so systems without insights data hide the card entirely rather than falling back.
- **No referential cascades.** `cfacts_systems` had no foreign keys in or out by design (see the comment in migration `0038systemenrichment.go`), so dropping the table is a clean operation.
- **Real-data migration check.** Verified locally against a postgres volume that holds a real-shape dataset: the drop ran cleanly and `make test-full` passed (91 emberfall cases, 0 failures, 0 skipped) against the resulting schema.

## Test plan

- [x] `make test-unit` clean
- [x] `make test-full` clean against the local postgres volume (real-data shape) — 91/0 emberfall, all integration paths green
- [x] `grep -rn cfacts backend/` shows only the immutable historical migration files (`0015`, `0021`, `0022`, `0028` and `0038` comments referencing the prior pipeline) and the opaque `cfacts` JSON key inside one `system_enrichment` payload (owned by the Snowflake view, not this table)
- [x] OpenAPI spec has no remaining `CfactsSystem` references or orphan `$ref`s
- [x] External repos checked: `gh search code --repo CMS-Enterprise/ztmf-ui cfactssystems` and `--repo CMS-Enterprise/ztmf-insights cfactssystems` both return empty

## Follow-ups

Once this lands, ztmf-misc#220 (rename `CfactsRecordCard` to `SystemEnrichmentCard` in ztmf-ui) is unblocked. After that, the only remaining open sub-issue of the extraction epic is ztmf-misc#224 (migrate the Kion plumbing into the insights stack), which is explicitly out of scope for this PR.

Refs ztmf-misc#214.


MackOverflow (approved):
Clean PR Danny! No nits!

PR #327, opened 2026-05-28
fix(#309): accept non-UUID system IDs and case-insensitive CFACTS matching

### LINKED ISSUE #309: Some HHS SystemIDs are not GUIDs
### Description

Based on a list I got from Evan, not all system IDs for HHS systems are in the GUID format.  I vaguely remember Richard doing something with GUID validation, but perhaps it wasn't implemented because CMS is inconsistent on whether the GUIDs contain dashes or not.

### Steps to Reproduce

CDC has the system IDs:
CDC8767221
CDC356346
CDC6700474
CDC1723552
CDC1723554

### Expected Behavior

We shouldn't judge the systemIDs too hard.  And do case insensitive matching where matching is needed.

### Actual Behavior

Some SystemID may be too short.

### Environment

all

### Acceptance Criteria

- [ ] We've either confirmed that the System IDs are not checked for being GUIDs, or we remove that task.

### Severity

Low - Minor inconvenience

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

voidspooks:
Closes #309

> **Note:** Rebased onto latest `main` after #326 landed (which retired the legacy CFACTS read surface and deleted `cfactssystems.go`). The original `cfactssystems.go` hunks were dropped as part of that rebase — the case-insensitive matching now lives in `UserCanAccessFismaSystemByUUID`, where the old `UserCanAccessCfactsSystem` access check was relocated.

## Summary

- Relaxed `fismauid` validation in `FismaSystem.validate()` to require only a non-empty string instead of a valid UUID. Some HHS OpDiv system IDs (e.g. `CDC8767221`) are not GUIDs and were being rejected with a 400. This is the core #309 fix — the gate that blocks creating a CDC-style system.
- Made the access-check `WHERE` clause in `UserCanAccessFismaSystemByUUID` (`fismasystems.go`) case-insensitive using `LOWER()`. Defensive future-proofing so non-GUID OpDiv IDs match regardless of case once system enrichment opens beyond CMS.
- Updated `fismasystems_test.go`: replaced the `InvalidUUID` test case with `ValidSystemNonUUID` (CDC-style ID should pass) and `EmptyFismaUID` (empty string should still fail).

## Testing

Go unit tests pass (run in a `golang:1.25` container — `go test -short ./...`):
```
--- PASS: TestFismaSystem_Validate/ValidSystemUUID
--- PASS: TestFismaSystem_Validate/ValidSystemNonUUID
--- PASS: TestFismaSystem_Validate/InvalidEmail
--- PASS: TestFismaSystem_Validate/EmptyFismaUID
ok  github.com/CMS-Enterprise/ztmf/backend/internal/model  0.003s
```
Full `go test -short ./...` suite is green; `go build ./...` and `go vet ./internal/model/` clean.

> Snyk / secret-dependent CI checks show red because this branch is from a fork (no repo secrets) — expected, maintainers re-run on their side.

### DIFF
--- a/backend/internal/model/fismasystems.go
+++ b/backend/internal/model/fismasystems.go
@@ -121,7 +121,7 @@ func UserCanAccessFismaSystemByUUID(ctx context.Context, userID string, fismaUUI
 		Select("1").
 		From("users_fismasystems").
 		InnerJoin("fismasystems ON fismasystems.fismasystemid = users_fismasystems.fismasystemid").
-		Where("fismasystems.fismauid = ? AND users_fismasystems.userid = ?", fismaUUID, userID).
+		Where("LOWER(fismasystems.fismauid) = LOWER(?) AND users_fismasystems.userid = ?", fismaUUID, userID).
 		Limit(1)
 
 	_, err := queryRow(ctx, sqlb, pgx.RowTo[int])
@@ -370,8 +370,8 @@ func ReactivateFismaSystem(ctx context.Context, input ReactivateInput) (*FismaSy
 func (f *FismaSystem) validate() error {
 	err := InvalidInputError{data: map[string]any{}}
 
-	if !isValidUUID(f.FismaUID) {
-		err.data["fismauid"] = f.FismaUID
+	if f.FismaUID == "" {
+		err.data["fismauid"] = "required"
 	}
 
 	if f.DataCallContact != nil && !isValidEmail(*f.DataCallContact) {
--- a/backend/internal/model/fismasystems_test.go
+++ b/backend/internal/model/fismasystems_test.go
@@ -162,9 +162,9 @@ func TestFismaSystem_Validate(t *testing.T) {
 		description string
 	}{
 		{
-			name: "ValidSystem",
+			name: "ValidSystemUUID",
 			system: FismaSystem{
-				FismaUID:              "12345678-1234-4abc-8def-123456789abc", // Valid UUID v4
+				FismaUID:              "12345678-1234-4abc-8def-123456789abc",
 				FismaAcronym:          "TEST",
 				FismaName:             "Test System",
 				DataCenterEnvironment: stringPtr("AWS"),
@@ -172,12 +172,25 @@ func TestFismaSystem_Validate(t *testing.T) {
 				ISSOEmail:             stringPtr("isso@example.com"),
 			},
 			wantErr:     false,
-			description: "A valid FISMA system should not return an error",
+			description: "A FISMA system with a UUID fismauid should not return an error",
+		},
+		{
+			name: "ValidSystemNonUUID",
+			system: FismaSystem{
+				FismaUID:              "CDC8767221",
+				FismaAcronym:          "CDC",
+				FismaName:             "CDC Test System",
+				DataCenterEnvironment: stringPtr("AWS"),
+				DataCallContact:       stringPtr("test@example.com"),
+				ISSOEmail:             stringPtr("isso@example.com"),
+			},
+			wantErr:     false,
+			description: "A FISMA system with a non-UUID fismauid (e.g. CDC-style) should not return an error",
 		},
 		{
 			name: "InvalidEmail",
 			system: FismaSystem{
-				FismaUID:              "12345678-1234-1234-1234-123456789abc",
+				FismaUID:              "CDC8767221",
 				FismaAcronym:          "TEST",
 				FismaName:             "Test System",
 				DataCenterEnvironment: stringPtr("AWS"),
@@ -188,9 +201,9 @@ func TestFismaSystem_Validate(t *testing.T) {
 			description: "Invalid email should return validation error",
 		},
 		{
-			name: "InvalidUUID",
+			name: "EmptyFismaUID",
 			system: FismaSystem{
-				FismaUID:              "not-a-uuid",
+				FismaUID:              "",
 				FismaAcronym:          "TEST",
 				FismaName:             "Test System",
 				DataCenterEnvironment: stringPtr("AWS"),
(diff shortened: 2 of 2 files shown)


backend/internal/model/cfactssystems.go
```diff
@@ -75,7 +75,7 @@ func FindCfactsSystems(ctx context.Context, input FindCfactsSystemsInput) ([]*Cf
 
 	if input.UserID != nil {
 		sqlb = sqlb.
-			InnerJoin("fismasystems ON fismasystems.fismauid = cfacts_systems.fisma_uuid").
+			InnerJoin("fismasystems ON LOWER(fismasystems.fismauid) = LOWER(cfacts_systems.fisma_uuid)").
```
voidspooks: Don't think this will cause any issues, but worth double checking me on this assumption.

backend/internal/model/fismasystems.go
```diff
@@ -348,8 +348,8 @@ func ReactivateFismaSystem(ctx context.Context, input ReactivateInput) (*FismaSy
 func (f *FismaSystem) validate() error {
 	err := InvalidInputError{data: map[string]any{}}
 
-	if !isValidUUID(f.FismaUID) {
-		err.data["fismauid"] = f.FismaUID
+	if f.FismaUID == "" {
```
voidspooks, line 373: coming from Ruby land I didn't think empty string check would be sufficient, but AFAICT this _should_ suffice. again - double check me here.

danielbowne (changes requested):
Hey Cameron - thanks for digging into this one, and sorry in advance: some of this collided with a big change that landed on main a few minutes ago, so this PR needs a rebase and a bit of rework. Let me walk through it so it's an easy fix.

## What changed upstream

We've been decoupling the ZTMF Insights / enrichment pipeline out of ztmf core into its own repo. The last step of that (#326, just merged) retired the whole legacy CFACTS read surface: it deleted `backend/internal/model/cfactssystems.go`, dropped the `cfacts_systems` table, and removed the `/api/v1/cfactssystems` endpoints. The UI already moved off that surface onto the new `/api/v1/systemenrichment/{fisma_uuid}` endpoint, so the old CFACTS read path is gone for good.

That means the two hunks in this PR that touch `cfactssystems.go` will hit a modify/delete conflict, since the file no longer exists.

## The good news

Your actual #309 fix is exactly right and we still want it. The validation relaxation in `fismasystems.go` (`validate()` accepting a non-empty `fismauid` instead of requiring a UUID) and the matching test cases (`ValidSystemNonUUID` / `EmptyFismaUID`) are clean against the new main and don't conflict. That's the real fix here: it's the gate that blocks creating a CDC-style system at all, and it bites right now regardless of enrichment. Please keep it as-is.

## What to do

1. Rebase onto the latest `main` (it'll flag the `cfactssystems.go` conflict).
2. Drop both `cfactssystems.go` hunks entirely. That file is gone, so there's nothing to keep there.
3. On the case-insensitive matching: the access check that used to live in `UserCanAccessCfactsSystem` got relocated into `fismasystems.go` as `UserCanAccessFismaSystemByUUID`, which now gates access to the `/systemenrichment` endpoint and still does an exact match on `fismauid`. The honest read is that case-insensitivity isn't load-bearing there today: system enrichment is CMS-only for now and matches on real UUIDs, so the non-GUID OpDiv systems like CDC aren't flowing through that path yet. So porting your `LOWER()` over is optional future-proofing for whenever enrichment opens up beyond CMS, not something I'd block on. If you want to fold it into `UserCanAccessFismaSystemByUUID` while you're in here it's a reasonable defensive add, otherwise drop it and we revisit when the pipeline goes multi-OpDiv. Your call.

So at minimum this PR is just the validation fix, which is a nice tight change.

## On whether the check is too loose

Answering your question directly: non-empty is the right floor, don't tighten it. #309 says not to judge the IDs too hard, and the CDC examples (CDC8767221, CDC356346, ...) are all different lengths with no shared format, so anything stricter risks bouncing a valid OpDiv ID we haven't seen yet. For what it's worth I traced the rest of the path so you don't have to: `fismauid` is `varchar(255)` in the DB, the fismasystems routes key off the integer `fismasystemid` (not the string), and `/systemenrichment/{fisma_uuid}` already accepts letters/digits/hyphens. So your `validate()` change is the only gate actually standing in CDC's way. Nice and contained.

One edge to keep in the back of your mind (not for this PR): that `/systemenrichment` route pattern matches letters, digits, and hyphens but not underscores or dots. CDC's IDs are clean alphanumeric so they're fine, but if some future OpDiv ever hands us a system ID with an underscore or dot in it, that route would 404 before it reaches the handler. Just worth remembering as more OpDivs onboard.

## Don't worry about the Snyk / CI checks

Since you're working from a fork on standard SSO access, the Snyk jobs and anything needing repo secrets won't run for you (forks don't get credentials), so you'll see those show up red. That's expected and not on you - we run them on our side after pulling the branch, same as we did on your macOS PR. Just get the Go unit tests green locally and we'll handle the rest.

Happy to push the rebase for you if you'd rather not wrestle the conflict, just say the word and I'll take care of it. Thanks again for chasing the non-GUID issue, it's a real one for the HHS systems.

danielbowne (approved):
This looks great, thanks for the quick turnaround. The rebase is clean, the CFACTS hunks are gone, and the validation relax is exactly the fix #309 needs. Nice call folding the `LOWER()` into `UserCanAccessFismaSystemByUUID` too - it's not load-bearing today (enrichment is still CMS-only on real UUIDs) but it's good defensive future-proofing for when the OpDivs come online, and it costs nothing on a table this size.

Verified locally: build and `go vet` clean, and the validate tests (`ValidSystemNonUUID`, `EmptyFismaUID`) plus `TestIsValidUUID` all pass. Approving. As before, ignore the Snyk reds - those are the fork-credential thing, we'll confirm CI on our side after merge.

PR #329, opened 2026-06-02
Fix local/test database seeding on a fresh Docker volume

danielbowne:
## Summary

`make dev-setup` was not loading the Empire test data (`_test_data_empire.sql`) into a fresh Postgres volume. A new clone, a new machine, or any `docker compose down -v` came up with an empty database (0 FISMA systems, 0 questions), which is what surfaced when setting up the backend on a freshly migrated OS.

## Root cause

The seed step was gated on the database host string:

```go
if cfg.Db.PopulateSql != nil && cfg.Db.Host == "localhost" {
```

The dev API container reaches Postgres over the `postgre` Docker Compose service name (`compose-dev.yml` overrides `DB_ENDPOINT=postgre` so the container can talk to Postgres over the internal Docker network). The `localhost` value in `dev.compose.env` only applies to host-side tooling. So inside the container the check evaluated `postgre == "localhost"`, was false, and seeding was silently skipped.

This was masked on existing machines two ways, which is why it did not show up for other developers:
- `backend/compose-dev.yml` was only regenerated when missing, so existing checkouts kept an older on-disk compose file and never picked up the `postgre` override.
- The `backend_postgres_data` volume persists across restarts, so databases seeded under the older configuration stayed seeded.

A fresh volume on current code has neither of those, so the seed never ran.

## Fix

Two changes, addressing both halves of the root cause:

1. **Gate seeding on `ENVIRONMENT`, not the database host.** `ENVIRONMENT` is host- and platform-agnostic: `local` for dev (`dev.compose.env`), `test` for the Emberfall E2E stack (`compose-test.yml`), and it defaults to `production` everywhere else.

   ```go
   if cfg.Db.PopulateSql != nil && cfg.IsLocalOrTest() {
   ```

2. **Make the `backend/compose-dev.yml` generation target PHONY** so it regenerates on every `make dev-setup` / `dev-up`. This stops developers from running a stale on-disk compose file that predates the `DB_ENDPOINT=postgre` override, which was the second reason fresh-volume seeding silently broke.

### Environment-gate helpers

The environment checks are now centralized as two `config` helpers, replacing scattered string comparisons:

- `config.IsLocal()` -> `ENVIRONMENT == "local"`
- `config.IsLocalOrTest()` -> `ENVIRONMENT == "local"` or `"test"`

These are deliberately distinct. Seeding uses `IsLocalOrTest()` (safe in local dev and the E2E test stack). Just-in-time ADMIN user creation in the auth middleware uses `IsLocal()` (local dev only, never the E2E test stack). Keeping them separate preserves that asymmetry, and a new unit test (`TestEnvironmentGates`) locks the contract across local/test/production/dev/impl/prod/empty.

## Why this is safe for deployed environments

Seeding requires two independent conditions, and deployed environments fail both:
- Deployed dev, impl, and prod resolve `ENVIRONMENT` to `dev`, `impl`, and `prod` respectively (see `infrastructure/ecs.tf` and the tfvars), none of which match `local` or `test`.
- Deployed task definitions never set `DB_POPULATE`, so `cfg.Db.PopulateSql` is nil and the populate call cannot run regardless.

The allowlist of environment names is deliberate: an unrecognized environment fails closed (no seeding) rather than open.

## Why this does not break macOS developers

The fix does not touch the compose networking, port mappings, or the `DB_ENDPOINT=postgre` override introduced for Apple Silicon. macOS dev uses the same `ENVIRONMENT=local`, so it seeds a fresh volume the same way Linux does. The PHONY change only ever regenerates `compose-dev.yml` to its current correct form (the file is gitignored and marked "DO NOT EDIT - Managed by Makefile"), and existing seeded volumes are unaffected. The auth-middleware refactor is behavior-preserving: JIT ADMIN creation still fires in local dev only.

## Dependency security fix

The PR build's Snyk container scan failed the pipeline on a critical vulnerability in a transitive dependency, fixed here to keep the branch green:

- **golang.org/x/net** bumped from v0.47.0 to v0.55.0. v0.47.0 carries a critical Improper Authentication vulnerability in `golang.org/x/net/idna` (SNYK-GOLANG-GOLANGORGXNETIDNA-17116876), first fixed in 0.54.0.
- `go mod tidy` also advanced the related modules it pulls: `golang.org/x/crypto` to v0.51.0 and `golang.org/x/text` to v0.37.0.

`x/net` is an indirect dependency, so this is a `go.mod` / `go.sum` change only with no source edits.

## Testing

- Fresh volume on Linux: `docker compose down -v` then `make dev-setup` now logs `Populating database with /app/_test_data_empire.sql` and the database comes up with 9 users, 4 FISMA systems, 18 questions, 30 scores. No manual psql load needed.
- Confirmed `compose-dev.yml` now regenerates on every run (PHONY), overwriting a deliberately corrupted copy.
- New `TestEnvironmentGates` unit test passes.
- `make test-full` (unit + Emberfall E2E on the isolated test stack with `ENVIRONMENT=test`) passes, confirming the E2E seed path is preserved and the middleware refactor is behavior-preserving.


MackOverflow (dismissed):
Looks great Danny, I see a couple to-dos that I think would be smart, but otherwise, very clean!

jsos3-cms (approved):
LGTM!

PR #330, opened 2026-06-03
feat(#314): remove legacy ADMIN / READONLY_ADMIN role values

### LINKED ISSUE #314: Database: Remove legacy ADMIN and READONLY_ADMIN role values (follow-up cleanup)

voidspooks:
## Summary

Stage D follow-up cleanup for the multi-OpDiv role swap. Closes #314.

The Stage B swap (`0036usersroleswap.go`) already migrated prod rows `ADMIN -> OWNER` and `READONLY_ADMIN -> HHS_READONLY_ADMIN`, so the legacy values are dead. This PR removes them everywhere they were still recognized and adds a DB-level guard rejecting any new write that carries one.

**Canonical mapping** (matches the Stage B migration): `ADMIN → OWNER`, `READONLY_ADMIN → HHS_READONLY_ADMIN`.

## Changes

**Production code**
- `internal/model/validations.go` — drop both legacy keys from the `roles` map so `isValidRole` rejects them.
- `internal/model/users.go` — drop legacy values from `HasUnscopedRead`, `IsAdmin`, `IsReadOnlyAdmin`.
- `cmd/api/internal/auth/middleware.go` — local-dev auto-create now stamps `OWNER`.
- `internal/model/massemails.go` — drop legacy `ADMIN` from the admin-recipient role list. The `"ADMIN"` mass-email **audience-group** key is an API contract and is intentionally kept.
- `cmd/api/internal/controller/fismasystems.go` — refresh a now-stale scope comment.

**Migration**
- `migrations/0040rolecleanup.go` — adds `CHECK (role NOT IN ('ADMIN','READONLY_ADMIN'))` as defense in depth. Down migration drops the constraint so the prior release can roll back during the soak window (the validation map / helper values are restored by redeploying the prior binary).

**Fixtures / spec / tests**
- `_test_data.sql`, `_test_data_empire.sql`, `emberfall_tests.yml`, `openapi.yaml` updated to the new tier names (legacy values removed from both role enums).
- `validations_test.go` flips the legacy-role assertions from valid → invalid.
- `users_test.go`, `authorization_test.go`, `events_test.go`, `scores_test.go`, `scores_integration_test.go` updated to the new tier names.

## Acceptance criteria

- [x] `isValidRole("ADMIN")` / `isValidRole("READONLY_ADMIN")` return false
- [x] All test fixtures use new-tier role names
- [x] OpenAPI role enum no longer lists `ADMIN` / `READONLY_ADMIN`
- [x] Local-dev auto-create middleware emits a new-tier role (`OWNER`)
- [x] Down migration drops the guard so a rollback to the prior release works during the soak window
- [x] `CHECK` constraint rejects legacy role values (defense in depth)
- [ ] Pre-flight (operational, not in this diff): `SELECT count(*) FROM users WHERE role IN ('ADMIN','READONLY_ADMIN')` returns 0 at the moment of merge

## Testing

`go build ./...` and `go test -short ./...` green in `golang:1.25`. Both YAML files re-parse cleanly.

> Note: this branch is from a fork without repo secrets, so Snyk / secret-dependent CI checks will show red — expected; re-run after pulling.

### DIFF
--- a/backend/_test_data.sql
+++ b/backend/_test_data.sql
@@ -4,10 +4,10 @@
 -- new columns in the multi-OpDiv migration and the positional form would
 -- silently shift values across columns on future schema bumps.
 INSERT INTO public.users (email, fullname, role, identity_provider)
-    VALUES ('Test.User@nowhere.xyz', 'Admin User', 'ADMIN', 'okta')
+    VALUES ('Test.User@nowhere.xyz', 'Admin User', 'OWNER', 'okta')
     ON CONFLICT DO NOTHING;
 INSERT INTO public.users (email, fullname, role, identity_provider)
-    VALUES ('Readonly.Admin@nowhere.xyz', 'Readonly Admin User', 'READONLY_ADMIN', 'okta')
+    VALUES ('Readonly.Admin@nowhere.xyz', 'Readonly Admin User', 'HHS_READONLY_ADMIN', 'okta')
     ON CONFLICT DO NOTHING;
 INSERT INTO public.users (email, fullname, role, identity_provider)
     VALUES ('Isso.User@nowhere.xyz', 'ISSO Test User', 'ISSO', 'okta')
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -13,12 +13,12 @@ INSERT INTO public.opdivs (code, name, is_parent, active)
 
 -- Test user for Emberfall E2E tests (matches _test_data.sql for CI/CD compatibility)
 INSERT INTO public.users (email, fullname, role, identity_provider)
-    VALUES ('Test.User@nowhere.xyz', 'Admin User', 'ADMIN', 'okta')
+    VALUES ('Test.User@nowhere.xyz', 'Admin User', 'OWNER', 'okta')
     ON CONFLICT DO NOTHING;
 
--- Test ADMIN User (Death Star Commander - full administrative access)
+-- Test OWNER User (Death Star Commander - full administrative access)
 INSERT INTO public.users (userid, email, fullname, role, identity_provider)
-    VALUES ('11111111-1111-1111-1111-111111111111', 'Grand.Moff@DeathStar.Empire', 'Grand Moff Tarkin', 'ADMIN', 'okta')
+    VALUES ('11111111-1111-1111-1111-111111111111', 'Grand.Moff@DeathStar.Empire', 'Grand Moff Tarkin', 'OWNER', 'okta')
     ON CONFLICT DO NOTHING;
 
 -- Test ISSO Users (Imperial Officers)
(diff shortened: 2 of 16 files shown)


danielbowne (approved):
Approving. Pulled the branch and ran the migration against a copy of prod data locally (restored the real DB, not seed data) so the new constraint got tested against actual role distributions.

The migration itself is fine. Since the Stage B swap in 0036 already moved everyone off ADMIN/READONLY_ADMIN, there are no legacy rows left, so the plain ADD CONSTRAINT validates instantly and the API comes up clean. I tried forcing a legacy value in (UPDATE users SET role='ADMIN') and it gets rejected by the check constraint as expected. The only thing that would make this fail a deploy is an environment that still has legacy rows when it runs, and neither prod nor dev does, so we're good there.

On the app side, I went looking for anything that still references the old strings. Nothing in live code except the two migrations and the "ADMIN" audience key in massemails.go, which is an API group name rather than a role (and sqlForADMIN now selects OWNER/HHS_ADMIN/OPDIV_ADMIN, so that's correct). Authorization all goes through the role helpers instead of raw string comparisons, and each old tier maps to its replacement, so nobody gets locked out and there's no gap. The allowed set lines up across validations.go, the DB constraint, and both OpenAPI enums. Both seed files use only the new values. The one role literal written at runtime is OWNER on the local-only JIT path, and events just persist whatever role the user already has, so nothing writes a legacy value back in.

Tests and emberfall expectations were updated to match, and the full E2E suite passes locally (91/91). Looks good to merge.

One thing you could leave or tidy, your call: the "ADMIN" key in massemails.go shares a name with the role we just removed, which might trip up someone later. There's already a comment explaining it, so not a blocker.

danielbowne:
Re-homed onto an org branch as #332 so the full CI pipeline (Snyk, build, deploy) can run, which fork PRs cannot. Same change plus a merge of current main. Continuing review and merge there. Thanks @voidspooks.

PR #331, opened 2026-06-08
Bump Go to 1.25.11 for standard-library CVEs

danielbowne:
## Summary

Snyk started failing the prod orchestration on two high-severity vulnerabilities in the Go standard library at 1.25.10:

- `std/mime`
- `std/crypto/x509`

Both are fixed in Go 1.25.11. The advisories surfaced after #329 had already passed its PR checks, so the same code that was green on the PR now blocks the prod pipeline at the analysis stage. Until this lands, every merge to main fails Snyk and nothing deploys.

## Change

- `backend/go.mod`: `go 1.25.10` -> `1.25.11`
- `backend/Dockerfile`: builder image `golang:1.25.10-bookworm` -> `golang:1.25.11-bookworm`

No code changes. Builds clean locally and the host unit tests pass.

## Note

The prod run failed at the analysis stage before the deploy step, so prod was not affected and is still running the prior image. This unblocks the pipeline so #329 (and subsequent work) can deploy.

PR #332, opened 2026-06-08
feat(#314): remove legacy ADMIN / READONLY_ADMIN role values

danielbowne:
Re-homes #330 (originally authored by @voidspooks) onto an org branch so the full CI pipeline runs. Fork PRs do not get the org secrets Snyk and the build/deploy steps need, so the change could not be validated or shipped from the fork.

Same content as #330, plus a merge of current main, which brings in the seeding fix (#329) and the Go 1.25.11 bump (#331). The one overlap with main was in `auth/middleware.go` (the JIT user-creation gate); resolved to keep main's `cfg.IsLocal()` helper and this branch's `OWNER` role.

## What this does

Stage D of the role cleanup: removes the legacy `ADMIN` and `READONLY_ADMIN` role values. A prior migration (`0036`) already converted existing rows (`ADMIN -> OWNER`, `READONLY_ADMIN -> HHS_READONLY_ADMIN`); this stops the app from accepting or emitting the legacy values and adds a DB guard (`0040`, a CHECK constraint rejecting the legacy strings).

## Validation

- Migration tested against a copy of prod data: zero legacy rows, constraint applies clean, API starts, and a legacy write is rejected as expected.
- Authorization routes through the role helpers; every legacy tier maps to its replacement, so no lockout or access gap.
- Allowed role set is consistent across `validations.go`, the DB constraint, and both OpenAPI enums; both seed files use only the new values.
- Emberfall E2E passes locally (91/91).

Closes #330.


jsos3-cms (approved):
Importing approval from #330 for this.

tarratsco (approved):

PR #333, opened 2026-06-09
Backend: enforce OpDiv-scoped RBAC and add OpDiv management endpoints

### LINKED ISSUE #313: Backend: Surface OpDiv and identity_provider fields and add scope predicates

danielbowne:
## Summary

Adds authorization enforcement for the multi-OpDiv role model on top of the
schema that already shipped (opdivs reference table, fismasystems.opdiv_id,
users_opdivs junction, the new role enum). Before this branch, reads were only
partially scoped and write paths were not scoped at all: an OPDIV_ADMIN could
edit or delete systems in any OpDiv, score any system, and manage any user, and
there was no endpoint to create an OpDiv or grant a user an OpDiv. This branch
closes those gaps so a scoped admin manages only their own OpDiv, while
OWNER/HHS tiers retain full cross-OpDiv access.

## What changed

**OpDiv administration (OWNER-only)**
- `POST /api/v1/opdivs`, `PUT /api/v1/opdivs/{opdiv_id}` to create / update / deactivate an OpDiv.

**OpDiv grants (user membership)**
- `GET/POST/DELETE /api/v1/users/{userid}/assignedopdivs[/{opdiv_id}]` to list, grant, and revoke OpDiv membership. An OPDIV_ADMIN may only grant/revoke an OpDiv they themselves hold, and only onto a user within their assignable tier.
- A user's `identity_provider` is derived from their OpDiv on grant/revoke (CMS to okta, otherwise entra) and is only overridable by an OWNER.

**Write-path scope**
- Editing, decommissioning, reactivating, and assigning a user to a system, scoring a system, and marking a data call complete are all gated so an OPDIV_ADMIN is limited to systems in an OpDiv they hold. OWNER/HHS_ADMIN are unrestricted; ISSO/ISSM keep their per-system assignment path.
- Role-escalation guard on user create/update: an OPDIV_ADMIN cannot mint or modify a user above the OpDiv tier, and cannot manage a user outside their OpDiv.

**Read-path scope (fail-closed)**
- User, score, and data-call-completion lists are scoped to the caller's OpDivs for OPDIV tiers. A scoped admin with zero grants matches nothing rather than falling through to an unscoped read.
- The audit trail (`GET /events`) and mass email (`POST /massemails`) are restricted to the unscoped admin tiers, since neither has a per-OpDiv scope to filter on.

**Migration**
- `0041` adds an idempotent index on `fismasystems.opdiv_id` to back the scope predicates. No data change.

**Docs and tests**
- OpenAPI updated for the new endpoints and the scoped / restricted behavior.
- Unit tests for the role helpers, the fail-closed SQL predicate, and the controller authorization gates; Emberfall E2E negative matrix asserting cross-OpDiv writes return 403 and scoped reads exclude out-of-OpDiv rows.

## Related issues

- Closes CMS-Enterprise/ztmf#313 (Backend: surface OpDiv / identity_provider fields and add scope predicates), and additionally delivers three items that issue had deferred (OpDiv scope on scores/events/datacalls, grant-management endpoints, and write-path scope beyond create) because the authorization layer must be complete before the dual-IdP work admits OpDiv-scoped users.
- Part of epic CMS-Enterprise/ztmf-misc#198 (HHS multi-OpDiv expansion).
- Builds on the database work in #311 / #312 and the role-value cleanup in #314, all already on `main`.

## Safety and rollout

A no-op for current production: no OPDIV_ADMIN users exist yet, and the
OWNER / HHS tiers short-circuit every new check. The only migration is an index
(idempotent up / reversible down). Rollback is a redeploy of the prior image.
Per team policy this branch is not deployed to a shared environment before merge.

## Testing

`make test-full` green: unit + isolated integration + 111 Emberfall E2E.

### How to verify

- The OpDiv negative matrix lives at the bottom of `backend/emberfall_tests.yml`: cross-OpDiv writes return 403, scoped reads exclude out-of-OpDiv rows, and `/events` + mass email are gated to unscoped admins.
- To poke it by hand, the seed adds an `OPDIV_ADMIN` (EMPIRE grant only) plus REBELLION systems 1005/1006 (token anchor `opDivAdminHeaders` in the test file). `PUT /fismasystems/1005` or `POST /scores` for 1006 returns 403; EMPIRE systems 1001-1004 succeed.
- The scope logic is centralized in `backend/internal/model/users.go` (`CanManageFismaSystem`, `CanManageUser`, `EffectiveOpDivScope`) and exercised in `backend/cmd/api/internal/controller/rbac_enforcement_test.go`, so the role/scope matrix is reviewable in one place.

## Sequencing

Lands before the dual-IdP branch, which rebases on top once this merges. A
paired ztmf-ui PR adds the matching UI gates (hide actions and views a scoped
admin would be 403'd on); the two land together so no admin loses UI in between.

### DIFF
--- a/backend/internal/model/opdivs.go
+++ b/backend/internal/model/opdivs.go
@@ -43,13 +43,17 @@ func FindOpDivs(ctx context.Context, input FindOpDivsInput) ([]*OpDiv, error) {
 }
 
 func (o *OpDiv) validate() error {
-	inputErr := InvalidInputError{data: map[string]any{}}
+	// Canonicalize before checking AND storing so the length bound can't be
+	// bypassed with padding and the active-code uniqueness index (LOWER(code),
+	// not trimmed) cannot be defeated by leading/trailing whitespace.
+	o.Code = strings.TrimSpace(o.Code)
+	o.Name = strings.TrimSpace(o.Name)
 
-	code := strings.TrimSpace(o.Code)
-	if code == "" || len(code) > 16 {
+	inputErr := InvalidInputError{data: map[string]any{}}
+	if o.Code == "" || len(o.Code) > 16 {
 		inputErr.data["code"] = "1-16 characters required"
 	}
-	if name := strings.TrimSpace(o.Name); name == "" || len(name) > 128 {
+	if o.Name == "" || len(o.Name) > 128 {
 		inputErr.data["name"] = "1-128 characters required"
 	}
 
--- a/backend/internal/model/opdivs.go
+++ b/backend/internal/model/opdivs.go
@@ -2,6 +2,8 @@ package model
 
 import (
 	"context"
+	"errors"
+	"strings"
 
 	"github.com/jackc/pgx/v5"
 )
@@ -38,4 +40,61 @@ func FindOpDivs(ctx context.Context, input FindOpDivsInput) ([]*OpDiv, error) {
 	}
 
 	return query(ctx, sqlb, pgx.RowToAddrOfStructByName[OpDiv])
+}
+
+func (o *OpDiv) validate() error {
+	inputErr := InvalidInputError{data: map[string]any{}}
+
+	code := strings.TrimSpace(o.Code)
+	if code == "" || len(code) > 16 {
+		inputErr.data["code"] = "1-16 characters required"
+	}
+	if name := strings.TrimSpace(o.Name); name == "" || len(name) > 128 {
+		inputErr.data["name"] = "1-128 characters required"
+	}
+
+	if len(inputErr.data) > 0 {
+		return &inputErr
+	}
+	return nil
+}
+
+// Save inserts a new OpDiv or updates an existing one (OpDivID > 0). A new
+// OpDiv is always created active; deactivation is done by updating an existing
+// row with active=false. The partial unique index on LOWER(code) WHERE active
+// means inserting a second active row with an existing code returns ErrNotUnique.
+func (o *OpDiv) Save(ctx context.Context) (*OpDiv, error) {
+	if err := o.validate(); err != nil {
+		return nil, err
+	}
+
+	var sqlb SqlBuilder
+	if o.OpDivID == 0 {
+		sqlb = stmntBuilder.
+			Insert("public.opdivs").
+			Columns("code", "name", "is_parent", "active").
+			Values(o.Code, o.Name, o.IsParent, true).
+			Suffix("RETURNING opdiv_id, code, name, is_parent, active")
+	} else {
+		sqlb = stmntBuilder.
+			Update("public.opdivs").
+			Set("code", o.Code).
+			Set("name", o.Name).
+			Set("is_parent", o.IsParent).
+			Set("active", o.Active).
+			Where("opdiv_id=?", o.OpDivID).
+			Suffix("RETURNING opdiv_id, code, name, is_parent, active")
+	}
+
+	saved, err := queryRow(ctx, sqlb, pgx.RowToStructByName[OpDiv])
+	if err != nil {
+		// The only unique constraint on opdivs is the active-code index, so map
+		// a uniqueness violation to a clear field-level message the frontend can
+		// show inline rather than leaking the raw constraint detail.
+		if errors.Is(err, ErrNotUnique) {
+			return nil, &InvalidInputError{data: map[string]any{"code": "an OpDiv with this code already exists"}}
+		}
+		return nil, err
+	}
+	return saved, nil
 }
--- a/backend/internal/model/opdivs.go
+++ b/backend/internal/model/opdivs.go
@@ -13,8 +13,13 @@ type OpDiv struct {
 	OpDivID  int32  `json:"opdiv_id" db:"opdiv_id"`
 	Code     string `json:"code"`
 	Name     string `json:"name"`
-	IsParent bool   `json:"is_parent" db:"is_parent"`
-	Active   bool   `json:"active"`
+	// IsParent and Active are pointers so an update can distinguish "not supplied"
+	// (nil, leave the column untouched) from an explicit false. With a non-pointer
+	// bool, any update that omits the field defaults to false: omitting is_parent
+	// would demote the HHS parent OpDiv (which drives HHS-wide authorization), and
+	// omitting active would silently deactivate an active OpDiv.
+	IsParent *bool `json:"is_parent" db:"is_parent"`
+	Active   *bool `json:"active"`
 }
 
 // FindOpDivsInput holds optional filters for listing OpDivs.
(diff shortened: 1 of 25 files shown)


backend/internal/model/opdivs.go
```diff
@@ -39,3 +41,64 @@ func FindOpDivs(ctx context.Context, input FindOpDivsInput) ([]*OpDiv, error) {
...
+			Columns("code", "name", "is_parent", "active").
+			Values(o.Code, o.Name, o.IsParent, true).
+			Suffix("RETURNING opdiv_id, code, name, is_parent, active")
+	} else {
+		sqlb = stmntBuilder.
+			Update("public.opdivs").
+			Set("code", o.Code).
+			Set("name", o.Name).
+			Set("is_parent", o.IsParent).
+			Set("active", o.Active).
```
voidspooks: This is very solid work! I think the only thing I saw that gives me some pause is that for active is a bool instead of a *bool. This is an optional field defined in `openapi.yaml`, so an update that doesn't pass in an active bool would be interpreted as `false` - so an opdiv that actually is active could inadvertently be switched to false.

voidspooks (approved):
Just see one thing to fix potentially - otherwise this looks very good

danielbowne:
Good catch, thank you - resolved. Made `active` a `*bool` so the update only writes it when the field is actually supplied; an omitted field now leaves the column alone, and an explicit false still deactivates.

While I was in there I realized `is_parent` had the exact same problem, and it is the more dangerous of the two: a PUT that only changes the name would have dropped `is_parent` to false and demoted the HHS parent OpDiv, which is what HHS-wide access keys off of. Gave it the same treatment - pointer field, only set on update when supplied, and the insert defaults it to false. Create still always comes up active.

Pushed in b29c278.

jsos3-cms (approved):
Looks great, and I love the clean tests :)

PR #335, opened 2026-06-10
fix(#328): refresh DB credentials and retry on auth failure after secret rotation

### LINKED ISSUE #328: API crashes on RDS secret rotation: DB credentials cached for process lifetime, never refreshed
## Description

The API caches its database credentials once at process startup and never refreshes them. When the RDS-managed master secret auto-rotates (every 7 days), the running task still holds the old password. The next database connection it opens fails SASL auth with `SQLSTATE 28P01`, and because the connect path treats this as fatal, the process exits. ECS then restarts the task, which re-fetches the now-current secret and recovers. The result is a brief self-healing outage at every rotation.

In `backend/internal/db/conn.go`, `getDbCreds()` loads the secret inside a `sync.Once`, so `dbSecret` is fetched exactly once per process lifetime. `Conn()` opens a fresh `pgx.Conn` per call using those cached credentials. After a rotation there is no code path that re-reads the secret or retries; the stale password is reused until the process restarts. The original implementation (#23) was described as "refresh when outdated," but the current code does not refresh, and #161 changed the password-error path to exit.

This is the same failure mode tracked for the Decoder Ring app in CMS-Enterprise/zt-decoder-ring#169.

## Steps to Reproduce

1. Deploy the API to an environment whose database credentials come from a rotation-enabled Secrets Manager secret (dev, impl, prod all qualify).
2. Leave the task running without a redeploy.
3. Wait for the RDS-managed secret to auto-rotate (7-day schedule).
4. Trigger any database-backed request after the rotation.

## Expected Behavior

When a database connection fails authentication, the API should re-fetch the credentials from Secrets Manager and retry the connection once before giving up. A routine secret rotation should not take the service down or require a task restart to recover.

## Actual Behavior

The task that was running across the rotation crashes on the first post-rotation connection:

```
could not connect to db failed to connect to `user=ztmfAdmin database=ztmf`: failed SASL auth: FATAL: password authentication failed for user "ztmfAdmin" (SQLSTATE 28P01)
```

ECS restarts the task ~35s later and it recovers on its own, so the impact is a short blip rather than a sustained outage.

## Environment

dev (observed 2026-05-29 11:17 UTC), impl, and prod. All use RDS-managed secret rotation on a 7-day schedule. Confirmed against the dev secret `rds!cluster-dafb0010-...` (rotation enabled, `AutomaticallyAfterDays: 7`).

## Acceptance Criteria

- Database credentials are re-fetched from Secrets Manager when a connection fails authentication, not cached for the full process lifetime.
- A `28P01` (or equivalent auth failure) triggers a single re-fetch + retry rather than a process exit.
- A secret rotation no longer causes a task crash or a gap in availability.
- Behavior is verified against an environment after a real or forced rotation.

## Severity

Low-to-moderate. Currently self-healing in ~35s via ECS restart, and historically masked entirely because frequent dev deploys reset the cached secret before each rotation could catch a long-lived task. With the ZTMF Insights extraction now complete, ztmf core sees far fewer deploys, so long-running tasks will reliably hit this at each weekly rotation. The same latent bug applies to prod.

## Affected Component

`backend/internal/db/conn.go` (`Conn`, `getDbCreds`, the `sync.Once` secret cache).

## Logs / Error Messages

```
2026-05-29T11:17:48 could not connect to db failed to connect to `user=ztmfAdmin database=ztmf`: 10.202.152.6:5432 (ztmf.cluster-...us-east-1.rds.amazonaws.com): failed SASL auth: FATAL: password authentication failed for user "ztmfAdmin" (SQLSTATE 28P01)
2026-05-29T11:18:23 initializing config...
2026-05-29T11:18:23 Fetching secret from arn:aws:secretsmanager:...:secret:rds!cluster-dafb0010-... ...
2026-05-29T11:18:23 executing migrations...
2026-05-29T11:18:23 dev environment listening on 443
```

## Workaround

The task self-heals on ECS restart. A manual `aws ecs update-service --force-new-deployment` forces an immediate restart and secret re-fetch if a rotation lands during a quiet period and an operator wants to clear it without waiting.

voidspooks:
## Summary

Closes #328.

The API caches its database credentials once at process startup (`sync.Once` in `backend/internal/db/conn.go`) and never refreshes them. When the RDS-managed master secret auto-rotates (7-day schedule), the running task still holds the old password. The next connection fails SASL auth (`SQLSTATE 28P01`), and the error path (`trapError`, added in #161) called `log.Fatal`, exiting the process. ECS then restarted the task and re-fetched the now-current secret, so each rotation became a brief self-healing outage.

## What changed

**`internal/db/conn.go` — reactive refresh + single retry**
- Connect logic is split into a `connect(ctx)` helper; `Conn(ctx)` wraps it. On an auth-failure connect, `Conn` forces a re-fetch of the secret from Secrets Manager and retries the connection **once**.
- `isAuthError(err)` matches `28P01` via `errors.As`, so it's recognized even when pgx wraps the `*pgconn.PgError`.
- `refreshDbCreds(ctx)` calls `Secret.Refresh(ctx)` to re-read the post-rotation password.
- The retry only runs on the secret-backed path (`Db.SecretId != ""`); local env-var credentials have nothing to re-fetch. A genuinely wrong password fails the retry too and surfaces as a normal error.

**`internal/model/errors.go` — stop killing the process**
- The `28P01` case no longer `log.Fatal`s (there was a literal TODO there pointing at this). With `Conn` handling rotation reactively, a `28P01` reaching `trapError` means even the retry failed, so it now returns `ErrDbConnection`.

## Tests

- `internal/db/conn_test.go` — `isAuthError` matrix (bare/wrapped 28P01 → true; other pg codes, non-pg, nil → false).
- `internal/model/errors_test.go` — `trapError` mappings, including the regression guard that `28P01 → ErrDbConnection` (the test completing at all proves the process no longer exits).

`go build ./...`, `go vet`, and `go test -short ./...` all green locally (run in the `golang:1.25.11` container).

## Acceptance criteria

- [x] DB credentials are re-fetched from Secrets Manager on an auth failure rather than cached for the full process lifetime
- [x] A `28P01` triggers a single re-fetch + retry rather than a process exit
- [ ] Verified against an environment after a real or forced rotation — **remaining manual step** (needs AWS access; see issue)

## Notes

Draft pending the live-rotation verification. CI checks that depend on repo secrets (Snyk, etc.) will show red on a fork PR — expected per our workflow; maintainers re-run after pulling the branch.

### DIFF
--- a/backend/internal/db/conn.go
+++ b/backend/internal/db/conn.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"context"
+	"errors"
 	"log"
 	"sync"
 
@@ -21,7 +22,38 @@ type dbCreds struct {
 	Password string
 }
 
+// Conn opens a fresh connection to the database. RDS-managed credentials in
+// Secrets Manager auto-rotate (7-day schedule); the secret we cache at startup
+// goes stale at each rotation and the next connect fails SASL auth (28P01).
+// Rather than let that take the process down, an auth failure forces a re-fetch
+// of the secret and a single retry, so a routine rotation is ridden out in
+// place instead of relying on an ECS task restart to recover. A genuinely wrong
+// password (not a rotation) fails the retry too and surfaces as a normal error.
 func Conn(ctx context.Context) (*pgx.Conn, error) {
+	conn, err := connect(ctx)
+	if err == nil {
+		return conn, nil
+	}
+
+	// Only the secret-backed path can recover by refreshing; local env-var
+	// credentials have nothing to re-fetch, so do not retry there.
+	if !isAuthError(err) || config.GetInstance().Db.SecretId == "" {
+		return nil, err
+	}
+
+	log.Println("db authentication failed; refreshing credentials and retrying once")
+	if rerr := refreshDbCreds(ctx); rerr != nil {
+		log.Println("could not refresh db credentials", rerr)
+		return nil, err
+	}
+
+	return connect(ctx)
+}
+
+// connect builds the connection config from the current credentials and opens a
+// single connection. Split out from Conn so the credential-refresh retry can
+// re-run the whole path against the freshly fetched secret.
+func connect(ctx context.Context) (*pgx.Conn, error) {
 	creds, err := getDbCreds()
 	if err != nil {
 		log.Println("could not get db credentials")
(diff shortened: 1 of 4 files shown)


danielbowne:
Re-homed onto an org branch as #337 so the full CI pipeline can run (fork PRs do not get the org secrets for Snyk/build/deploy). Same change merged up to current main, plus one addition: a mutex on the cached Secret, since making the refresh reactive on the per-request path means a rotation can have many goroutines reading and writing the cached secret at once. Continuing review and merge there. Thanks @voidspooks - clean fix.

PR #337, opened 2026-06-11
fix(#328): refresh DB credentials and retry on auth failure after secret rotation

danielbowne:
Re-homes #335 (originally authored by @voidspooks) onto an org branch so the full CI pipeline runs, and merges current main. Fork PRs do not get the org secrets Snyk and the build/deploy steps need.

## What this fixes (ztmf#328)

The API caches the RDS-managed DB secret at startup. RDS rotates it on a ~7-day schedule; afterward the cached password is stale, the next connection fails SASL auth (28P01), and the old code `log.Fatal`'d on 28P01 - so every rotation took the process down until ECS restarted it.

Cameron's change:
- `db/conn.go`: `Conn()` retries once on a 28P01 auth error when a secret is configured, re-fetching the secret first (`refreshDbCreds` -> `Secret.Refresh`), so a routine rotation is ridden out in place. Env-var (local) credentials are not retried since there is nothing to re-fetch.
- `model/errors.go`: the 28P01 case returns `ErrDbConnection` instead of `log.Fatal`, so a genuinely-wrong password surfaces as a normal connection error rather than killing the process.
- Adds unit tests for the error classification (`isAuthError`, `trapError`), including wrapped-error cases.

## Added during re-home: synchronize the cached secret

Making the refresh reactive on the per-request path means a rotation can have many goroutines reading (`Value`/`Unmarshal`) and writing (`Refresh`) the cached `Secret` at once, and those fields had no synchronization (a read race already existed in `Value`'s proactive-refresh path; the reactive trigger widens it). Added a mutex on `Secret` guarding `secret`/`metadata` across the fetch RPC, so a refresh storm serializes instead of racing the pointers. `Value`/`Refresh` lock and delegate to an unexported `refreshLocked` to avoid re-locking.

## Testing

- `go build`, `go vet`, and the touched-package unit tests pass under `-race`.
- Error-classification regression is covered; the live refresh-and-retry flow and the concurrent-refresh path are not yet unit-tested (they need an injectable Secrets Manager / connect seam) - tracked as a follow-up under ztmf#328.

## Note

A single retry rides out the modeled failure. Right at a rotation boundary a refreshed secret can still briefly mismatch the cluster, so expect the occasional single failed request at rotation time - the process stays up, which is the goal.

Closes #335.

### DIFF
--- a/backend/internal/db/conn.go
+++ b/backend/internal/db/conn.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"context"
+	"errors"
 	"log"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


MackOverflow (dismissed):
looks good Danny, two small things, neither of them blocking

 re: the dbSecret fix: dbSecret is written inside once.Do but also read bare in refreshDbCreds and getDbCreds from concurrent request goroutines. The new mutex protects the Secret's fields, but the pointer itself isn't synchronized. In practice the once happens at startup so it's benign, but since this PR is adding -race-clean synchronization anyway, an atomic.Pointer or moving the nil-check inside once.Do would close it fully.

Tiny nit: on the retry path, if the second connect also fails with 28P01 the log line is identical to the first failure ("could not connect to db"). A one-word distinction ("retry after refresh failed") would make rotation-boundary incidents easier to read in CloudWatch. Take it or leave it.

danielbowne:
Thanks Mack - both addressed in 7f2b2c8:

- dbSecret pointer: dropped the bare `if dbSecret == nil` pre-check and call `once.Do` unconditionally, so the pointer is only read after once.Do establishes the happens-before. Added a post-once nil check so a failed init still surfaces to every caller. Race detector clean.
- Retry log: the post-refresh retry now logs a distinct "db connection still failing after credential refresh" instead of reusing connect()'s generic line, so rotation-boundary failures read clearly in CloudWatch.

Re-requested your review when you have a sec.

MackOverflow (approved):
Appreciate the fix!

PR #338, opened 2026-06-11
Expand empire seed data: multi-OpDiv org, 3-year history, and audit trail

danielbowne:
Gives a fresh local dev environment realistic sample data and makes the OpDiv and audit features exercisable without pulling production data.

## What's added

- **Multi-OpDiv org.** Six Imperial-branch OpDivs (Navy, Army, Starfighter Corps, ISB, Intelligence, Sienar) with 10 new systems and 12 officers scoped across them as ISSO / OPDIV_ADMIN / OPDIV_READONLY_ADMIN, so OpDiv-scoped RBAC isolation can be seen in the UI (an OWNER sees all 16 systems; a branch OPDIV_ADMIN sees only their branch).
- **Three-year history.** Data calls FY2022-FY2025, with each new system scored every cycle and maturity climbing year over year (Traditional toward Advanced).
- **Richer questionnaire.** Three new `datacenterenvironment`s (Ground-Assault, Surveillance-Net, Shipyard-RnD), each a full six-pillar function set with four maturity options, plus per-pillar questions.
- **Audit trail.** `last_edited_at` / `last_edited_by` are derived from the events table, which raw SQL seed data bypasses, so seeded scores previously showed a blank "last edited by" in the dashboard footer. This seeds one `updated` event per new score, attributed to the system's assigned officer, so the audit display and features work on a fresh local DB.

## Tests

- `TestFindScoresResolvesSeededAuditFieldsIntegration` covers the read of audit fields from a pre-existing (seed) event. The existing audit tests all create the event in-test, so the historical read path that the seed and the UI footer depend on was untested.
- `make test-e2e` passes 111/111 and `make test-integration` passes. All data uses new IDs and new OpDivs only; existing systems (1001-1006), data calls (1-3), and scores (9001-9030) are untouched, so the OpDiv-scoped and aggregate E2E assertions are unaffected.

## Note

The score notes carry some lightly buried thematic flavor; they remain plausible as assessment text and contain no real data.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -483,6 +483,313 @@ INSERT INTO public.system_enrichment (fisma_uuid, payload, synced_at) VALUES (
     '2026-05-20 00:00:00+00'
 ) ON CONFLICT (fisma_uuid) DO NOTHING;
 
+-- ============================================================================
+-- EXPANDED EMPIRE FIXTURE (dev/demo): multi-OpDiv org, more systems and
+-- officers, a 4-cycle (FY2022-FY2025) data-call history with maturity that
+-- climbs over time, and three additional questionnaires (datacenterenvironments).
+--
+-- Everything below uses fresh ID ranges and NEW OpDivs only. It deliberately
+-- does not touch systems 1001-1006, datacalls 1-3, or scores 9001-9030, which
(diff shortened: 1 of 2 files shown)


a-dipietro (approved):
lgtm!

PR #340, opened 2026-06-11
feat(#260): auto-generate OpenAPI spec from Go annotations via swaggo/swag

### LINKED ISSUE #260: Auto-generate OpenAPI spec from Go code using swaggo/swag
## Problem

The OpenAPI spec (`backend/openapi.yaml`) is manually maintained and has drifted from the actual API multiple times:
- `GET /datacalls/latest` was missing
- `DELETE /fismasystems/{fismasystemid}` was missing
- FismaSystem decommission fields were missing from the schema

As we add more endpoints, this will only get worse.

## Proposal

Adopt [swaggo/swag](https://github.com/swaggo/swag) to auto-generate `openapi.yaml` from Go handler annotations and struct definitions. This ensures the spec always matches the actual code.

### 1. Add swag annotations to existing handlers
Add comment blocks above each controller function:
```go
// @Summary Get the latest data call
// @Tags datacalls
// @Produce json
// @Success 200 {object} DataCall
// @Router /datacalls/latest [get]
func GetLatestDataCall(w http.ResponseWriter, r *http.Request) {
```

### 2. Add general API info annotation to main.go
```go
// @title ZTMF API
// @version 1.0
// @description API for Zero Trust Maturity Framework
// @BasePath /api/v1
```

### 3. Makefile target
```
make generate-openapi    # runs swag init, outputs openapi.yaml
```

### 4. Git pre-commit hook
If any file in `backend/cmd/api/internal/controller/` or `backend/internal/model/` is staged, automatically run `swag init` and stage the updated spec. This way the spec can never drift.

### 5. CI validation (optional)
Add a CI step that runs `swag init` and fails if the generated spec differs from what's committed — catches cases where someone bypasses the hook.

## Benefits
- Spec always matches code — no more drift
- Annotations double as inline docs for handlers
- Works with existing gorilla/mux + manual handlers — no rewrite needed
- Incremental adoption: annotate handlers one at a time

## Scope
- Annotate all ~40 handler functions in `backend/cmd/api/internal/controller/`
- Add `make generate-openapi` target
- Add pre-commit hook for controller/model changes
- Remove manually maintained `backend/openapi.yaml` once migration is complete

voidspooks:
Closes #260.

## Problem

`backend/openapi.yaml` was hand-maintained and had drifted from the real API multiple times (missing `GET /datacalls/latest`, missing `DELETE /fismasystems/{id}`, missing schema fields). As endpoints grow, manual upkeep keeps losing.

## What this does

Generates `openapi.yaml` from [swaggo/swag](https://github.com/swaggo/swag) v2 annotations on the handler functions, so the spec can never silently drift again — and enforces it in CI.

- **swag v2 tool dependency** pinned in `backend/go.mod` (`go tool swag`), so CI and local runs produce byte-identical output.
- **General API info** block in `cmd/api/main.go`; a generic **`apiResponse[T]`** wrapper (`controller/docs.go`) documents the standard `{ data, error }` envelope, e.g. `@Success 200 {object} apiResponse[model.DataCall]`.
- **All 48 routes** annotated across the 14 controller files (summaries, tags, path/query params, request bodies, `@Security bearerAuth`, failure codes).
- **`make generate-openapi`** regenerates the spec; the new **`OpenAPI-Spec-Check`** CI job (in `analysis.yml`) regenerates and fails the build if the committed spec is out of date.
- **`redocly.yaml`** lint config — passes clean as OpenAPI 3.1; a couple of default rules are relaxed where swag's output style requires it, each with an inline reason.
- Workflow documented in `backend/README.md` (generate, don't hand-edit).

## Coverage / safety

- Path+method coverage of the generated spec **exactly matches** the previous curated spec (48 operations, zero dropped, zero spurious) — verified by a set diff.
- `go build ./...` and `go test -short ./...` green.
- Generation is deterministic: a second regen is byte-identical (the CI gate passes).

## Notes for reviewers

- **swag v2 is `v2.0.0-rc5`** — it's the only swag release that emits OpenAPI 3.x (v1 emits Swagger 2.0). Flagging the RC status; the spec moves from `3.0.0` → `3.1.0`. There are no automated consumers of `openapi.yaml` in this repo, so the version bump is low-risk.
- The 6 shared create/update handlers (one Go func registered for both `POST /x` and `PUT /x/{id}`) share a single annotation block, so the collection `POST` inherits the item `PUT`'s `{id}` path param (`required: false`). Surfaced as a redocly warning rather than splitting runtime handlers purely for doc fidelity.
- `SystemEnrichment.Payload` (`json.RawMessage`) carries a `swaggertype:"object"` tag so swag can resolve the stdlib type — the only model change.

Snyk / secret-dependent checks will show red on this fork PR (no repo secrets); that's expected.


danielbowne:
@voidspooks Nice approach,  this is the right way to kill the spec drift.  I pulled the branch and ran it: the generator is deterministic (the committed openapi.yaml regenerates byte-identical),  all 48 routes have exactly matching @Router annotations with no gaps or extras,  the CI drift check actually fails on a mismatch,  and it builds clean.  The annotation-as-source-of-truth model holds up.

One thing before this can merge: the branch conflicts with main now,  since it predates the OpDiv and dual-IdP work that landed.  After you rebase,  a few of the newly merged endpoints will need annotations,  and the new GET /api/v1/auth/lookup route has to be marked unauthenticated (no @Security bearerAuth) so it does not inherit the global auth in the spec.  Then re-run make generate-openapi and commit so the drift check stays green.

Three smaller things,  none blocking:

1. SaveFunction (functions.go) and SaveOpDiv (opdivs.go) annotate the path id as required: true,  but those same handlers also serve POST /functions and POST /opdivs,  which have no {id} in the path.  That renders a required path param on a path that cannot supply one.  SaveDataCall and SaveUser use false for the same shared-handler pattern,  so matching them fixes it.

2. GetEvents (events.go) documents fismasystemid,  scoreid,  datacallid,  and questionid as top-level query params,  but those bind as nested payload.xxx through the schema decoder,  so as written they would not bind.  It also drops the action and resource filters that do work.  Documenting action,  resource,  and payload matches what the handler actually decodes.

3. redocly.yaml is committed but no workflow runs redocly lint,  so right now it enforces nothing.  The drift check only proves the spec matches the annotations,  not that the annotations are correct (both items above generate cleanly and pass the gate).  Either wire the lint into CI or drop the config so it does not imply enforcement that is not there.

Happy to re-review the rebase delta once the new endpoints are annotated.

voidspooks:
Thanks for the thorough pull-through, @danielbowne. Rebased onto `main` and worked the review — branch is back to mergeable.

**Blocking:**
- **Rebased** past the OpDiv (#333) and dual-IdP (#341) merges. Only conflicts were generated/lock files: resolved `go.sum` as a union + `go mod tidy`, regenerated `openapi.yaml`.
- **`GET /api/v1/auth/lookup`** annotated, deliberately with **no `@Security bearerAuth`** — it renders as the one public route, no inherited auth.

**Smaller items:**
1. `SaveFunction` / `SaveOpDiv` path id `required: true → false` to match the `SaveDataCall`/`SaveUser` shared-handler pattern. Confirmed the PUT ops for `/functions/{functionid}`, `/opdivs/{opdiv_id}` and the reference `/datacalls/{datacallid}` now render identically.
2. `GetEvents` — documented what the schema decoder actually binds: top-level `action`/`resource` plus nested `payload.fismasystemid|scoreid|datacallid|questionid`. Dropped the top-level `fismasystemid/...` that never bound.
3. **Wired `redocly lint` into the `OpenAPI-Spec-Check` CI job** (pinned `redocly/cli:2.32.2`) + added a `make lint-openapi` target — so it enforces, not just describes. Kept `security-defined` a hard error for *every* route (a new handler that forgets `@Security` now fails CI); acknowledged only the intentional public route in a scoped `backend/.redocly.lint-ignore.yaml`. Left the shared-handler `path-parameters-defined` and `info-license` findings as visible warnings rather than silencing them.

Verified: regeneration is deterministic (byte-identical, drift check green), `go build` + `go test -short` pass, `redocly lint` is clean (0 errors, 8 warnings). One open call for you: on the redocly item I went with wiring it in rather than dropping the config, since it matches the spirit of the PR — shout if you'd rather keep CI lean and drop it instead. Ready for the re-review when you have cycles.

danielbowne:
Thanks @voidspooks, the rebase and review fixes all check out (lookup annotated public, shared-handler path params fixed, GetEvents params corrected, redocly lint wired and pinned; regeneration is byte-identical and the drift check is green).

Moving this onto an in-repo branch as #348 so the Snyk and other org-secret CI gates run, which they cannot on a fork PR. Your three commits are pushed as-is, so authorship is unchanged. Closing this in favor of #348; the review continues there.

PR #342, opened 2026-06-15
Close leaked DB connections (Aurora connection exhaustion in prod)

danielbowne:
## Summary

Fixes a database connection leak that exhausted the Aurora connection limit in prod today. db.Conn() opens a dedicated pgx connection on every call (there is no shared pool), and the central model helpers plus a few raw call sites never closed the connection, so every query leaked one. Two active users were enough to pile up around 197 open connections, peg CPU, and time out requests. In the UI that showed as intermittently missing or slow data, like the OpDiv column blanking on first load.

## Root cause

The leak has been in the data layer since the open-per-call, no-close pattern was first written (mid-2024). It stayed hidden because Aurora reaped idle connections faster than they leaked. The OpDiv RBAC work added more DB round trips per authenticated request (auth middleware, then the user lookup, then the OpDiv grant queries), which pushed the leak rate past the reap rate, so connections piled up until the cluster hit its limit. The recent multi-OpDiv change triggered it but was not the root.

## Fix

Added a deferred connection close at every runtime site that opens one:

- model.go query() and queryRow(), which back most request-path queries
- scores.go copyPreviousScores, an Exec site that leaked
- fismasystems.go and users.go transaction paths, which rolled back the tx but never closed the connection. The close is declared before the tx rollback defer so it runs after the transaction resolves (defers are LIFO)
- populate.go, the local seed path

The tern migrator connection in migrations.go is left open on purpose. It is a startup-only connection the migrator owns for the life of the process.

## Not included

A connection pool (pgxpool) is the real fix for the per-query connect cost, which is the remaining slowness and the high DB CPU, and it will matter at HHS scale. That change also touches the credential rotation path, so it is a separate follow-up. This PR only stops the leak.

## Testing

- make test-integration (isolated DB) passes
- make test-e2e (Emberfall) passes
- go build, staticcheck, and go vet are clean
- Confirmed every runtime connection site closes once, the tx ordering is correct, there is no double close or use after close, and closing on an already-canceled context still releases the socket in our pinned pgx version

## Context

Replaces the earlier partial fix that only closed connections in two scores functions (#322, refs #321). This closes the leak across the whole data layer.

### DIFF
--- a/backend/cmd/api/internal/migrations/populate.go
+++ b/backend/cmd/api/internal/migrations/populate.go
@@ -21,6 +21,7 @@ func populate(path string) error {
 	if err != nil {
 		return err
 	}
+	defer conn.Close(ctx)
 	_, err = conn.Exec(ctx, string(sql))
 	return err
 }
--- a/backend/internal/model/fismasystems.go
+++ b/backend/internal/model/fismasystems.go
@@ -300,6 +300,10 @@ func ReactivateFismaSystem(ctx context.Context, input ReactivateInput) (*FismaSy
 	if err != nil {
 		return nil, trapError(err)
 	}
+	// Close the dedicated connection last (after the tx rollback/commit). defers
+	// run LIFO, so declaring this before the tx defer guarantees the transaction
+	// is resolved before the connection is closed.
+	defer conn.Close(ctx)
 
 	tx, err := conn.Begin(ctx)
 	if err != nil {
--- a/backend/internal/model/fismasystems.go
+++ b/backend/internal/model/fismasystems.go
@@ -300,16 +300,19 @@ func ReactivateFismaSystem(ctx context.Context, input ReactivateInput) (*FismaSy
 	if err != nil {
 		return nil, trapError(err)
 	}
-	// Close the dedicated connection last (after the tx rollback/commit). defers
-	// run LIFO, so declaring this before the tx defer guarantees the transaction
-	// is resolved before the connection is closed.
-	defer conn.Close(ctx)
 
 	tx, err := conn.Begin(ctx)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


MackOverflow (dismissed):
This is a solid, focused hotfix. Small diff, well-scoped, no scope creep. The LIFO defer ordering for tx sites is the tricky part and it's handled correctly. One thing worth calling out:

backend/internal/model/fismasystems.go and users.go — LIFO ordering is correct but fragile

The LIFO ordering comment is good. One thing worth noting: this pattern is implicit — a future dev adding another defer between conn.Close and the tx rollback could silently break the ordering. Might be worth a slightly more defensive comment like "do not add defers between this line and the tx rollback defer below." Up to you if that's over-engineering it.

danielbowne:
Thanks @MackOverflow. Agreed on the fragility. The two-defer LIFO ordering was implicit and a defer added between them later would have broken it silently.

Rather than document the constraint with a comment, I made it structural. Both tx sites (fismasystems.go and users.go) now roll back and then close in a single deferred func, so the order is enforced by the code and nothing can interleave between them. The connection is closed explicitly on the Begin-error path. Pushed in ea8329e.

One nuance worth recording: even if the order had ever flipped, it was not a data-correctness risk. Closing a connection with an open transaction makes the server abort that transaction, and a rollback on an already-closed connection just returns an error that we ignore. So the worst case was a less-clean abort, not a half-committed transaction. The single defer still removes the footgun.

The pgxpool follow-up will eventually remove the manual connection handling here entirely, since pool.Begin releases its connection on commit or rollback. Re-requested your review.

MackOverflow (approved):
bellissimo!

PR #343, opened 2026-06-16
Pool DB connections with pgxpool (fixes Aurora connection exhaustion under normal use)

danielbowne:
## Summary

The data layer had no connection pool. db.Conn() opened a brand-new connection (TCP, TLS, SCRAM auth) on every query. Once the OpDiv RBAC and dual-IdP work made every authenticated request run a DB user lookup, normal use opened a burst of concurrent connections that hit the Aurora connection limit. New connections then died mid-handshake (failed SASL auth: context canceled), which showed in the UI as the OpDiv column going empty and edits failing to save. One active user was enough to trigger it, so this is a design problem, not load.

This replaces the per-call connection model with a shared, bounded pgxpool.

## What changed

- db.Conn acquires a connection from a shared pool and returns it; callers release it back to the pool when done instead of closing it. The pool caps connections at 16, well under the cluster limit, and reuses them, so a burst of concurrent requests no longer opens a connection each.
- Credentials are supplied per new connection, so the 7-day RDS secret rotation is picked up as connections recycle (MaxConnLifetime 50m). An auth failure on acquire refreshes the secret and retries once, preserving the rotation handling from the earlier credential-refresh fix.
- The tern migrator keeps a dedicated single connection (MigrationConn), since it needs a plain connection and is not pool-managed.
- Pool is built once with a background context so a canceled request cannot poison the one-time init.

## Relationship to the earlier fix

The previous change (#342) added a close to every connection site to stop a slow leak. That was necessary but did not stop the concurrent burst, because closing a connection after use does nothing about many connections being opened at the same time. The pool is what actually bounds and reuses connections. With the pool, connections are released back rather than closed, which is the correct pattern.

## Validation

- make test-integration (isolated DB) passes
- make test-e2e (Emberfall) passes
- go build, staticcheck, and go vet are clean
- Reviewed for pgx pool correctness: every acquire has a matching release on all paths including the transaction error paths, the transaction sites release exactly once, rows are fully read before release, MaxConns is a hard cap, and the rotation path recovers across the 7-day cycle

## Sizing notes

MaxConns 16 against the observed cluster connection limit leaves headroom for the migrator connection, the ops task, and pgAdmin sessions. MinConns 2 keeps a warm pair so the first requests after idle do not each pay a full handshake. These are conservative for a low-traffic app and can be tuned later if needed.

### DIFF
--- a/backend/cmd/api/internal/migrations/migrations.go
+++ b/backend/cmd/api/internal/migrations/migrations.go
@@ -53,7 +53,7 @@ func Run() {
 func getMigrator() *migrate.Migrator {
 	if migrator == nil {
 		once.Do(func() {
-			conn, err := db.Conn(context.Background())
+			conn, err := db.MigrationConn(context.Background())
 			if err != nil {
 				log.Fatal(err)
 				return
--- a/backend/cmd/api/internal/migrations/populate.go
+++ b/backend/cmd/api/internal/migrations/populate.go
@@ -21,7 +21,7 @@ func populate(path string) error {
 	if err != nil {
 		return err
 	}
-	defer conn.Close(ctx)
+	defer conn.Release()
 	_, err = conn.Exec(ctx, string(sql))
 	return err
 }
--- a/backend/go.mod
+++ b/backend/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/lann/builder v0.0.0-20180802200727-47ae307949d0
 	github.com/stretchr/testify v1.11.1
 	github.com/xuri/excelize/v2 v2.9.0
+	golang.org/x/time v0.15.0
 )
 
 require (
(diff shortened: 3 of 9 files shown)


MackOverflow (dismissed):
Looks really good Danny, I am catching a handful of minors: 

sslmode=allow → sslmode=require — should be a one-word change; no reason to allow plaintext fallback on RDS.
sslmode=allow — this is weaker than sslmode=require or sslmode=verify-full. For an RDS/Aurora instance, requiring TLS is a standard hardening step. allow means the connection will fall back to plaintext if TLS negotiation fails, which may not be intentional for a production database.
Verify golang.org/x/time direct dependency promotion — confirm it's intentional and not a go mod tidy artifact that could be reverted.
DSN string concatenation — low risk for hostname/port/dbname, but consider fmt.Sprintf at minimum for readability, or pgx's ConnConfig struct fields directly to avoid DSN parsing entirely.

danielbowne:
Thanks,  addressed in the latest push.

sslmode: agreed in principle,  but `require` everywhere would break local and CI,  where the Docker Postgres has no TLS configured,  so it is now env-gated: `require` in deployed environments (no plaintext fallback to RDS/Aurora),  `allow` in local/test.

golang.org/x/time: confirmed intentional,  not a tidy artifact.  It is imported directly by the rate limiter (auth/ratelimit.go and router.go use golang.org/x/time/rate);  it came in with the dual-IdP work on main,  and the rebase's go mod tidy correctly promoted it from indirect to direct.

DSN: switched to fmt.Sprintf,  and the single-connection path now passes username and password through ConnConfig fields instead of the DSN string,  so a credential with special characters cannot break parsing.

MackOverflow (approved):
looks awesome Danny!

PR #344, opened 2026-06-16
Return OpDiv grants inline on the users list (fix users-table load N+1)

### LINKED ISSUE #336: Include OpDiv grants in the /users list response
### Feature Type

General Feature

### Workflow Description

The `/users` list endpoint currently returns each user without their OpDiv grants (the field is always null in the list response). The UI's user admin table needs grants to render the OpDivs column, so today it calls `GET /users/{id}` once per user after the list loads to pull grants from the detail endpoint. On a list of N users that is N extra requests per page load.

If the list response includes each user's granted OpDiv IDs (same shape as the detail endpoint), the UI can read grants straight off the list payload and drop the per-user fan-out entirely.

Scoping should match the existing list behavior: an OPDIV_ADMIN caller should only see grants within their own visibility, same as the detail endpoint enforces today.

### Acceptance Criteria

- [ ] `GET /users` list response includes each user's OpDiv grants (same field shape as `GET /users/{id}`)
- [ ] Grant visibility in the list response matches the detail endpoint's RBAC scoping
- [ ] No regression in list endpoint response time for typical user counts
- [ ] ztmf-ui can remove the per-user `fetchUserOpDivs` fan-out (tracked separately on the UI side)

Refs CMS-Enterprise/ztmf-ui#379. Came out of review on CMS-Enterprise/ztmf-ui#378, which flagged the N+1 in the user table.

danielbowne:
## Summary

The users table was slow to load its OpDiv column. `FindUsers` (the `GET /api/v1/users` list query) omitted the user's OpDiv and FISMA grants, so the frontend backfilled them by calling `/users/{id}` once per user. With ~600 users that is hundreds of round-trips trickling through the browser's per-host connection limit, which is the delay. The list query itself is a few milliseconds.

This loads the grants inline on the list, using the same correlated subqueries `findUser` already uses for the detail endpoint, so the table renders the OpDiv column straight from the list response with no per-row fetch.

## Why this is cheap

Measured against real prod-scale data locally (595 non-deleted users): the list query with both subqueries runs in a few milliseconds. Each subquery is a composite-PK index-only scan on the junction table (one row group per user), so it does not add a meaningful cost and there is no N*M join blowup. The endpoint returns in about 11ms.

## Coordination

The frontend (ztmf-ui) drops its per-user `fetchUserOpDivs` bulk loop and reads `assignedopdivids` from each list row, mapping ids to codes with the opdiv map it already builds once. The frontend change is written to fall back gracefully if the field is absent, so the two can deploy in either order.

## Scope

This fixes the page-load read N+1 only. The OpDivGrantModal per-item write loop is lower severity (user-triggered write, small N, now pool-bounded) and is tracked separately; it needs a backend batch-grant endpoint.

## Validation

- Verified against real prod-scale local data: list returns `assignedopdivids` for all 595 users, endpoint ~11ms
- `make test-integration` and `make test-e2e` pass
- `go build`, staticcheck, `go vet` clean
- Emberfall `/users` list assertions are status/header only, so the additive fields do not break them

Closes #336

### DIFF
--- a/backend/internal/model/users.go
+++ b/backend/internal/model/users.go
@@ -318,11 +318,24 @@ func FindUsers(ctx context.Context, fui *FindUsersInput) ([]*User, error) {
 
 	// Explicit column list (vs SELECT *) so new schema columns that do not
 	// have a corresponding User struct field do not break pgx struct scans.
-	// AssignedFismaSystems and AssignedOpDivIDs are populated on a per-user
-	// detail lookup (findUser), not on the list view where they would force
-	// extra joins for no UI benefit.
+	// AssignedOpDivIDs is loaded with the same correlated subquery findUser
+	// uses, so the users table renders the OpDiv column straight from this
+	// list response instead of the client backfilling grants with one
+	// /users/{id} call per row (an N+1 that made the page crawl). The subquery
+	// is a composite-PK index-only scan; the whole list is a few milliseconds.
+	// AssignedFismaSystems is intentionally not loaded here: it is json:"-"
+	// (server-side authz only, used by findUser) and the list view neither
+	// serializes nor needs it, so computing it would be wasted work.
 	sqlb := stmntBuilder.
-		Select("users.userid", "email", "fullname", "role", "deleted", "identity_provider").
+		Select(
+			"users.userid",
+			"users.email",
+			"users.fullname",
+			"users.role",
+			"users.deleted",
+			"users.identity_provider",
+			"(SELECT ARRAY_AGG(opdiv_id) FROM users_opdivs WHERE userid = users.userid) AS assignedopdivids",
+		).
 		From("public.users").
 		Where("deleted=?", fui.Deleted)
 


MackOverflow (approved):
Looks good, one nit, not sure if its worth addressing, up to you Danny: 

NULL handling for users with no OpDiv grants — ARRAY_AGG returns NULL (not []) when there are no matching rows. The AssignedOpDivIDs field is typed []*int32, and pgx will scan a SQL NULL into a Go nil slice. That's probably fine if callers already guard against nil, but it's an inconsistency worth verifying: does the detail endpoint (findUser) return [] or null in JSON for a user with no opdiv grants? If they differ, the frontend's "fall back gracefully" logic may need to handle both shapes.

danielbowne:
Verified,  thanks.  The list (FindUsers) and detail (findUser) paths use the same subquery,  so they behave identically: a user with zero OpDiv grants gets ARRAY_AGG returning SQL NULL,  which pgx scans into a nil slice and serializes as JSON null on both responses.  So there is no shape divergence between them for the frontend to special-case.

On the frontend the gate is key presence (`'assignedopdivids' in u`),  and it maps the value through `?? []`,  so null vs an array is handled either way.

If we would rather the contract always be an array,  I can COALESCE(ARRAY_AGG(...), '{}') on both queries to keep them consistent.  Happy to do that,  otherwise leaving it as-is since it is consistent and already handled.

Issue #317, opened 2026-05-19
Bump GitHub Actions workflows to Node.js 24 and pin third-party actions to commit SHAs


danielbowne:
## Type

Chore

## Description

GitHub is forcing Actions runners off Node.js 20. The default runner switches
to Node.js 24 on June 2, 2026, and Node.js 20 is removed from runners entirely
on September 16, 2026. Every workflow run on this repo currently emits the
deprecation warning across `Check for Changes`, `Analysis / lint terraform`,
`Analysis / snyk test`, `Analysis / snyk iac test`, and the deploy workflows.
Named offenders include `actions/checkout@v4`, `actions/setup-go@v5`,
`terraform-linters/setup-tflint@v4`, and `aws-actions/configure-aws-credentials@v4`.

Until we bump, every PR carries the deprecation warning noise and every
workflow run is on borrowed time against the September 16, 2026 cutoff.

While we are in there, the same set of files would benefit from a security
hardening pass: pin every third-party action to a commit SHA with a version
comment rather than a floating tag. GitHub's own guidance is to SHA-pin for
production workflows so a maintainer pushing malicious code under an existing
tag cannot silently affect our CI. Today some references are pinned to
`@master` (`snyk/actions/setup`, `snyk/actions/golang`, `snyk/actions/docker`),
which is the worst case.

## Scope

Workflows under `.github/workflows/` that need a major-version bump to a
Node.js 24 compatible release AND a SHA pin:

- `analysis.yml`: `actions/checkout`, `actions/setup-go`,
  `terraform-linters/setup-tflint`, `snyk/actions/setup`,
  `snyk/actions/golang`
- `backend.yml`: `actions/checkout`, `actions/setup-go`,
  `aws-actions/configure-aws-credentials`, `snyk/actions/docker`
- `infrastructure.yml`: `actions/checkout`,
  `aws-actions/configure-aws-credentials`
- `orchestration-dev.yml`, `orchestration-impl.yml`: `actions/checkout`
- `ops-image.yml`: `actions/checkout`,
  `aws-actions/configure-aws-credentials`

Pinning format to standardize on (example):

    - uses: actions/checkout@<commit-sha>  # v5.0.0

Each entry needs: bump to latest stable major running on Node.js 24, pin to
the commit SHA of that release, and a trailing comment with the human-readable
tag so reviewers can grok the version at a glance.

Also in scope:

- `actions/setup-go@v5` emits
[...]

PR #349, opened 2026-06-17
chore(ci): move backend workflow actions to Node.js 24, pin to SHAs

danielbowne:
First slice of #317, scoped to the backend workflow. GitHub is forcing Actions runners off Node.js 20, so every run currently emits the deprecation warning. This moves the backend workflow's Node-based actions to their latest Node.js 24 releases, which clears the warnings, and SHA-pins the rest as a security hardening pass while we are in the file.

## Node.js 24 bumps (the actual warning fix)

These five were running on Node 20; each is now on its latest Node 24 release:

- `actions/checkout` v4 -> v6.0.3
- `docker/setup-buildx-action` v3 -> v4.1.0
- `hoverkraft-tech/compose-action` v2.0.2 -> v3.0.0
- `aws-actions/configure-aws-credentials` v4 -> v6.2.0
- `actions/setup-go` v5 -> v6.4.0

`orchestration-dev.yml` also had `actions/checkout@v4` in the change-detection job; bumped to v6.0.3 so that job stops warning too.

## SHA pinning (security bonus)

Every third-party action in `backend.yml` is now pinned to a commit SHA with a `# vX.Y.Z` comment. This drops the two floating refs that were the worst case:

- `snyk/actions/docker@master` -> pinned SHA (no tagged releases upstream)
- `aquia-inc/emberfall@main` -> pinned SHA (composite action; the emberfall binary version still comes from the `version` input)

## Also

- `actions/setup-go` now sets `cache-dependency-path: backend/go.sum`, so the dependency cache restores instead of missing on every run (go.sum lives under `backend/`, not repo root).
- `orchestration-dev.yml` now runs the backend smoke test when `backend.yml` itself changes, not only when `backend/**` code changes, so workflow edits like this one are exercised by CI instead of shipping unvalidated. That is what lets this PR test its own bumps.
- Added `.github/dependabot.yml` with a grouped `github-actions` entry so the SHA pins receive automated update PRs and do not rot.

## Scope

Backend workflow only, as the pilot. The remaining workflows in #317 (`analysis.yml`, `infrastructure.yml`, `orchestration-impl.yml`, `ops-image.yml`) are follow-ups once this pattern is proven green.

## Test plan

- [ ] Backend smoke-test job runs on this PR (via the new workflow-change trigger) and passes
- [ ] No Node.js 20 deprecation warning in the backend job logs
- [ ] `compose-action` v3 still starts the API + Postgres e2e stack; emberfall smoke tests pass
- [ ] `configure-aws-credentials` v6 still assumes the role and the image pushes to ECR
- [ ] setup-go cache restores from `backend/go.sum` (no cache-miss warning)

Refs #317

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  # Keep the SHA-pinned GitHub Actions current. Without this, pinned digests
+  # rot and only land security/feature fixes when someone bumps them by hand.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+      day: monday
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: chore(ci)
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -91,7 +91,10 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
-          go-version: '1.24.1'
+          # Read the required Go from go.mod rather than hardcoding it. setup-go
+          # v6 sets GOTOOLCHAIN=local, so a pinned version below go.mod's
+          # requirement no longer auto-downloads and the Lambda builds fail.
+          go-version-file: backend/go.mod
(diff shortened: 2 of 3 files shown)


a-dipietro (approved):
lgtm!

PR #347, opened 2026-06-16
chore(infra): enable Entra dual-IdP in dev (entra_enabled = true)

voidspooks:
## Summary

Flip `entra_enabled = true` for the **dev** account to activate the dual-IdP (HHS Entra) infrastructure that landed dormant in #341. This is the lever for the dev Entra pilot login.

CMS/Okta login is unchanged — this only adds the second IdP path. Dev only; prod stays `false` until the dev pilot is validated.

## What flipping the flag does (on apply)

- Creates the per-IdP `/login/entra*` ALB listener rule (with its distinct session cookie name) and the forward-only rule for `/api/v1/auth/lookup`.
- Moves `/api/*` off ALB OIDC onto backend session validation.
- Injects the Entra + session-signing env into the API task (the `data` sources for `ztmf_entra_oidc` / `ztmf_session_signing_key` go live).

## Why it's ready

- Both Secrets Manager values (`ztmf_entra_oidc`, `ztmf_session_signing_key`) are **seeded and verified in dev** (client secret confirmed authenticating to the HHS tenant).
- Both dev + prod **redirect URIs are already registered** on the HHS Entra app registration.
- The backend code + ALB rules merged in #341.

## Draft — do not merge yet

Held as a draft on purpose so the DEV orchestration does **not** auto-apply (the orchestration jobs gate on `draft == false`). Merge/ready-for-review only once:

- [ ] Frontend email landing page (CMS-Enterprise/ztmf-ui#328) is ready
- [ ] HHS / test pilot users are provisioned in ZTMF (`identity_provider = entra`)
- [ ] We're ready to run the end-to-end dev pilot login (confirms the two remaining HHS-side unknowns: Conditional Access allowlist + v2 token setting, which only surface on a real login)

## Rollback

Revert to `entra_enabled = false` and re-apply; restores today's Okta-only behavior.

Refs #341, #264, CMS-Enterprise/ztmf-misc#170

PR #341, opened 2026-06-15
Add dual-IdP authentication (Entra + Okta), dormant behind entra_enabled

danielbowne:
## Summary

Adds the backend and infrastructure for dual-IdP authentication: HHS OpDiv users authenticate through Microsoft Entra ID alongside the existing CMS Okta. The entire feature ships **dormant behind the `entra_enabled` Terraform flag (default `false` in dev and prod)**, so merging and deploying this changes nothing about today's Okta-only behavior. It only takes effect after an operator loads the Entra secret and flips the flag, per the runbook below.

Implements the backend/infra slice of the dual-IdP epic (CMS-Enterprise/ztmf-misc#170). The frontend landing page is tracked separately (CMS-Enterprise/ztmf-ui#328, #329).

## What this delivers

- **Multi-issuer token validation.** The API validates tokens from both issuers: Okta (ES256) and Entra (RS256 via JWKS), with the Entra tenant pinned and a per-IdP issuer allowlist. Local dev and the E2E suite continue to use HS256.
- **Pre-auth IdP lookup.** `GET /api/v1/auth/lookup?email=` (unauthenticated, rate-limited) returns `{ "data": { "idp": "okta" | "entra" | null } }` so the browser can route each user to the correct login path. Unknown and unprovisioned emails return the same shape (no enumeration), and the email is not logged.
- **Application session layer.** After OIDC login the backend mints its own signed session cookie and gates `/api/*` by that session or an IdP bearer, with an origin (CSRF) check on state-changing requests.
- **Per-IdP ALB listener rules.** A new `/login/entra*` rule with a distinct session cookie name, a forward-only rule for the lookup endpoint, and the `/api/*` switch from ALB OIDC to backend-gated. The existing Okta `/login*` and `/api/*` rules are unchanged while the flag is off.
- **Token-validation hardening** (from a security review of this work):
  - Optional per-IdP audience pinning, enforced only when configured, so a validly signed token minted for a different application in the same issuer or tenant is rejected. The Entra audience is wired to the ZTMF app client id; the Okta audience is intentionally left unset for now since Okta traffic stays ALB-gated until cutover.
  - HS256 tokens are rejected on the IdP path outside local/test, so a deployed environment is asymmetric-only (ES256/RS256) regardless of environment configuration.

## Why this is safe to merge now

With `entra_enabled = false` (the committed value in both dev and prod):

[...]

PR #350, opened 2026-06-17
chore(infra): enable Entra dual-IdP in dev (entra_enabled = true)

danielbowne:
Flips `entra_enabled = true` for dev, activating the dormant dual-IdP infra from #341: the per-IdP ALB listener rules (`/api/v1/auth/lookup` forward, `/login/entra` Entra OIDC, `/login` Okta), `/api/*` moved off ALB OIDC to backend session validation, and the Entra + session env injected into the API task. Okta login is unchanged.

This is the in-repo version of #347 by @voidspooks, moved onto a CMS-Enterprise branch so the infrastructure deploy (terraform apply, which needs the org OIDC role) runs; that does not run on a fork PR. The single commit and authorship are unchanged.

## Why now

The frontend landing page (ztmf-ui#404) is deployed to dev and calls `GET /api/v1/auth/lookup` on submit, but that ALB rule only exists when `entra_enabled = true`. With the flag still false, the lookup falls through to the OIDC-gated `/api/*` rule, fails for a logged-out visitor, and the page shows "We can't determine an identity provider" for everyone, including Okta users. This flip routes the lookup and unblocks dev testing.

## Scope

- dev only. `infrastructure/tfvars/dev.tfvars`, one line plus a comment. prod (`prod.tfvars`) stays `false`.
- Preconditions met: `ztmf_entra_oidc` and `ztmf_session_signing_key` are seeded and verified in the dev account.

## Deploy behavior

Once this is non-draft, orchestration-dev applies it to dev. It only persists across later dev applies once merged to main, so merge after the dev login is validated end to end.

Refs #347

### DIFF
--- a/infrastructure/tfvars/dev.tfvars
+++ b/infrastructure/tfvars/dev.tfvars
@@ -4,11 +4,12 @@ domain_name_prefix     = "dev."
 ecs_service_task_count = 1
 # job_code = "ZTMF_SCORING_USER"
 
-# Entra dual-IdP. Keep false until both secrets are seeded in the dev
-# account (scripts/bootstrap-entra-secrets.sh), then flip to true to add the
-# per-IdP ALB rules, move /api/* off ALB OIDC to backend session validation,
-# and inject the Entra + session env into the API task.
-entra_enabled = false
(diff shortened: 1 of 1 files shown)


MackOverflow (approved):
Looks good, have you confirmed that unauthenticated /api/v1/* gets a 401?

Issue #264, opened 2026-02-18
Infrastructure: Per-IdP ALB listener rules


danielbowne:
# Infrastructure: Per-IdP ALB listener rules

**Epic:** CMS-Enterprise/ztmf-misc#170 (Multi-IdP Authentication)
**Sibling:** CMS-Enterprise/ztmf#265 (Backend multi-issuer token validation)

## Summary

Add new ALB listener rules for the HHS Entra OIDC flow alongside the existing Okta rule. The existing Okta `/login*` rule, app registration, redirect URI, and default session cookie name all stay unchanged. Only the new Entra rule sets a distinct `SessionCookieName` to avoid collision with the Okta default. Resolve the `/api/*` route auth strategy by flipping it from Okta `authenticate-oidc` to `allow` and moving token validation into the backend (Option A, tracked under CMS-Enterprise/ztmf#265).

## Current State

`alb-internal.tf` rules:
- Priority 1: `/login*` → `authenticate-oidc` (Okta) → redirect `/`
- Priority 2: `/api/*` → `authenticate-oidc` (Okta) → forward to API

`local.oidc_options` is a single JSON secret decoded in `locals.tf`.

## Target State

Final ALB listener rule order (agreed in the multi-IdP epic, 2026-05-12):

| Priority | Path | Action | Session cookie | Status |
| --- | --- | --- | --- | --- |
| 1 | `/api/v1/auth/lookup*` | forward, no OIDC | n/a | new |
| 2 | `/login/entra*` | `authenticate-oidc` (Entra) | `AWSELBAuthSessionCookie-Entra` | new |
| 3 | `/login*` | `authenticate-oidc` (Okta), default session cookie | `AWSELBAuthSessionCookie` (default) | unchanged |
| 4 | `/api/*` | `allow` (backend gates token, see ztmf#265) | n/a | flip from Okta-OIDC to allow |

### Secrets Manager

- [ ] Add a new secret for Entra OIDC configuration. JSON shape mirrors the existing Okta secret: `client_id`, `client_secret`, `issuer`, `authorization_endpoint`, `token_endpoint`, `user_info_endpoint`. HHS-provided tenant-specific values are documented in CMS-Enterprise/ztmf-misc#170 (internal repo).
- [ ] Add `data.aws_secretsmanager_secret_version` for the Entra OIDC secret
- [ ] Add `local.oidc_options_entra` in `locals.tf` (mirrors the existing Okta shape)

### ALB Listener Rules

[...]

Issue #166, opened 2025-01-14
refactor cloudfront to point to ALB using new VPC origins


talentedmrjones:
Currently the ALB is public and access is limited by using secret headers from CloudFront. Now that VPC origins is an option, replace the ALB with an internal one and point CloudFront API origin to that

PR #351, opened 2026-06-17
fix(infra): route /login/* through CloudFront to the ALB for Entra login

danielbowne:
Fixes the Entra login path being unreachable from the public CloudFront URL.

## Problem

CloudFront routes only the exact path `/login` to the ALB origin (`cloudfront.tf`); there is no `/login/*` behavior. So `/login/entra` does not match, falls through to the S3 default origin, and S3 returns an `AccessDenied` XML page. Entra login is unreachable for everyone, not specific to any user or email. Okta is unaffected because it lives at exactly `/login`.

The dual-IdP work added the `/login/entra*` rule on the ALB but never added the matching CloudFront behavior. The two layers are owned separately (the CloudFront login behavior was set up under #166 as an exact match), and the end-to-end "user lands at /login/entra" check was deferred to runtime, so the gap only surfaced once dev was flipped to `entra_enabled = true` and Entra login was first exercised.

## Fix

Add a `/login/*` ordered cache behavior pointing at the `ztmf_api` (ALB) origin, mirroring the existing `/login` behavior (same forwarding, no caching). It is gated with `dynamic ... for_each = var.entra_enabled ? [1] : []` so it exists only when Entra is enabled, matching the per-IdP ALB listener rules that are created under the same flag. The exact `/login` behavior is left untouched, so Okta keeps working exactly as before.

On dev (`entra_enabled = true`) this creates the behavior and `/login/entra` reaches the ALB. On prod (`entra_enabled = false`) nothing changes.

## Validation

- `terraform fmt` clean, `terraform validate` passes, `tflint` clean.
- After this applies to dev: `/login/entra` routes to the ALB (Entra OIDC) instead of returning S3 `AccessDenied`; `/login` (Okta) unchanged.

Refs #264

### DIFF
--- a/infrastructure/cloudfront.tf
+++ b/infrastructure/cloudfront.tf
@@ -188,6 +188,37 @@ resource "aws_cloudfront_distribution" "ztmf" {
     viewer_protocol_policy = "redirect-to-https"
   }
 
+  # Dual-IdP login subpaths (e.g. /login/entra) must reach the ALB origin. The
+  # exact "/login" behavior above does not match subpaths, so without this
+  # CloudFront serves the S3 default origin (AccessDenied) for /login/entra and
+  # Entra login is unreachable. Gated on entra_enabled to match the per-IdP ALB
+  # rules, which are created under the same flag (the /login/entra* listener
+  # rule). Okta keeps using the exact "/login" behavior above, unaffected.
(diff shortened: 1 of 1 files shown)


a-dipietro (approved):
Verify listener rules are UTD, otherwise g2g!

Issue #150, opened 2024-11-15
Epic: Multi-IdP Authentication with Domain-Based IdP Routing


talentedmrjones:
# Epic: Multi-IdP Authentication with Domain-Based IdP Routing

## Problem Statement

ZTMF currently authenticates all users through a single OIDC identity provider (Okta/IDM) via the AWS ALB `authenticate-oidc` action. A new requirement has emerged: users from another organization that authenticates via **Microsoft Entra ID** need access to ZTMF. The current architecture cannot route users to different identity providers — every request hits the same Okta OIDC config, and there is no mechanism for users to choose or be routed to an alternative IdP.

This issue replaces the original one-line description of #150 with a comprehensive design for multi-IdP support.

## Current Architecture & Single-IdP Assumptions

12 components are hardcoded to a single OIDC provider:

| # | Component | File | Assumption |
|---|-----------|------|------------|
| 1 | ALB listener rules | `infrastructure/alb-internal.tf` | Both rules use same `local.oidc_options` |
| 2 | OIDC secret | `infrastructure/locals.tf` | Single secret decoded into `oidc_options` |
| 3 | Verified Access | `infrastructure/verified-access.tf` | Hardcoded to `ztmf_idmokta`, `*@cms.hhs.gov` policy |
| 4 | Backend auth config | `backend/internal/config/config.go` | One `TokenKeyUrl`, one `HeaderField` |
| 5 | Token validation | `backend/cmd/api/internal/auth/token.go` | `cfg.Auth.TokenKeyUrl + kid` — single key endpoint |
| 6 | Auth middleware | `backend/cmd/api/internal/auth/middleware.go` | Single header, no issuer validation |
| 7 | User model | `backend/internal/model/users.go` | No `identity_provider` field |
| 8 | Login page | ztmf-ui `LoginPage.tsx` | Single "Login" button → `/login` |
| 9 | Auth gating | ztmf-ui (15+ components) | Duplicated 401 handling, no IdP context |
| 10 | `VITE_IDP_ENABLED` | ztmf-ui env config | Only toggles dev/prod, no IdP selection |
| 11 | `on_unauthenticated_request` | `alb-internal.tf` | `/login*` auto-redirects to single Okta |
| 12 | Session management | ALB cookies | `AWSELBAuthSessionCookie` tied to single provider |

## Desired User Experience

```
User visits ZTMF
       │
       ▼
┌─────────────────────────┐
│    ZTMF Landing Page    │
│                         │
[...]

PR #352, opened 2026-06-17
fix(users): let HHS-wide admins set identity_provider on create

danielbowne:
Backend half of letting an admin choose a new user's identity provider. Opened as a draft so it does not deploy to dev until the frontend add-user form is ready; on its own this change is inert because nothing sends `identity_provider` on create yet.

## Change

ZTMF is CMS-origin, so Okta stays the create default and Entra is the exception for HHS users. The gate that lets an actor set an explicit `identity_provider` on create was OWNER-only, so when an HHS_ADMIN chose Entra for an HHS user the value was blanked and the user defaulted to Okta - then the pre-auth lookup routed them to Okta until an OpDiv grant was toggled to re-derive it.

Widen the override gate from `IsOwner()` to `HasUnscopedRead()` (OWNER, HHS_ADMIN, HHS_READONLY_ADMIN). OpDiv-scoped admins still cannot set it, so their CMS users stay on Okta. `HHS_READONLY_ADMIN` is already write-blocked on `/users` by `IsAdmin()`, so in practice only `HHS_ADMIN` newly gains the override. The Okta create default and the OpDiv-grant derivation are unchanged.

## Depends on

Frontend work to send `identity_provider` (and OpDiv assignment) from the add-user form, shown only to HHS-wide admins: ztmf-ui#406 (CMS-Enterprise/ztmf-ui#406). This PR stays a draft until that lands so they deploy together.

## Test plan

- [x] `go build ./...`
- [x] `make test-e2e` passes 117/0, including an assertion that a created user with no explicit IdP defaults to `okta`
- [x] Independent review (gate widening confirmed safe; HHS_READONLY_ADMIN cannot reach the write path)

Refs #150

### DIFF
--- a/backend/cmd/api/internal/controller/users.go
+++ b/backend/cmd/api/internal/controller/users.go
@@ -104,10 +104,13 @@ func SaveUser(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// identity_provider is derived from OpDiv on grant and is only overridable
-	// by an OWNER. Ignore any client-supplied value from a non-OWNER so it
-	// cannot be used to misroute a user's login.
-	if !authdUser.IsOwner() {
+	// identity_provider is derived from OpDiv membership (see deriveIdentityProvider).
+	// An explicit override is only honored from an HHS-wide actor (OWNER, HHS_ADMIN,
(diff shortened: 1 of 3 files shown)


jsos3-cms (approved):
LGTM!

PR #354, opened 2026-06-18
fix(auth): validate Okta JWT kid and harden key-fetch client

danielbowne:
Hardens JWT verification key fetching against an unsanitized-input flaw in the Okta path.

## Problem

`oktaKey()` took the `kid` from the (necessarily unverified) token header and concatenated it directly into the key-fetch URL, then fetched it with `http.DefaultClient` (no timeout, follows redirects) and used the returned PEM as the verification key. Because `kid` selects the key, it is read before the signature is checked - so it is untrusted input being used to shape an outbound request and choose a trust anchor. The Entra (RS256) path was already safe (fixed JWKS URL, `kid` as a map key); this brings the Okta path to the same standard.

## Change

- Validate `kid` against `^[A-Za-z0-9_-]{1,200}$` before any cache, config, or network access; reject otherwise. The allowlist (URL-safe characters only, matching real Okta/ALB key ids) is what makes the existing `TokenKeyUrl + kid` concatenation safe - no path traversal, host/userinfo injection, scheme, or encoded tricks can reshape the request.
- Route both the Okta per-`kid` fetch and the Entra JWKS fetch through a dedicated `http.Client` with a 10s timeout that refuses to follow redirects, so a slow/hung endpoint cannot pin a request goroutine and a 3xx cannot bounce the fetch to an unexpected host.
- Capture the previously-ignored `http.NewRequest` error.

Behavior is unchanged for legitimate tokens: a valid `kid` still fetches exactly `TokenKeyUrl + kid` and parses the ECDSA key.

## Test plan

- [x] `make test-full` (unit + Emberfall E2E) passes 117/0
- [x] New unit tests: `kid` allowlist (valid vs traversal/host-injection/scheme/oversized), rejection before any outbound fetch, and a positive path (valid `kid` -> correct URL -> ECDSA parse)
- [x] `go build`, `go vet`, `gofmt` clean
- [x] Independent review

Refs CMS-Enterprise/ztmf-misc#240

### DIFF
--- a/backend/cmd/api/internal/auth/token.go
+++ b/backend/cmd/api/internal/auth/token.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"math/big"
 	"net/http"
+	"regexp"
 	"slices"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


MackOverflow (dismissed):
This is a well-scoped, correctly implemented security fix with thorough tests and clear documentation. The vulnerability (unsanitized kid in URL construction + no-timeout HTTP client) is real and the remediation is appropriate. Nice catch Cam. 

Two nits: 
TestOktaKeyValidKidFetchesAndParses could additionally assert that gotPath == "/"+kid when the base URL has no trailing slash (edge case: double-slash if TokenKeyUrl already ends in /). Currently the test sets ts.URL + "/" so it only tests the slash-appended case.
Consider a test case for a kid of exactly 200 chars (boundary) and exactly 201 chars — the current invalid list uses 201 and 500, but 200 (valid boundary) isn't in the valid list.

PR #355, opened 2026-06-18
feat(#133): add scores diff endpoint to compare two data calls

### LINKED ISSUE #133: Method to show all the changes in a time period

voidspooks:
Closes #133.

## What

Adds `GET /api/v1/scores/diff?from={datacallid}&to={datacallid}[&fismasystemid={id}]` — compares the score (functionoption) answers of two data calls and returns **only the questionnaire functions whose answer changed**, each annotated with **who made the later change and when**.

This is the "show all the changes in a time period" ask in #133: the two data calls bound the period; the response is the questionnaire delta plus attribution.

## Response shape

One entry per changed function:

```jsonc
{
  "fismasystemid": 1001,
  "functionid": 7007,
  "function": "Battle Station Device Management",
  "question": "Does your system track all assets with comprehensive inventory?",
  "from": { "scoreid": 9001, "functionoptionid": 25, "optionname": "Defined",  "score": 2, "notes": "..." },
  "to":   { "scoreid": 9025, "functionoptionid": 26, "optionname": "Managed",  "score": 3, "notes": "..." },
  "changed_at": "2026-06-18T16:20:18Z",
  "changed_by": { "userid": "...", "name": "Grand Moff Tarkin", "email": "...", "role": "OWNER" }
}
```

- `from` / `to` are the answer in each cycle; either is `null` when the function was answered in only one cycle (added or removed).
- `changed_by` / `changed_at` attribute the write that produced the `to` answer, resolved from the `events` audit trail (mirrors the `last_edited_by` lateral in `FindScores`). Absent when no write event exists for the `to` score (seed data, or a function answered only in the `from` cycle).

## What counts as a change

A function differs when the selected functionoption changed, the notes changed, or it was answered in exactly one of the two cycles. `nil` and empty notes compare equal (same normalization as the no-op rule on the write path). Unchanged rows are dropped.

## Scope & validation

- Tier scoping matches `ListScores`, applied after query decode so a client cannot widen it: unscoped admins see all systems, OpDiv-scoped admins their OpDivs' systems (fail-closed), ISSO/ISSM their assigned systems.
- `from` and `to` are required and must differ → `400` otherwise.

## Implementation

- New `FindScoreDiff` / `buildScoreDiffSQL` in `internal/model/scorediff.go`. Built as parameterized raw SQL via the existing `rawQuery` read-path helper (FULL OUTER JOIN of the two cycle CTEs on `(system, function)` + catalog joins for labels + events lateral for attribution exceed squirrel's ergonomics, matching the precedent set by `buildPillarScoresSQL`). SELECT-only, so it never touches the event-recording write path.
- New `GetScoresDiff` controller handler + `/api/v1/scores/diff` route.
- Regenerated `openapi.yaml` (swag); redocly lint clean (pre-existing warnings only).

## Testing

- Unit: input validation; SQL builder filter/scope shaping and arg ordering (single-system, ISSO user scope, OpDiv grants + fail-closed).
- Integration (real Postgres, empire seed): changed answer, unchanged-dropped, one-sided via the outer join, and `to`-side attribution.
- Verified locally on a freshly wiped + reseeded dev DB: `go build/vet/test ./...` all green (full suite incl. integration), and live HTTP smoke tests through the running API (400 validation cases, unchanged→empty, a rich non-empty diff, and an edit-then-diff that surfaced the change attributed to the editor).

> Note: as a fork PR without repo secrets, Snyk / secret-dependent CI checks won't run and will show red — expected; maintainers re-run after pulling.

### DIFF
--- a/backend/cmd/api/internal/controller/scores.go
+++ b/backend/cmd/api/internal/controller/scores.go
@@ -110,6 +110,50 @@ func SaveScore(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, score, err)
 }
 
+//	@Summary	Diff scores between two data calls
+//	@Description	Compares the score (functionoption) answers of two data calls and returns only the questionnaire functions whose answer changed, each annotated with who made the later change and when. Scoped to the caller's tier: unscoped admins see all systems, OpDiv-scoped admins their OpDivs' systems, and ISSO/ISSM their assigned systems.
+//	@Tags		scores
+//	@Produce	json
+//	@Security	bearerAuth
+//	@Param		from			query		int	true	"Data call ID to compare from (earlier cycle)"
+//	@Param		to				query		int	true	"Data call ID to compare to (later cycle)"
+//	@Param		fismasystemid	query		int	false	"Limit the diff to a single FISMA system"
+//	@Success	200				{object}	apiResponse[[]model.ScoreDiff]
+//	@Failure	400				{object}	apiResponse[any]
+//	@Failure	500				{object}	apiResponse[any]
+//	@Router		/scores/diff [get]
+func GetScoresDiff(w http.ResponseWriter, r *http.Request) {
+	var (
+		diffs []*model.ScoreDiff
+		err   error
+	)
+
+	user := model.UserFromContext(r.Context())
+	input := model.FindScoreDiffInput{}
+
+	err = decoder.Decode(&input, r.URL.Query())
+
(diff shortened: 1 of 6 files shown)


voidspooks:
I think this will need follow up UI ticket - taking a look. I will put one up if that's the case.

MackOverflow:
Rehomed to #356 so the Snyk/deploy CI gates run (they cannot on a fork PR). Commits pushed unchanged, authorship preserved. Closing in favor of #356. Thanks @voidspooks.

PR #369, opened 2026-06-26
create batch grant opdivs endpoint

### LINKED ISSUE #345: Backend: batch endpoint for setting a user's OpDiv grants
### Feature Type
General Feature

### Workflow Description

The OpDivGrantModal grants and revokes a user's OpDiv memberships by making one API call per OpDiv (a per-item write loop). This is the write-side counterpart to the users-list read N+1 that was already fixed. Severity is lower (user-triggered, small N, now bounded by the connection pool), so it was deferred from that work and is tracked here.

Add a batch endpoint that sets a user's full OpDiv grant set in a single request. The backend reconciles the desired set against current grants (adds missing, removes extra) in one transaction, and re-derives `identity_provider` once at the end rather than per grant.

### Acceptance Criteria

- [ ] A single endpoint accepts the desired set of `opdiv_id`s for a user (e.g. `PUT /api/v1/users/{userid}/opdivs` with the full set)
- [ ] Backend adds/removes grants to match the set in one transaction
- [ ] `identity_provider` is re-derived once after the reconcile, not per grant
- [ ] Authorization unchanged: admin tier + OpDiv scope still enforced (an OPDIV_ADMIN can only grant OpDivs they hold)
- [ ] OpenAPI (swag annotations) and Emberfall tests updated
- [ ] Frontend (paired issue) switches OpDivGrantModal to this endpoint

### Additional Context

Pairs with the frontend issue: CMS-Enterprise/ztmf-ui#395
Deferred from the users-list read N+1 fix (backend PR CMS-Enterprise/ztmf#344, frontend pair). Relates to the multi-OpDiv work.

costacalvin:
## Summary

Closes #345. Paired with ztmf-ui#437 (frontend switches `OpDivGrantModal` to use this endpoint).

Adds `PUT /api/v1/users/{userid}/opdivs` — a batch endpoint that sets a user's full OpDiv grant set in one transaction, replacing the N per-item POST/DELETE calls previously required by the frontend.

- **Controller** (`controller/usersopdivs.go`): adds `SetUserOpDivs` with a two-stage authorization check — scope gate first (pure memory, short-circuits before any DB call), then tier-ceiling check via `FindUserByID`. Reconciles desired vs current grants into `toAdd`/`toRemove` slices; OPDIV_ADMIN's `toRemove` is further filtered to only grants within their scope so out-of-scope grants are left untouched.
- **Model** (`internal/model/usersopdivs.go`): adds `SetUserOpDivs` which executes DELETE then INSERT in a single pgx transaction on a dedicated pinned connection (`db.Conn`), then calls `deriveIdentityProvider` once after commit. `ON CONFLICT DO NOTHING` on INSERT makes the add step idempotent.
- **Router** (`router/router.go`): registers `PUT /api/v1/users/{userid}/opdivs` → `SetUserOpDivs`.
- **Tests** (`controller/usersopdivs_test.go`): adds four cases — non-admin tier forbidden, OPDIV_ADMIN out-of-scope forbidden, malformed body returns 400, OPDIV_ADMIN in-scope passes the gate (500 from no test DB, not 403).

## Acceptance criteria

- [x] `PUT /api/v1/users/{userid}/opdivs` accepts `{ "opdiv_ids": [1, 2, 3] }` and reconciles grants in one transaction
- [x] Returns 204 on success
- [x] Non-admin roles receive 403
- [x] OPDIV_ADMIN receives 403 if any requested ID is outside their own grants
- [x] OPDIV_ADMIN cannot remove grants outside their scope (left unchanged)
- [x] `identity_provider` is re-derived exactly once after the transaction commits
- [x] Sending an empty array clears all in-scope grants

## Test plan

- [ ] `PUT /opdivs` with a valid admin token and a subset of OpDiv IDs → 204, grants updated
- [ ] Same request as OPDIV_ADMIN with only their own OpDiv IDs → 204
- [ ] Same request as OPDIV_ADMIN with an OpDiv they don't hold → 403
- [ ] Non-admin token → 403
- [ ] Malformed body → 400
- [ ] Send empty `opdiv_ids: []` → 204, all in-scope grants removed
- [ ] Verify `identity_provider` on the user reflects the new grant set after the call

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/cmd/api/internal/controller/usersopdivs.go
+++ b/backend/cmd/api/internal/controller/usersopdivs.go
@@ -128,6 +128,102 @@ func CreateUserOpDiv(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, saved, err)
 }
 
+type setUserOpDivsInput struct {
+	OpDivIDs []int32 `json:"opdiv_ids"`
+}
+
+// SetUserOpDivs replaces a user's full OpDiv grant set in one batch. The desired
+// set is reconciled against current grants (adds missing, removes extra) in one
+// transaction, and identity_provider is re-derived once at the end. An
+// OPDIV_ADMIN may only include OpDivs they hold; grants outside their scope are
+// left unchanged.
+//
+//	@Summary	Set a user's OpDiv grants (batch replace)
+//	@Tags		users
+//	@Accept		json
+//	@Produce	json
+//	@Security	bearerAuth
+//	@Param		userid	path	string				true	"User ID"
+//	@Param		body	body	setUserOpDivsInput	true	"Desired grant set"
+//	@Success	204	"No Content"
+//	@Failure	400	{object}	apiResponse[any]
+//	@Failure	403	{object}	apiResponse[any]
+//	@Failure	404	{object}	apiResponse[any]
+//	@Failure	500	{object}	apiResponse[any]
+//	@Router		/users/{userid}/opdivs [put]
+func SetUserOpDivs(w http.ResponseWriter, r *http.Request) {
+	authdUser := model.UserFromContext(r.Context())
+	if !authdUser.IsAdmin() {
+		respond(w, r, nil, ErrForbidden)
+		return
+	}
+
+	userID, ok := mux.Vars(r)["userid"]
+	if !ok {
+		respond(w, r, nil, ErrNotFound)
+		return
+	}
+
+	var input setUserOpDivsInput
+	if err := getJSON(r.Body, &input); err != nil {
+		log.Println(err)
+		respond(w, r, nil, ErrMalformed)
+		return
+	}
+
+	// Scope gate: OPDIV_ADMIN may only request OpDivs they hold. Pure memory
+	// check — runs before the tier-ceiling DB call to short-circuit early.
+	if !authdUser.HasUnscopedRead() {
+		for _, id := range input.OpDivIDs {
+			if !authdUser.IsAssignedOpDiv(id) {
+				respond(w, r, nil, ErrForbidden)
+				return
+			}
+		}
+	}
+
+	// Tier ceiling: cannot manage a higher-tier user.
+	target, err := model.FindUserByID(r.Context(), userID)
+	if err != nil {
+		respond(w, r, nil, err)
+		return
+	}
+	if !authdUser.CanAssignRole(target.Role) {
+		respond(w, r, nil, ErrForbidden)
+		return
+	}
+
+	current, err := model.FindUserOpDivsByUserID(r.Context(), userID)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


jsos3-cms:
Thanks for this, @costacalvin. I've moved the commit onto an in-repo branch as #370 so the Snyk and Infrastructure/Deploy gates can run (they can't on a fork PR) — your commit is pushed unchanged, authorship preserved. I've set #370 to draft and **blocked** while we work through a few authorization findings from review. Leaving this PR open as the place to discuss — respond here or push fixes and we'll carry them onto the rehome branch.

Blocking items before it can land:

1. **Empty vs missing `opdiv_ids` can silently clear all grants.** A missing key and an explicit `[]` both deserialize to an empty slice and fall through to the "remove all in-scope grants" path, so a frontend/serialization bug that drops the field would silently revoke a user's grants. Make the field a pointer (`*[]int32`) and return 400 on nil, so "clear all" has to be an explicit `[]`.

2. **`toAdd` isn't scope-guarded locally.** `toRemove` is re-filtered to the caller's scope, but `toAdd` is built straight from the request IDs and relies on the earlier scope gate. Correct today, but fragile if that gate is ever loosened — mirror the scope check on `toAdd` for defense in depth.

Security item to fix alongside:

3. **No shared-OpDiv check on the target user.** After `FindUserByID` the tier ceiling is checked, but not whether the caller shares an OpDiv with the target. An OPDIV_ADMIN holding OpDiv 5 could `PUT /users/{any-id}/opdivs` with `[5]` and attach their OpDiv to a user they share nothing with — reaching outside their management boundary. Add the shared-OpDiv check after resolving the target (the single-grant handlers have the same gap, worth fixing together).

Smaller notes: `deriveIdentityProvider` failing after commit returns a 500 on an already-committed write; the route is `/opdivs` while the existing family is `/assignedopdivs`; and a couple of tests for the explicit `[]` clear path and an in-scope reconciliation would help once the nil-guard is in.

This is paired with the UI change (ztmf-ui#438), so the two need to land together — coordination note is on #370.

costacalvin:
@jsos3-cms 
Addressed the blocking items, as well as the shared opdiv check. 

Added some tests. 

Kept route as `/opdivs` which is what was listed in the original issue, as we would need to update the ztmf-ui PR and rehome as well. Could definitely address in a quick follow up though. The deriveIdentityProvider returning 500 is how it is handled for the previous OpDivGrant methods so I believe some more convo could be had there about how we want to handle it if we do decide to change it (rollback tx if derive fails or logging failure of deriveIdentityProvider and returning status code of the granted opdivs)

Tested locally + tests passing, good for re-review!

jsos3-cms:
Closed by #370

PR #370, opened 2026-06-27
create batch grant opdivs endpoint

### LINKED ISSUE #345: Backend: batch endpoint for setting a user's OpDiv grants

jsos3-cms:
In-repo version of #369 by @costacalvin, moved onto a CMS-Enterprise branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) can run — they cannot on a fork PR. Commits are pushed unchanged; authorship is preserved.

Refs #369

## Summary

Closes #345. Paired with ztmf-ui#437 (frontend switches `OpDivGrantModal` to use this endpoint).

Adds `PUT /api/v1/users/{userid}/opdivs` — a batch endpoint that sets a user's full OpDiv grant set in one transaction, replacing the N per-item POST/DELETE calls previously required by the frontend.

- **Controller** (`controller/usersopdivs.go`): adds `SetUserOpDivs` with a two-stage authorization check — scope gate first (pure memory, short-circuits before any DB call), then tier-ceiling check via `FindUserByID`. Reconciles desired vs current grants into `toAdd`/`toRemove` slices; OPDIV_ADMIN's `toRemove` is further filtered to only grants within their scope so out-of-scope grants are left untouched.
- **Model** (`internal/model/usersopdivs.go`): adds `SetUserOpDivs` which executes DELETE then INSERT in a single pgx transaction on a dedicated pinned connection (`db.Conn`), then calls `deriveIdentityProvider` once after commit. `ON CONFLICT DO NOTHING` on INSERT makes the add step idempotent.
- **Router** (`router/router.go`): registers `PUT /api/v1/users/{userid}/opdivs` → `SetUserOpDivs`.
- **Tests** (`controller/usersopdivs_test.go`): adds four cases — non-admin tier forbidden, OPDIV_ADMIN out-of-scope forbidden, malformed body returns 400, OPDIV_ADMIN in-scope passes the gate (500 from no test DB, not 403).

## Acceptance criteria

- [x] `PUT /api/v1/users/{userid}/opdivs` accepts `{ "opdiv_ids": [1, 2, 3] }` and reconciles grants in one transaction
- [x] Returns 204 on success
- [x] Non-admin roles receive 403
- [x] OPDIV_ADMIN receives 403 if any requested ID is outside their own grants
- [x] OPDIV_ADMIN cannot remove grants outside their scope (left unchanged)
- [x] `identity_provider` is re-derived exactly once after the transaction commits
- [x] Sending an empty array clears all in-scope grants

## Test plan

- [ ] `PUT /opdivs` with a valid admin token and a subset of OpDiv IDs → 204, grants updated
- [ ] Same request as OPDIV_ADMIN with only their own OpDiv IDs → 204
- [ ] Same request as OPDIV_ADMIN with an OpDiv they don't hold → 403
- [ ] Non-admin token → 403
- [ ] Malformed body → 400
- [ ] Send empty `opdiv_ids: []` → 204, all in-scope grants removed
- [ ] Verify `identity_provider` on the user reflects the new grant set after the call

Closes #345

### DIFF
--- a/backend/cmd/api/internal/controller/usersopdivs.go
+++ b/backend/cmd/api/internal/controller/usersopdivs.go
@@ -128,6 +128,102 @@ func CreateUserOpDiv(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, saved, err)
 }
 
+type setUserOpDivsInput struct {
+	OpDivIDs []int32 `json:"opdiv_ids"`
+}
+
+// SetUserOpDivs replaces a user's full OpDiv grant set in one batch. The desired
+// set is reconciled against current grants (adds missing, removes extra) in one
+// transaction, and identity_provider is re-derived once at the end. An
+// OPDIV_ADMIN may only include OpDivs they hold; grants outside their scope are
+// left unchanged.
+//
+//	@Summary	Set a user's OpDiv grants (batch replace)
+//	@Tags		users
+//	@Accept		json
+//	@Produce	json
+//	@Security	bearerAuth
+//	@Param		userid	path	string				true	"User ID"
+//	@Param		body	body	setUserOpDivsInput	true	"Desired grant set"
+//	@Success	204	"No Content"
+//	@Failure	400	{object}	apiResponse[any]
+//	@Failure	403	{object}	apiResponse[any]
+//	@Failure	404	{object}	apiResponse[any]
+//	@Failure	500	{object}	apiResponse[any]
+//	@Router		/users/{userid}/opdivs [put]
+func SetUserOpDivs(w http.ResponseWriter, r *http.Request) {
+	authdUser := model.UserFromContext(r.Context())
+	if !authdUser.IsAdmin() {
+		respond(w, r, nil, ErrForbidden)
+		return
+	}
+
+	userID, ok := mux.Vars(r)["userid"]
+	if !ok {
+		respond(w, r, nil, ErrNotFound)
+		return
+	}
+
+	var input setUserOpDivsInput
+	if err := getJSON(r.Body, &input); err != nil {
+		log.Println(err)
+		respond(w, r, nil, ErrMalformed)
+		return
+	}
+
+	// Scope gate: OPDIV_ADMIN may only request OpDivs they hold. Pure memory
+	// check — runs before the tier-ceiling DB call to short-circuit early.
+	if !authdUser.HasUnscopedRead() {
+		for _, id := range input.OpDivIDs {
+			if !authdUser.IsAssignedOpDiv(id) {
+				respond(w, r, nil, ErrForbidden)
+				return
+			}
+		}
+	}
+
+	// Tier ceiling: cannot manage a higher-tier user.
+	target, err := model.FindUserByID(r.Context(), userID)
+	if err != nil {
+		respond(w, r, nil, err)
+		return
+	}
+	if !authdUser.CanAssignRole(target.Role) {
+		respond(w, r, nil, ErrForbidden)
+		return
+	}
+
+	current, err := model.FindUserOpDivsByUserID(r.Context(), userID)
+	if err != nil {
+		respond(w, r, nil, err)
+		return
+	}
+
+	desiredSet := make(map[int32]bool, len(input.OpDivIDs))
+	for _, id := range input.OpDivIDs {
+		desiredSet[id] = true
+	}
+	currentSet := make(map[int32]bool, len(current))
+	for _, id := range current {
+		currentSet[id] = true
+	}
+
+	var toAdd, toRemove []int32
+	for _, id := range input.OpDivIDs {
+		if !currentSet[id] {
+			toAdd = append(toAdd, id)
+		}
+	}
+	for _, id := range current {
+		if !desiredSet[id] && (authdUser.HasUnscopedRead() || authdUser.IsAssignedOpDiv(id)) {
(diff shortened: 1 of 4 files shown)


jsos3-cms:
**Cross-repo coordination — batch OpDiv grants (backend + UI)**

This change spans two PRs that must land together:

- Backend: CMS-Enterprise/ztmf#370 (rehome of #369) — adds `PUT /api/v1/users/{userid}/opdivs`
- Frontend: CMS-Enterprise/ztmf-ui#438 (rehome of #437) — switches `OpDivGrantModal` to the batch endpoint

Both are currently **draft / blocked** pending authorization fixes on the backend (detail on ztmf#369). Two things to settle across both before merge:

1. **Backend authz fixes land first** — the UI depends on the endpoint's contract (403 behavior, empty-array semantics), so those need to be settled before the UI is finalized.
2. **Route naming** — the backend currently exposes `/opdivs` while the existing family is `/assignedopdivs`. Agree the final path here so the UI calls the right one.

Holding both until the backend findings are resolved.

jsos3-cms:
**Status — rebased onto main**

Rebased onto the current `main`, which now includes #352 ("let HHS-wide admins set identity_provider on create"). The batch-endpoint commit replays linearly on top — single commit, no merge commit, no conflicts (#352 touched the user controller/model; this PR touches the OpDiv handlers, so no overlap). The branch is mergeable.

Still in draft and **blocked** pending the authorization fixes from review before it can land:

1. **Empty vs missing `opdiv_ids` can silently clear all grants** — a missing key and an explicit `[]` both deserialize to an empty slice and fall through to the "remove all in-scope grants" path. Make the field a pointer (`*[]int32`) and return 400 on nil, so "clear all" has to be an explicit `[]`.
2. **`toAdd` isn't scope-guarded locally** — `toRemove` is re-filtered to the caller's scope, but `toAdd` is built from the request IDs and relies on the earlier scope gate. Mirror the scope check on `toAdd` for defense in depth.
3. **No shared-OpDiv check on the target user** — after `FindUserByID` the tier ceiling is checked, but not whether the caller shares an OpDiv with the target. An OPDIV_ADMIN could attach their OpDiv to a user they share nothing with. Add the shared-OpDiv check after resolving the target (the single-grant handlers have the same gap, worth fixing together).

Detail discussed on #369.

This also gates the frontend: ztmf-ui#438 (now rebased batch-only and mergeable) needs this endpoint merged and deployed to dev before its modal Save works. Holding until the fixes land.

jsos3-cms:
**Updated with the authorization fixes — un-drafting for review**

Brought @costacalvin's fixes in-repo (his fork was rebased onto current `main`, so this is current). All three blocking items from the earlier review are addressed and verified:

1. **Empty vs missing `opdiv_ids`** — the field is now `*[]int32` with a nil check returning 400, so a missing or absent body can no longer silently clear grants; an explicit `[]` still clears in-scope grants.
2. **`toAdd` scope guard** — both the add and remove sets are now scope-filtered (`HasUnscopedRead() || IsAssignedOpDiv(id)`), closing the asymmetry.
3. **Shared-OpDiv check on the target** — after `FindUserByID`, a scoped admin acting on a user they share no OpDiv with is now rejected (mirrors `CreateUserOpDiv`); HHS-wide admins and the new-user bootstrap path are preserved.

Tests added for the reachable authorization paths (nil body → 400, explicit empty array, malformed → 400, out-of-scope rejected, non-admin → 403).

Non-blocking follow-ups:
- A controller-level test for the shared-OpDiv denial path specifically — the new-user exemption is tested, but not the "scoped admin blocked from a non-shared user" case.
- Route kept as `/opdivs` per the original issue (vs the existing `/assignedopdivs` family) — can revisit in a follow-up.
- `deriveIdentityProvider` returning 500 after a successful commit is left as-is to match the existing OpDiv-grant handlers; worth a separate conversation on rollback-on-derive-fail vs log-and-return.

CI now runs on this branch in-repo. Requesting review. Once green, this can merge and deploy so the paired UI (ztmf-ui#438) works.

MackOverflow (approved):
Look clean, no nits

PR #371, opened 2026-06-29
Regenerate openapi.yaml for the batch OpDiv grants endpoint

jsos3-cms:
Regenerates `backend/openapi.yaml` from the Go handler annotations.

#370 added `PUT /api/v1/users/{userid}/opdivs` (`SetUserOpDivs`) with complete swag annotations but did not commit the regenerated spec, so the "openapi spec up to date" check is failing on `main` and gating the dev deploy. This adds the missing operation and its `setUserOpDivsInput` request schema — generated with the pinned swag (v2.0.0-rc5) the same way CI does (`make generate-openapi`), so it is byte-identical to what CI regenerates (60 additive lines, no other churn).

Refs #370

## Test plan
- [x] `make generate-openapi` produces no further diff (spec matches the annotations)
- [ ] CI: "openapi spec up to date" passes; Redocly lint passes (the operation declares `bearerAuth`)

### DIFF
--- a/backend/openapi.yaml
+++ b/backend/openapi.yaml
@@ -230,6 +230,14 @@ components:
         idp:
           type: string
       type: object
+    controller.setUserOpDivsInput:
+      properties:
+        opdiv_ids:
+          items:
+            type: integer
+          type: array
+          uniqueItems: false
+      type: object
     model.AuditRef:
       properties:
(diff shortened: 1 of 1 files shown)


MackOverflow (approved):
Clean!

PR #373, opened 2026-06-30
ci: always run the rollout in Manual Deploy to DEV

jsos3-cms:
## What

Fixes the `Manual Deploy to DEV` workflow so it always performs the rollout. Removes the `include_infrastructure` input and runs the Infrastructure job unconditionally.

## Why

`backend.yml` only builds, smoke-tests, and pushes the image to ECR — it does **not** deploy it. The actual rollout is the `terraform apply` in `infrastructure.yml` (`ecs.tf` reads the image tag from the `ztmf_api_tag` SSM parameter). The previous `include_infrastructure: false` default therefore built an image without ever deploying it — a no-op for a "deploy to dev" action. Making the apply optional/off-by-default was a mistake; it is the rollout.

## Notes

- Known limitation (separate fix): re-running the manual deploy on an unchanged `main` SHA fails on the immutable ECR tag push in `backend.yml`. A future change should make that push idempotent (skip when the tag already exists) so the run falls through to the rollout.
- Opening this PR also exercises the standard dev pipeline: the `diff` gate sees no `backend/` change, so the build is skipped and Infrastructure rolls out the current API image already in ECR to dev.

### DIFF
--- a/.github/workflows/manual-deploy-dev.yml
+++ b/.github/workflows/manual-deploy-dev.yml
@@ -1,17 +1,15 @@
 name: Manual Deploy to DEV
 
-# Deploy main to DEV on demand. API/app only by default; Terraform apply is opt-in.
+# Deploy main to DEV on demand.
 #   - orchestration-dev (open/sync a PR): preview a feature branch on DEV.
 #   - this workflow: (re)deploy main to DEV, e.g. to reset DEV after branch deploys.
-# Always redeploys (skips the diff gate). Run:
-#   gh workflow run "Manual Deploy to DEV" --ref main [-f include_infrastructure=true]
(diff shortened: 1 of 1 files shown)


MackOverflow (approved):
looks good

MackOverflow:
Hey @jsos3-cms — this PR is approved and all CI/DEV gates are green, but it's blocked by the `required_signatures` branch ruleset on `main`. The commit needs to be GPG or SSH signed before it can merge.

To fix: sign the commit locally and force-push:
```bash
git commit --amend --no-edit -S
git push origin ci/manual-deploy-rollout-fix --force-with-lease
```

If your Git is already configured with a signing key, that's all it takes.

PR #374, opened 2026-06-30
Tarratsco/bugfix/prevent admin self deletion

tarratsco:
## Summary

Closes `CMS-Enterprise/ztmf#366`. Pairs with the FE half at `CMS-Enterprise/ztmf-ui#433`, which shipped first.

Pre-existing bug: `DELETE /api/v1/users/<id>` accepted the request regardless of whether the caller was deleting their own account. An admin who self-deleted got soft-deleted, the next `/users/current` call no longer found their row, and they were locked out with no FE-recoverable path back in — recovery required a sibling admin running `PUT /users/<id>/restore`.

The FE half disables the row-action icon on the caller's own row plus a handler short-circuit. That's sufficient for the accidental click. This PR is the **defense-in-depth half**: the BE rejects the request server-side so a direct `curl`, a scripted client, or a future FE refactor that re-enables the affordance cannot bypass.

## Changes

* **`backend/cmd/api/internal/auth/middleware.go`**
  * `writeJSONError` → `WriteJSONError` (export so it's callable from `controller`). Signature, body, and behavior unchanged. The six internal call sites in this file are updated in the same diff.
  * Added `CodeSelfDeleteForbidden = "SELF_DELETE_FORBIDDEN"` alongside the existing `Code*` family. Keeping all response codes in one place since the FE mirrors them in `utils/authCodes.ts`, regardless of which BE package emits them.
* **`backend/cmd/api/internal/controller/users.go`**
  * New import: `auth`.
  * In `DeleteUser`, after the admin check and path-param lookup, compare `userID == authdUser.UserID` and short-circuit with `auth.WriteJSONError(w, http.StatusForbidden, "You cannot delete your own account.", auth.CodeSelfDeleteForbidden)`. Placed before the DB lookup so the rejection costs no query and fires even if the row was somehow corrupted mid-session.
* **`backend/cmd/api/internal/controller/users_test.go`**
  * Three new tests covering the AC's three cases.

Updated order of checks in `DeleteUser`:

```
1. authdUser.IsAdmin()              -> 403 ErrForbidden if not
2. vars["userid"] present?          -> 404 ErrNotFound if missing
3. userID == authdUser.UserID?      -> 403 SELF_DELETE_FORBIDDEN if true
4. model.FindUserByID               -> errors as today (404/500)
5. authdUser.CanManageUser(target)  -> 403 ErrForbidden if scope fails
6. model.DeleteUser                 -> soft-delete (204 success)
```

## Test plan

Backend unit tests:

[...]

PR #375, opened 2026-06-30
Bugfix/prevent admin self deletion

### LINKED ISSUE #366: Reject self-delete on `DELETE /users/:id`
## Recommended labels

`bug`, `security`, `area/api`, `lang/go`, `cross-repo`

## Context

`DELETE /api/v1/users/<id>` today accepts the request regardless of whether the authenticated caller is deleting their own account. A logged-in admin who self-deletes locks themselves out of the app with no FE-recoverable path: the next API call 401s (the BE no longer recognizes the deleted user), the FE interceptor redirects to `/signin` with "session expired," and the user cannot get back in because their row is gone. Recovery requires another admin to run `PUT /users/<id>/restore` against the soft-deleted row.

The FE half of this has shipped at `CMS-Enterprise/ztmf-ui#433` (PR open against `tarratsco/bugfix/add-guard-against-admins-deleting-themselves`): the Delete row-action icon on the current user's own row is disabled with a tooltip, and `handleConfirmDelete` short-circuits as a backstop. That is sufficient to prevent the accidental UI click. This ticket is the **defense-in-depth half** — the BE should reject the request server-side so a direct API call (older client, scripted call, future refactor that wires a new entry point) cannot bypass the FE guard.

## Problem

`DELETE /api/v1/users/<id>` is currently a single check: caller has admin role + target row exists + target is within scope → mark deleted. No "is the target the caller?" check.

Consequences:

1. **Account lockout.** The deleted admin cannot sign back in. They authenticate at the IdP (session still valid), but `/users/current` now 401s because no row matches the identity. Post-`#403` shipping, this becomes a `403 + ACCOUNT_NOT_PROVISIONED` and the user sees a terminal "contact your administrator" state — still locked out, but at least with honest copy.
2. **Recovery requires a sibling admin.** If the deleted admin was the only admin in scope, the system has no path to restoration short of database-level intervention.
3. **FE-only fix is bypassable.** The shipped FE guard is two checks (disabled icon + handler short-circuit). A direct `curl -X DELETE /api/v1/users/<own-id>` with a valid bearer token still succeeds today.

## Proposed change

In the handler for `DELETE /api/v1/users/<id>`, before issuing the soft-delete, compare the path id against the authenticated caller's id (already resolved on the request context by the auth middleware). If they match,  403 with a typed JSON body.

### 403 Forbidden

Argument: the caller has admin-level write rights generally, but not for this specific target (themselves). Authorization rule, not validation rule.

```json
HTTP/1.1 403 Forbidden
{"error": "You cannot delete your own account.", "code": "SELF_DELETE_FORBIDDEN"}
```
### Code path

The auth middleware (`backend/cmd/api/internal/auth/middleware.go`) already resolves the authenticated user onto the request context. The DELETE handler should:

1. Pull the caller's id from context.
2. Parse the path id from the URL.
3. If equal, call `writeJSONError(w, http.StatusForbidden, "You cannot delete your own account.", "SELF_DELETE_FORBIDDEN")` (matching the helper from `#403`'s middleware refactor) and return.
4. Otherwise, proceed with the existing delete logic.

### Steps to Reproduce

Surfaced during ad-hoc smoke testing on the `ztmf-ui` Users page. Reproduces today against any deployed environment: log in as Admin/OWNER, open DevTools → Network, click the Delete icon on your own row, confirm the dialog. The DELETE goes through, the row disappears, the next API call 401s, and the page redirects to `/signin` with no path back in.

Tracked publicly at `CMS-Enterprise/ztmf-ui#433`.

### Expected Behavior

Admins get an error message when they try to delete their own accounts.

### Actual Behavior

Admins can successfully delete their own accounts.

### Environment

Local Dev API 

### Acceptance Criteria

- [ ]  `DELETE /api/v1/users/<id>` rejects the request with 403 + `code: "SELF_DELETE_FORBIDDEN"` when the path id matches the authenticated caller's id.
- [ ] The error body is JSON in the same shape as other middleware-tier rejections (`{error, code}`).
- [ ] Deleting other users continues to work as today (admin permissions and OpDiv scope rules unchanged).
- [ ] Unit test on the handler covers all three cases (self-delete rejected, other-delete allowed, no-id/malformed → existing behavior).
- [ ] BE PR description records the chosen status code (Option A vs B) so the FE side knows what to branch on.

### Severity

High - Significantly impacts usage

### Affected Component

_No response_

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

_No response_

a-dipietro:
## Summary

Pre-existing bug: `DELETE /api/v1/users/<id>` accepted the request regardless of whether the caller was deleting their own account. An admin who self-deleted got soft-deleted; their next `/users/current` no longer found their row and they were locked out, recoverable only by a sibling admin running `PUT /users/<id>/restore`. Closes #366. Pairs with the FE half at CMS-Enterprise/ztmf-ui#433 (shipped first). In-repo rehome of #374 by @tarratsco so the org-secret CI gates (Snyk, Infrastructure/Deploy) can run — they can't on a fork PR; original authorship preserved.

The FE half disables the row action on the caller's own row plus a handler short-circuit — enough for the accidental click. This is the defense-in-depth backend half: the server rejects the request so a direct curl, a scripted client, or a future FE regression cannot bypass it.

## Changes

- **`controller/users.go`** — `DeleteUser` now rejects self-deletion. After the admin check and path-param lookup, it compares the target id to the caller's id with `strings.EqualFold(userID, authdUser.UserID)` and short-circuits with `respond(w, r, nil, ErrSelfDelete)` **before** the DB lookup — so the rejection costs no query and fires against the JWT-derived caller id even if the caller's own row is in an odd state. A `case_variant` boolean is logged to tell an honest self-click apart from a differently-cased probe. `EqualFold` mirrors the case-insensitive semantics of the native Postgres `uuid` column. Also adds package-level `findUserByID`/`deleteUser` seams (same pattern as `auth/middleware.go`) so `DeleteUser` is unit-testable without a database, with compile-time assertions guarding against model-signature drift.
- **`controller/errors.go`** — adds `ErrSelfDelete` (`"cannot delete your own account"`).
- **`controller/controller.go`** — `sanitizeErr` now returns `(status, code, err)` and maps `ErrSelfDelete` → 403 with `code = auth.CodeSelfDeleteForbidden`. The `response` struct gains an optional `code` field (`omitempty`), so successful responses and generic errors keep their existing shape; only typed-code errors carry it.
- **`auth/middleware.go`** — adds `CodeSelfDeleteForbidden = "SELF_DELETE_FORBIDDEN"` to the existing `Code*` block, where response codes live so the FE can mirror them in `utils/authCodes.ts`.
- **`controller/authorization_test.go`** — four new tests (below).

Check order in `DeleteUser`:
1. `authdUser.IsAdmin()` → 403 `ErrForbidden` if not
2. `vars["userid"]` present? → 404 `ErrNotFound` if missing
3. self? (`EqualFold(userID, caller)`) → 403 `SELF_DELETE_FORBIDDEN`
4. `findUserByID` → 404/500 as today
5. `CanManageUser(target)` → 403 `ErrForbidden` if OpDiv scope fails
6. `deleteUser` → 204

## Test plan

- `TestDeleteUser_SelfDeleteRejected` — self-delete returns 403; body `code` is `SELF_DELETE_FORBIDDEN`.
- `TestDeleteUser_SelfDeleteRejected_CaseInsensitive` — upper/mixed-case variants of the caller's own id are still rejected (locks in the `uuid`/`EqualFold` semantics).
- `TestDeleteUser_OtherDeleteSucceeds` — deleting a different user (including an upper-cased id) still succeeds.
- `TestDeleteUser_MissingIDIsNotFound` — pre-existing 404 branch unchanged.
- `go test ./cmd/api/...` and `go vet ./cmd/api/...` green.

## Notes for reviewers

No API shape change for existing consumers: the new `response.code` field is `omitempty` and set only for typed-code errors. The self-delete path reuses the existing `respond()`/`sanitizeErr()` error-mapping pipeline rather than writing responses directly, keeping error handling in one place.

Refs #374. Original contribution by @tarratsco.

### DIFF
--- a/backend/cmd/api/internal/auth/middleware.go
+++ b/backend/cmd/api/internal/auth/middleware.go
@@ -22,6 +22,7 @@ const (
 	CodeUnauthorized          = "UNAUTHORIZED"
 	CodeForbiddenOrigin       = "FORBIDDEN_ORIGIN"
 	CodeAccountNotProvisioned = "ACCOUNT_NOT_PROVISIONED"
+	CodeSelfDeleteForbidden = "SELF_DELETE_FORBIDDEN"
 )
 
 // Package-level seams over the model lookups so tests can stub them without a
@@ -39,10 +40,10 @@ type errorBody struct {
 	Code  string `json:"code,omitempty"`
 }
 
-// writeJSONError writes a standardized JSON error response and is the only
+// WriteJSONError writes a standardized JSON error response and is the only
 // rejection surface used by Middleware. Centralizing the shape keeps the FE
 // interceptor's contract single-sourced.
-func writeJSONError(w http.ResponseWriter, status int, msg, code string) {
+func WriteJSONError(w http.ResponseWriter, status int, msg, code string) {
 	w.Header().Set("Content-Type", "application/json")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.WriteHeader(status)
@@ -70,7 +71,7 @@ func Middleware(next http.Handler) http.Handler {
 
 		claims, isSession, ok := claimsFromRequest(r)
 		if !ok {
-			writeJSONError(w, http.StatusUnauthorized,
+			WriteJSONError(w, http.StatusUnauthorized,
 				"Your session has expired. Please sign in again.",
 				CodeUnauthorized)
 			return
@@ -82,7 +83,7 @@ func Middleware(next http.Handler) http.Handler {
 		// cookie path is browser-driven; the bearer path is for API clients and
 		// is not subject to CSRF.
 		if isSession && !isSafeMethod(r.Method) && !sameOrigin(r) {
-			writeJSONError(w, http.StatusForbidden,
+			WriteJSONError(w, http.StatusForbidden,
 				"Request blocked: origin not allowed.",
 				CodeForbiddenOrigin)
 			return
@@ -120,7 +121,7 @@ func Middleware(next http.Handler) http.Handler {
 			user, err = user.Save(r.Context())
 			if err != nil {
 				log.Printf("Failed to auto-create user: %s\n", err)
-				writeJSONError(w, http.StatusInternalServerError,
+				WriteJSONError(w, http.StatusInternalServerError,
 					"internal error", "")
 				return
 			}
@@ -131,7 +132,7 @@ func Middleware(next http.Handler) http.Handler {
 			// this code to render a terminal "contact your administrator"
 			// message instead of looping the user back through the IdP.
 			log.Printf("authenticated identity has no ZTMF account: %s\n", IdentifierFromClaims(claims))
-			writeJSONError(w, http.StatusForbidden,
+			WriteJSONError(w, http.StatusForbidden,
 				"Your ZTMF account is not set up. Contact your administrator to request access.",
 				CodeAccountNotProvisioned)
 			return
@@ -139,7 +140,7 @@ func Middleware(next http.Handler) http.Handler {
 			// DB connection blip, decode failure, etc. Not a credential
 			// problem, so do not present as one to the FE.
 			log.Printf("user lookup failed: %s\n", err)
-			writeJSONError(w, http.StatusInternalServerError,
+			WriteJSONError(w, http.StatusInternalServerError,
 				"internal error", "")
 			return
 		}
@@ -150,7 +151,7 @@ func Middleware(next http.Handler) http.Handler {
 			// distinctly so support can tell "offboarded" from "never
 			// onboarded" without grepping the users table.
 			log.Printf("deleted user attempted to access the API: %s\n", user.Email)
-			writeJSONError(w, http.StatusForbidden,
+			WriteJSONError(w, http.StatusForbidden,
 				"Your ZTMF account is no longer active. Contact your administrator.",
 				CodeAccountNotProvisioned)
 			return
--- a/backend/cmd/api/internal/auth/middleware.go
+++ b/backend/cmd/api/internal/auth/middleware.go
@@ -40,10 +40,10 @@ type errorBody struct {
 	Code  string `json:"code,omitempty"`
 }
 
-// WriteJSONError writes a standardized JSON error response and is the only
+// writeJSONError writes a standardized JSON error response and is the only
 // rejection surface used by Middleware. Centralizing the shape keeps the FE
 // interceptor's contract single-sourced.
-func WriteJSONError(w http.ResponseWriter, status int, msg, code string) {
+func writeJSONError(w http.ResponseWriter, status int, msg, code string) {
 	w.Header().Set("Content-Type", "application/json")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.WriteHeader(status)
@@ -71,7 +71,7 @@ func Middleware(next http.Handler) http.Handler {
 
 		claims, isSession, ok := claimsFromRequest(r)
 		if !ok {
-			WriteJSONError(w, http.StatusUnauthorized,
+			writeJSONError(w, http.StatusUnauthorized,
 				"Your session has expired. Please sign in again.",
 				CodeUnauthorized)
 			return
@@ -83,7 +83,7 @@ func Middleware(next http.Handler) http.Handler {
 		// cookie path is browser-driven; the bearer path is for API clients and
 		// is not subject to CSRF.
 		if isSession && !isSafeMethod(r.Method) && !sameOrigin(r) {
-			WriteJSONError(w, http.StatusForbidden,
+			writeJSONError(w, http.StatusForbidden,
 				"Request blocked: origin not allowed.",
 				CodeForbiddenOrigin)
 			return
@@ -121,7 +121,7 @@ func Middleware(next http.Handler) http.Handler {
 			user, err = user.Save(r.Context())
 			if err != nil {
 				log.Printf("Failed to auto-create user: %s\n", err)
-				WriteJSONError(w, http.StatusInternalServerError,
+				writeJSONError(w, http.StatusInternalServerError,
 					"internal error", "")
 				return
 			}
@@ -132,7 +132,7 @@ func Middleware(next http.Handler) http.Handler {
 			// this code to render a terminal "contact your administrator"
 			// message instead of looping the user back through the IdP.
 			log.Printf("authenticated identity has no ZTMF account: %s\n", IdentifierFromClaims(claims))
-			WriteJSONError(w, http.StatusForbidden,
+			writeJSONError(w, http.StatusForbidden,
 				"Your ZTMF account is not set up. Contact your administrator to request access.",
 				CodeAccountNotProvisioned)
 			return
@@ -140,7 +140,7 @@ func Middleware(next http.Handler) http.Handler {
 			// DB connection blip, decode failure, etc. Not a credential
 			// problem, so do not present as one to the FE.
 			log.Printf("user lookup failed: %s\n", err)
-			WriteJSONError(w, http.StatusInternalServerError,
+			writeJSONError(w, http.StatusInternalServerError,
 				"internal error", "")
 			return
 		}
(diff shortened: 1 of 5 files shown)


jsos3-cms:
## Review — one must-fix, recommended tests, logging/monitoring, and a merge-timing hold

Thanks for rehoming this so the CI gates can run. The approach (server-side defense-in-depth so a direct `curl` / scripted client / future FE refactor can't self-delete) is right, and the ordering table in the description is sound. The diff is byte-identical to #374, so this applies to both. Below: a blocking correctness issue, tests to lock it down, how it should surface in logs/monitoring, and the merge sequence relative to #377.

---

### Merge timing — please do not merge or mark "ready for review" until #377 is on `main`

Dev deploys run **from the PR branch** (Orchestration DEV builds the API image + runs `terraform apply` on `ready`/`synchronize` whenever the PR is non-draft). This branch is `main` + the self-delete fix, but it does **not** contain the Entra login fix from #377 (`tid`/`aud` relaxation, commit `bac7b1e`, currently live on dev as task-def `api:341`). So marking this ready — or pushing another commit while it's non-draft, or merging — would deploy an API image **without** the Entra fix and revert it on dev, breaking the Entra sign-in that's mid-test there.

Sequence:
1. Merge **#377 → `main`** (Entra fix).
2. **Rebase this branch on `main`** so it picks up the Entra fix.
3. Land the case fix + test below.
4. Then mark ready / merge #375.

Until then, keeping this in draft (or holding pushes) is the safe state.

---

### Must fix — the self-delete guard is bypassable via UUID case

The guard compares the raw path param to the session user with a case-sensitive string `==`:

```go
// controller/users.go — DeleteUser
if userID == authdUser.UserID {
    auth.WriteJSONError(w, http.StatusForbidden, "You cannot delete your own account.", auth.CodeSelfDeleteForbidden)
    return
}
```

This is a **schema-typing mismatch**. `users.userid` is a Postgres **`uuid`** column (`migrations/0001initialschema.go`: `userid uuid DEFAULT gen_random_uuid()`), which compares case-insensitively and stores canonically lowercase. The route pattern (`userIdPattern`, `router/router.go`) accepts `[a-zA-Z0-9]`, and `isValidUUID`/`rgxUUID` (`model/validations.go`) accept mixed-case hex — so a request carrying the caller's own id with hex letters uppercased routes through, `model.FindUserByID` resolves it to the same row (uuid equality), and `model.DeleteUser` deletes it. But `authdUser.UserID` is the canonical-lowercase value from the DB, so the Go `==` is **false** and the guard falls through. Net:

> `DELETE /api/v1/users/<own-id-with-hex-letters-uppercased>` self-deletes successfully — re-opening exactly the bug this PR closes.

**Fix — reconcile the comparison with the `uuid` semantics.** Minimal:

```go
if strings.EqualFold(userID, authdUser.UserID) { ... }
```

Stronger and most faithful to the schema: parse both sides as a UUID and compare (also rejects malformed input) — e.g. `google/uuid` `uuid.Parse(userID) == uuid.Parse(authdUser.UserID)`. Either way, this mirrors the discipline already used for the email key, which is lowercased before comparison in `FindUserByEmail`. Worth a quick grep for other spots that string-`==` a path/param userid against a DB value — the same fragility would apply, though this guard is the one that matters here.

---

### Recommended tests

The three new tests cover the happy paths but wouldn't have caught the above (the fixtures use literal lowercase UUIDs). Suggested additions, in the existing `withUser` / `httptest.NewRecorder` style:

1. **Case-insensitive self-delete is rejected (regression test for this finding).** Table-driven over case variants of the admin's own id — all-upper and a mixed-case flip — each asserting **403** with code `SELF_DELETE_FORBIDDEN`:

   ```go
   func TestDeleteUser_SelfDeleteRejected_CaseInsensitive(t *testing.T) {
       admin := ownerFixture // UserID stored lowercase
       for _, id := range []string{strings.ToUpper(admin.UserID), mixedCase(admin.UserID)} {
           rec := httptest.NewRecorder()
           DeleteUser(rec, withUser(deleteReq(id), admin))
           assert.Equal(t, http.StatusForbidden, rec.Code)
           assert.Contains(t, rec.Body.String(), auth.CodeSelfDeleteForbidden)
       }
   }
   ```
   With the fix in place the guard short-circuits before any DB call, so this needs no DB. Against the current code it goes red (bypass), against the fixed code it's green — a true regression guard.

2. **Strengthen the other-user test.** `TestDeleteUser_OtherDeleteNotSelfError` currently only asserts the code *isn't* `SELF_DELETE_FORBIDDEN`; it would still pass if the delete were wrongly blocked for an unrelated reason. Assert the other-user delete actually **succeeds** (its real status), and add a case-variant of a *different* user's id to confirm the (now case-insensitive) guard doesn't over-match and block a legitimate other-user delete.

3. **(Optional, integration)** A model-level test documenting that `FindUserByID`/`DeleteUser` treat an uppercased UUID as the same row — this pins the DB case-insensitivity that motivates normalizing in the guard, so the reasoning survives future refactors. Needs a DB, so mark it integration.

---

### Logging / erroring / monitoring

Worth tightening, because right now this event is nearly invisible:

- **The block path does not log.** `auth.WriteJSONError` writes only the HTTP 403 — no `ztmf_api` line — so a blocked self-delete (accidental *or* a probing attempt) leaves no server-side trace.
- **Successful deletes are audited, self-deletes included.** `model.DeleteUser` goes through `queryRow`'s automatic `recordEvent` hook → an `events` row (`userid`=actor, `action`, `resource=users`, `payload`). So today's bypass is recorded as an ordinary successful delete where **actor == target** — recoverable forensically by querying `events`, but with no active signal.

Recommendations:

- **Emit a structured log line when the guard fires**, matching the existing `login: reject branch=…` style (ids/booleans only — `userid` is an internal UUID, not PII; the path value is route-constrained to `[a-zA-Z0-9-]`, so it's safe to log). Include a boolean for whether the raw strings differed only by case, which distinguishes an accidental click (exact match) from a deliberately case-mangled id (probing):

  ```go
  caseVariant := strings.EqualFold(userID, authdUser.UserID) && userID != authdUser.UserID
  log.Printf("user: self-delete blocked actor=%s target=%s case_variant=%t\n",
      authdUser.UserID, userID, caseVariant)
  ```
  (`case_variant=true` means the id matched only after case-folding — i.e. a deliberately mangled self-id, worth distinguishing from an exact-match accidental click.)

- **Optional alerting**, if we want this monitored rather than just logged: a CloudWatch Logs metric filter + alarm on that line, routed to the existing `ztmf-alarms` SNS topic (same pattern as the ELBAuth login alarms). A backstop worth noting: a periodic check of the `events` table for a delete where actor == target would catch any self-delete that ever slips the guard again.

---

### Suggestion (non-blocking) — error-shape consistency

The guard uses `auth.WriteJSONError`, which emits `{"error":…, "code":…}`, while the other 403s in this same handler (the `IsAdmin` and `CanManageUser` checks) go through `respond(...)` → `{"data":null,"error":"forbidden"}`. Two different JSON shapes for a 403 from one endpoint means the FE has two contracts to branch on, and exporting `writeJSONError`→`WriteJSONError` to call the middleware's writer from a controller crosses the layer boundary that helper documents for itself. Since the FE keys off `SELF_DELETE_FORBIDDEN`, a cleaner fit is a controller-side `ErrSelfDelete` sentinel with a `sanitizeErr` case that still routes through `respond`, keeping the error shape uniform. Not blocking, but it'll save the FE a special case.

The self-check-before-scope-check ordering is fine as-is (fails fast, self-delete is disallowed regardless of scope).

---

Once #377 is merged and this is rebased, with the case fix + regression test in, this is good to go.

jsos3-cms:
Hey @tarratsco, please make the changes on your branch under #374, then tag me or @MackOverflow to rehome the new commit onto this PR.

@MackOverflow , we may want to consider moving this to draft until #377 is in and doing a rebase, to catch any accidental deploys.

MackOverflow (dismissed):
caught two fixes, not blockers per se

jsos3-cms (approved):

PR #380, opened 2026-07-02
Enhancement/hhs schema migration

### LINKED ISSUE #379: HHS datacall integration: schema migration, model + RBAC changes, and local mock data
### Feature Type

General Feature

#### Summary

Land the backend schema and API changes required to support HHS onboarding datacalls. Three areas:

1. Database migration adding 11 nullable HHS metadata columns to `fismasystems` and a `notes_is_ai_summary` boolean to `scores`.
2. Model and API changes to expose the new fields through all relevant read and write endpoints.
3. RBAC gate ensuring only HHS-wide admins (OWNER / HHS_ADMIN) can write the 11 HHS metadata fields; scoped admins see them read-only.

Also includes local development mock data: 3 HHS datacalls, 15 MOCK-\* systems, and 378 scores (21 AI-flagged) appended to `_test_data_empire.sql` so HHS functionality is fully exercisable in a local dev environment.

#### Depends on

- Frontend PR: renders the 11 new fields on the system detail page / edit modal and the AI Summary badge on questionnaire notes. Can be developed against this backend PR independently; requires a coordinated local test before shipping.

##### 1. Migration (`backend/cmd/api/internal/migrations/0043hhsschemadelta.go`)

- `ALTER TABLE fismasystems ADD COLUMN IF NOT EXISTS` for 11 columns, all `varchar(255)` nullable:
  `hva`, `fips`, `system_type`, `cloud_system`, `cloud_service_model`, `cloud_vendor`, `system_operator`, `goco_coco_gogo`, `system_owner`, `system_owner_email`, `legacy`
- `ALTER TABLE scores ADD COLUMN IF NOT EXISTS notes_is_ai_summary boolean NOT NULL DEFAULT FALSE`
- Fully reversible DOWN migration using `DROP COLUMN IF EXISTS`.

##### 2. Model — fismasystems (`backend/internal/model/fismasystems.go`)

- Extend `fismaSystemColumns` with the 11 new column names.
- Add 11 `*string` fields to the `FismaSystem` struct with matching `json` and `db` tags.
- Replace the positional `fismaSystemColumns[1:13]` INSERT slice with an explicit named column list (the decommission/reactivation audit fields in positions 13–19 must be excluded from INSERT; no clean slice range covers this).
- Add `.Set()` calls for the 11 new fields in the UPDATE path.

##### 3. Model — scores (`backend/internal/model/scores.go`)

- Add `NotesIsAISummary *bool` to the `Score` struct (`*bool` not `bool` — see design note below).
- Include `notes_is_ai_summary` in INSERT `Columns()`, `Values()` (dereferenced with a `false` default), and RETURNING suffix.
- On UPDATE: include `notes_is_ai_summary` in the SET clause **only when the incoming pointer is non-nil**. A `nil` value means the client omitted the field and the stored value is preserved; an explicit `false` clears the flag.
- Include `notes_is_ai_summary` in the `scoresEqualForUpdate` comparison only when the incoming pointer is non-nil (same nil-skip rule as the UPDATE path).
- Include in `FindScores` SELECT, `copyPreviousScores` INSERT/SELECT (positional match required), and RETURNING on UPDATE (returns the stored value).

##### 4. Model — answers (`backend/internal/model/answers.go`)

- Add `NotesIsAISummary *bool` to the `Answer` struct with `db:"notes_is_ai_summary"`.
- Add `scores.notes_is_ai_summary` to the `FindAnswers` SELECT so exports carry the flag.

##### 5. Controller — RBAC gate (`backend/cmd/api/internal/controller/fismasystems.go`)

- `clearHHSMetadata(fs)` — nils all 11 HHS fields; called on INSERT when the acting user lacks `HasUnscopedRead()`.
- `copyHHSMetadata(src, dst)` — copies stored HHS field values from the existing DB record onto the incoming request body; called on UPDATE for scoped users so their edit cannot wipe metadata they cannot see.
- Gate is in `SaveFismaSystem` after the existing `IsAdmin()` and OpDiv-scope checks.

##### 6. OpenAPI spec

- Regenerated via `make generate-openapi` to include all 11 new `fismasystems` fields (`type: string`) and `notes_is_ai_summary` (`type: boolean`).

##### 7. Local mock data (`backend/_test_data_empire.sql`)

- 3 HHS datacalls (FY23 / FY24 / FY25 ZTM, IDs 101–103)
- 15 MOCK-\* fismasystems (IDs 2001–2015) with HHS metadata populated
- 378 scores across the 3 datacalls; 21 scores with `notes_is_ai_summary = true`
- Block is idempotent (wrapped in `BEGIN/COMMIT`, DELETEs its own rows before re-inserting) and gated to local dev by the existing `DB_POPULATE` mechanism

#### Design note — `notes_is_ai_summary` as `*bool`

The field uses a pointer rather than a plain `bool` to distinguish two cases that a zero-value `bool` cannot:

| Client sends | Decoded as | Effect |
|---|---|---|
| Field omitted / `null` | `nil` | Stored value unchanged |
| `false` | `*false` | Flag cleared — user replaced the AI-generated note |
| `true` | `*true` | Flag set (AI summarizer path) |

Without the pointer, a questionnaire PUT from a client that omits the field decodes to `false` and would silently clear the flag on every save. The frontend should send `"notes_is_ai_summary": false` when the user edits the notes content, and omit the field otherwise.

#### Out of scope

- `series` column on `datacalls` to distinguish HHS vs CMS datacall types (deferred — requires team discussion on API filtering and UI grouping).
- Spreadsheet export emitting `notes_is_ai_summary` as a column (fast follow up or can add to this PR).
- Any frontend rendering (separate PR).

### Acceptance Criteria

- [ ] `psql \d fismasystems` shows all 11 new columns; `psql \d scores` shows `notes_is_ai_summary boolean not null default false`.
- [ ] `go build ./...` compiles cleanly; `go test ./...` passes.
- [ ] `GET /api/v1/fismasystems/{id}` returns all 11 HHS fields for MOCK-\* systems.
- [ ] `GET /api/v1/answers` returns `notes_is_ai_summary` on each answer row via `FindAnswers`.
- [ ] As OWNER: `PUT /api/v1/fismasystems/{id}` with HHS fields set → fields persist (verified by subsequent GET).
- [ ] As OPDIV_ADMIN: `PUT /api/v1/fismasystems/{id}` with changed HHS values → fields are unchanged after save (stored values copied back by `copyHHSMetadata`).
- [ ] Questionnaire PUT that **omits** `notes_is_ai_summary` on an AI-flagged score → flag remains `true` in DB.
- [ ] Questionnaire PUT that sends `"notes_is_ai_summary": false` on an AI-flagged score → flag is cleared to `false` in DB.
- [ ] `make dev-setup` loads MOCK-\* data; 15 systems appear in the system list and carry HHS metadata; 21 scores are AI-flagged.
- [ ] OpenAPI spec reflects all new fields.

### How to test locally

1. Pull this branch and run `make dev-setup` (migration runs automatically on API startup).
2. Confirm migration:

```
psql "postgresql://admin:localdevpassword@localhost:54321/ztmf" -c "\d fismasystems"
psql "postgresql://admin:localdevpassword@localhost:54321/ztmf" -c "\d scores"
```

3. Confirm mock data:

```
SELECT COUNT(*) FROM fismasystems WHERE fismaacronym LIKE 'MOCK-%';  -- expect 15
SELECT COUNT(*) FROM scores WHERE notes_is_ai_summary = true;        -- expect 21
```

4. Use curl with HS256 JWTs (secret zeroTrust, email claim) to test the RBAC gate as OWNER vs OPDIV_ADMIN per the acceptance criteria above.

costacalvin:
## HHS Schema Integration — Backend
Closes https://github.com/CMS-Enterprise/ztmf/issues/379
### Summary

- Adds a migration (`0043hhsschemadelta.go`) that appends 11 nullable `varchar(255)` HHS inventory columns to `fismasystems` (`hva`, `fips`, `system_type`, `cloud_system`, `cloud_service_model`, `cloud_vendor`, `system_operator`, `goco_coco_gogo`, `system_owner`, `system_owner_email`, `legacy`) and a `notes_is_ai_summary boolean NOT NULL DEFAULT FALSE` column to `scores`. All `ADD COLUMN IF NOT EXISTS` — safe to retry and non-destructive to existing rows.

- Exposes all new fields through the Go model (`fismasystems.go`, `scores.go`, `answers.go`, `scorediff.go`) and regenerates `openapi.yaml`.

- Gates write access to the 11 HHS fields on `HasUnscopedRead()` (OWNER + HHS_ADMIN only). Scoped admins see the fields read-only — on INSERT their values are zeroed (`clearHHSMetadata`), on UPDATE the stored values are copied back from the DB before `Save()` (`copyHHSMetadata`) so a scoped edit cannot silently wipe HHS metadata.

- `notes_is_ai_summary` uses `*bool` semantics in Go: `nil` = client omitted the field (preserve stored value), `false` = explicitly clear the flag, `true` = set it. The flag carries forward verbatim in `copyPreviousScores` — notes that haven't been user-edited are still AI-generated content; the flag only clears when the frontend sends `false` after the user edits the notes field.

- Extends `_test_data_empire.sql` with the HHS mock layer: 3 historical datacalls (FY23–FY25 ZTM), 15 `MOCK-*` FISMA systems across HHS OpDivs, and 378 scores with realistic null patterns and AI-flagged notes for local dev testing.

### Test plan

- [x] `go build ./...` — clean compile
- [x] `go test ./backend/internal/model/...` — all unit tests pass, including 3 new `TestScoresEqualForUpdate` cases covering `*bool` nil/false/true semantics
- [x] `psql \d fismasystems` — 11 new columns present, all nullable varchar(255)
- [ ] `psql \d scores` — `notes_is_ai_summary boolean NOT NULL DEFAULT FALSE`
- [ ] With local dev stack + mock data: `GET /fismasystems` returns MOCK-* systems with HHS fields populated
- [ ] As OWNER: `PUT /fismasystems/{id}` with HHS fields — fields are written to DB
- [ ] As OPDIV_ADMIN: same PUT — HHS fields are preserved from stored values (scoped admin cannot wipe them)
- [ ] As OWNER: `GET /scores/diff?from=101&to=102` — diff includes `notes_is_ai_summary` on `ScoreDiffSide`; flag-only changes surface in the diff
- [ ] Migration rollback: `down` migration drops all added columns cleanly

### Notes for frontend PR

The questionnaire `PUT /scores` endpoint now accepts `"notes_is_ai_summary": false` to clear the AI flag. The field should be sent when the user edits the notes text. Omitting the field (or sending `null`) preserves the stored value — no action needed for read-only views.


danielbowne (commented):
Nice work on this. It's a well-structured PR, especially for how much surface it touches. A few things are genuinely clean, worth calling out before the asks:

- The access-control handling is the strongest part. Gating the 11 HHS fields on unscoped read (clearing them on insert and copying the stored values back on update for scoped admins) means an OpDiv admin can't blank out inventory metadata they can't even see. Right instinct, and correctly implemented.
- The `notes_is_ai_summary *bool` treatment (nil = leave alone, false = clear, true = set) is exactly right, and the three unit tests pin the behavior down well.
- The migration is symmetric: the down cleanly drops everything the up adds, and the `IF [NOT] EXISTS` guards make it safe to re-run.

Two things to square away before this comes out of draft:

**1. The owner / HHS-admin update path can wipe the HHS fields on a partial PUT.**
The `copyHHSMetadata` back-fill only runs for scoped admins (`!HasUnscopedRead()`). For OWNER and HHS_ADMIN the update builder sets all 11 columns unconditionally, so a `PUT` that doesn't include, say, `hva` or `system_owner` writes `NULL` and drops whatever the importer populated. Two ways to handle it:
- confirm the edit form always sends all 11 fields back (then it's a non-issue), or
- make those 11 `.Set()` calls conditional on non-nil, the same way you handled `notes_is_ai_summary`, so an omitted field is preserved for every role and the copy logic becomes belt-and-suspenders.

I'd lean toward the conditional set, since these fields are populated out-of-band and are the most likely to be missing from an interactive form.

**2. E2E coverage for the new behavior.**
We try to keep the Emberfall suite covering every new response field and behavior, not just the happy path; it's caught real regressions for us. Right now the new surface has no E2E cases. Three worth adding:
- an OWNER writing the HHS fields and reading them back,
- a scoped (OpDiv) admin's write being stripped on create / preserved on update, which is the security-relevant path and the most valuable to lock in,
- `notes_is_ai_summary` flipping to false when the notes are edited.

The fixtures in `emberfall_tests.yml` are anchor-based, so it's mostly extending an existing block; happy to point you at the right spots.

Minor: the mock layer in `_test_data_empire.sql` is great for local testing. When you touch seed data, keep `_test_data.sql` in step with it too so both load paths stay green.

Thanks for pushing this forward. The hard parts (access control, the flag semantics) are already done right; it's mostly the write-path edge case and test coverage left.

danielbowne (commented):
One more note to make the coverage point easy to act on: the full end-to-end suite runs in an isolated, throwaway database, so it is safe to run locally as often as you like and it never touches your dev data.

From the repo root:

```
make test-e2e
```

That spins up a fresh Postgres in Docker, applies every migration on a clean schema, seeds it from `backend/_test_data_empire.sql` (including the new `MOCK-*` HHS layer you added), runs the assertions in `backend/emberfall_tests.yml`, and tears the whole thing down at the end. Nothing persists, and it will not interfere with `make dev-up`.

The same suite runs in CI, so it is worth running locally before you push. Catching a red here saves a failed CI round-trip.

I ran it against this branch and it comes back green (118 passing), so the schema change and the mock data both load cleanly on a fresh database. The only gap is that none of those 118 cases exercise the 11 HHS fields or `notes_is_ai_summary` yet, which is what item 2 above is about. The workflow for closing it:

1. add the new fields, plus the OWNER and OpDiv-admin cases, to `backend/emberfall_tests.yml` (the fixtures are anchor-based, so it is mostly extending an existing block),
2. run `make test-e2e` and confirm it stays green.

If you want a faster loop while writing assertions, `make test-integration` runs the Go-level DB tests against the same kind of isolated database.

**On macOS:** it just needs Docker running (Docker Desktop). One catch right now: the install line the Makefile prints on a missing binary points at the Linux build of the test runner. On a Mac, grab the Darwin build:

```
curl -L https://github.com/aquia-inc/emberfall/releases/download/v0.4.1/emberfall_Darwin_arm64.tar.gz | tar -xz && mv emberfall /usr/local/bin/
```

I will open an issue to make that install step OS-aware so it just works on a Mac.

Running it inside tmux needs no special command. Just make sure `docker ps` works in that shell first. If `docker` is not found, your tmux session predates the Docker Desktop install, so start a fresh shell to pick up the updated PATH.

Happy to pair on the first HHS case if that helps.

costacalvin:
@danielbowne Added new tests to emberfall, verified all passing with `make test-e2e` + write edge cases fixed following similar pattern of notes_is_ai_summary. small change to seed data, verified seed data still loads correctly on rebuild + in the UI. Updated the openapi.yaml. Ready for another review!

danielbowne (approved):
Re-reviewed the latest push, and ran the e2e suite locally to confirm (0 failures). Both points from the last round are handled well.

**Partial-PUT clobber (fixed).** The conditional `SetMap` in `fismasystems.go` is exactly right: the HHS columns only join the SET clause when non-nil, so an owner's partial update no longer nulls out importer-populated data. Test A locks it in - create with HHS values, then a core-only PUT, then a GET proving the values survived. That was the regression I was most concerned about, now covered end to end.

**E2E coverage (resolved).** Tests A/B/C are a thorough addition. Test B is the one I'm happiest to see: a scoped admin's "OVERWRITE ATTEMPT" PUT followed by a GET confirming the owner's values held - that exercises the actual RBAC enforcement, not just the happy path. Test C covers the AI-flag semantics including the nil-means-preserve case.

**Seed fix is correct.** Switching the score inserts to explicit column lists is the right call - the positional `VALUES` would have broken once `notes_is_ai_summary` landed.

One small non-blocking note: the controller refactor now calls `guardManageFismaSystem` only on the scoped-admin branch, so an owner PUT to a non-existent id skips the guard and falls through to Save instead of returning the guard's not-found. It still errors (the UPDATE matches no row), just by a different route - fine to leave as-is, just flagging it as a conscious choice.

Approving. Nice turnaround on this.

danielbowne:
Rehomed to #386 so CI and Snyk can run against an in-repo branch (fork PRs can't reach the org secrets Snyk needs). All commits are yours and authorship is preserved on the rehomed branch. Closing this in favor of #386 and leaving it here as the review and credit record. Thanks @costacalvin.

PR #383, opened 2026-07-06
Make the Emberfall e2e runner install OS-aware for macOS

### LINKED ISSUE #382: Make `make test-e2e` emberfall install OS-aware for macOS (Apple Silicon)
**Type:** Chore

## Description

`make test-e2e` (and `make test-integration`) do not bootstrap cleanly on macOS because the emberfall install hint the Makefile prints when the binary is missing is Linux-only:

```
curl -L https://github.com/aquia-inc/emberfall/releases/download/v0.4.1/emberfall_Linux_x86_64.tar.gz | tar -xz && mv emberfall ~/.local/bin/
```

All of our dev machines are Apple Silicon Macs, so this pulls a Linux binary that will not execute. A developer copy-pasting the hint gets an "exec format error" with no clear reason why. The correct asset for these machines is `emberfall_Darwin_arm64.tar.gz`.

## Proposed fix

Make the install step OS- and arch-aware. Map `uname -s` / `uname -m` to the matching emberfall release asset (`Darwin_arm64`, `Darwin_x86_64`, `Linux_x86_64`, and so on) and either print the correct curl line or install it automatically. While here, confirm the destination directory is on PATH on macOS: `~/.local/bin` is not on the default zsh PATH, so `/usr/local/bin` is the safer target.

## Acceptance criteria

- On an Apple Silicon Mac with Docker running, a fresh checkout can run `make test-e2e`, and when emberfall is missing the printed hint (or auto-install) produces a working binary.
- The existing Linux path keeps working unchanged.
- The same handling applies to `make test-integration`.

Surfaced while reviewing #380.

danielbowne:
The `test-e2e` target printed a hardcoded Linux install command for the Emberfall runner, so macOS (Apple Silicon) developers who followed it got a Linux binary that will not execute. This resolves the host platform and installs the matching release build.

## Changes

- Detect host OS and architecture (`uname -s` / `uname -m`) and build the correct `emberfall_<os>_<arch>.tar.gz` asset URL.
- Add a `make install-emberfall` target that downloads and installs the right build into a user-writable directory (`~/.local/bin` by default, override with `EMBERFALL_BIN_DIR`), and warns if that directory is not on `PATH`.
- Point the `test-e2e` "not installed" hint at `make install-emberfall`.

The isolated e2e flow itself is unchanged; only the install path was platform-specific.

Fixes #382

## Test plan

- [x] `make install-emberfall` on Linux installs a working `emberfall` (`emberfall version 0.4.1`)
- [x] OS/arch resolution: Linux to `Linux_x86_64`, Apple Silicon to `Darwin_arm64`, Intel Mac to `Darwin_x86_64` (all published release assets)
- [x] `make help` parses; existing `test-e2e` run path is unchanged when the runner is already installed
- [ ] `make install-emberfall` on an Apple Silicon Mac installs a working binary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ EMBERFALL_UNAME_M := $(shell uname -m)
 EMBERFALL_ARCH := $(if $(filter arm64 aarch64,$(EMBERFALL_UNAME_M)),arm64,$(if $(filter i386 i686,$(EMBERFALL_UNAME_M)),i386,x86_64))
 EMBERFALL_ASSET := emberfall_$(EMBERFALL_OS)_$(EMBERFALL_ARCH).tar.gz
 EMBERFALL_URL := https://github.com/aquia-inc/emberfall/releases/download/$(EMBERFALL_VERSION)/$(EMBERFALL_ASSET)
+# Checksums file is published per release as emberfall_<version-no-v>_checksums.txt.
+EMBERFALL_CHECKSUMS_URL := https://github.com/aquia-inc/emberfall/releases/download/$(EMBERFALL_VERSION)/emberfall_$(patsubst v%,%,$(EMBERFALL_VERSION))_checksums.txt
 # Default to a user-writable dir: /usr/local/bin is root-owned (and often absent
 # on Apple Silicon, where Homebrew lives under /opt/homebrew), so it fails the
 # install without sudo. Override for a system-wide install, e.g.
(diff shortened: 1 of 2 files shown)


MackOverflow (dismissed):
Looks clean, I like the $(HOME) expansion. Two nits, can be fixed or not, neither are blockers: 

1. Drop the redundant curl one-liner from the test-e2e error output (nit).
2. Consider a sha256sum verification step (low, but fits the repo's security posture).

Makefile
```diff
@@ -7,7 +7,24 @@ SHELL := /bin/bash
...
+EMBERFALL_OS := $(shell uname -s)
+EMBERFALL_UNAME_M := $(shell uname -m)
+EMBERFALL_ARCH := $(if $(filter arm64 aarch64,$(EMBERFALL_UNAME_M)),arm64,$(if $(filter i386 i686,$(EMBERFALL_UNAME_M)),i386,x86_64))
+EMBERFALL_ASSET := emberfall_$(EMBERFALL_OS)_$(EMBERFALL_ARCH).tar.gz
+EMBERFALL_URL := https://github.com/aquia-inc/emberfall/releases/download/$(EMBERFALL_VERSION)/$(EMBERFALL_ASSET)
+# Default to a user-writable dir: /usr/local/bin is root-owned (and often absent
+# on Apple Silicon, where Homebrew lives under /opt/homebrew), so it fails the
+# install without sudo. Override for a system-wide install, e.g.
+# `make install-emberfall EMBERFALL_BIN_DIR=/usr/local/bin`.
+EMBERFALL_BIN_DIR ?= $(HOME)/.local/bin
```
jsos3-cms, line 27: Not a blocker, genuine question here; why not parse for PATH instead here? Wouldn't that be slightly more inclusive, without any additional risk of typo squatting?

Makefile
```diff
@@ -301,11 +319,22 @@ test-coverage-text:
...
+	@echo "✅ emberfall $(EMBERFALL_VERSION) installed to $(EMBERFALL_BIN_DIR)/emberfall"
+	@case ":$$PATH:" in *":$(EMBERFALL_BIN_DIR):"*) ;; *) echo "⚠️  $(EMBERFALL_BIN_DIR) is not on your PATH. Add it: export PATH=\"$(EMBERFALL_BIN_DIR):$$PATH\"" ;; esac
+
 test-e2e:
 	@echo "🧪 Running Emberfall E2E tests (isolated environment)..."
 	@if ! command -v emberfall >/dev/null 2>&1; then \
 		echo "❌ Emberfall not installed"; \
-		echo "   Install: curl -L https://github.com/aquia-inc/emberfall/releases/download/v0.4.1/emberfall_Linux_x86_64.tar.gz | tar -xz && mv emberfall ~/.local/bin/"; \
+		echo "   Run: make install-emberfall"; \
+		echo "   (or: mkdir -p $(EMBERFALL_BIN_DIR) && curl -fsSL $(EMBERFALL_URL) | tar -xzf - -C $(EMBERFALL_BIN_DIR) emberfall)"; \
```
jsos3-cms: Non-blocking nit - do you think we should move these echo errors into a var and then do multi-line returns?

danielbowne:
Thanks both. Addressed in `592b773`:

- **@MackOverflow (sha256)**: the install now pulls the release checksums file and verifies the tarball before extracting (`sha256sum` on Linux, `shasum -a 256` on macOS); a hash mismatch fails the target.
- **@MackOverflow (redundant curl)**: dropped. The not-installed hint is a single line pointing at `make install-emberfall`.
- **@jsos3-cms (echoes into a var)**: that hint collapsed to one line with the change above, so the multi-line block is gone.
- **@jsos3-cms (parse PATH for the bin dir)**: I kept the fixed `~/.local/bin` default for predictability - a dev always knows where the binary landed, and auto-picking a writable dir off `PATH` could install somewhere surprising or hit a non-writable dir. The post-install check already prints an export hint when that dir isn't on `PATH`. Glad to switch to a `PATH` scan if you feel strongly.

Re-requesting review.

MackOverflow (dismissed):
clean!

jsos3-cms:
Heads up @danielbowne — I've folded the compose-test macOS networking fix (originally #385, rehomed as #388 by @voidspooks) onto this branch as `eb9c284`, with @voidspooks preserved as author and the commit signed/verified. This branch is now the complete two-part fix: the emberfall OS-aware installer plus compose-test port publishing. #387 and #388 are being closed as superseded. Once this has an approving review and the DEV deploy is green, it's ready to squash-merge.

jsos3-cms (approved):
Submitting approval, but CI should be green

PR #385, opened 2026-07-06
Publish compose-test ports instead of host networking so local e2e/integration work on macOS

### LINKED ISSUE #384: `make test-e2e` / `make test-integration` can't connect on macOS: compose-test uses `network_mode: host`
**Type:** Bug

## Description

`make test-e2e` and `make test-integration` cannot complete on macOS (Apple Silicon) because the isolated test stack in `backend/compose-test.yml` relies on Docker `network_mode: host`. Both the `test-db` and `test-api` services set `network_mode: host  # Use host network like CI`. Host networking bridges container ports directly onto the host only on native Linux (which is what CI runs). On Docker Desktop for Mac, containers run inside a Linux VM, so `network_mode: host` attaches them to the VM's network, **not** the macOS host — the container's ports are never exposed to the Mac. As a result:

- `make test-e2e`: `test-api` comes up healthy and logs `test environment listening on 8090`, but nothing on the host can reach `localhost:8090`. The (correctly installed) emberfall runner fires every request against `localhost:8090` and gets `connection refused`, so 0 tests run and all 118 report as failed.
- `make test-integration`: the host-side `go test` connects to the test DB at `localhost:54399`; that port is likewise unreachable from the Mac host.

This is a follow-up to #382 / PR #383, which fixed the *installer* (`make install-emberfall` now installs the correct native `Darwin_arm64` binary). That fix is necessary but not sufficient: the runner installs and runs natively, but the suite it drives still can't connect on macOS because of the compose host-networking assumption. The stated goal of #382 — "a fresh checkout can run `make test-e2e`" on an Apple Silicon Mac — is not met until this is resolved.

## Steps to Reproduce

1. On an Apple Silicon Mac with Docker Desktop running, from a fresh checkout: `make install-emberfall` (installs a working `emberfall version 0.4.1`).
2. Run `make test-e2e`.
3. Observe the `test-api` container start and log `test environment listening on 8090`.
4. Observe every emberfall request fail with `connection refused` on `localhost:8090`.

(Equivalently: run `make test-integration` and observe the host `go test` fail to reach the test DB on `localhost:54399`.)

## Expected Behavior

On an Apple Silicon Mac with Docker running, `make test-e2e` brings up the isolated stack, the host-side emberfall runner reaches the API, the suite executes, and results reflect real test outcomes. The same isolated stack keeps working unchanged on Linux/CI.

## Actual Behavior

The `test-api` container is `Up` and listening on 8090 inside the VM, but its `PORTS` column is empty and the port is not published to the macOS host. `curl http://127.0.0.1:8090/...` returns nothing (exit/HTTP `000`), and emberfall reports:

```
Reading config from /tmp/emberfall_test_isolated.yml
Get "http://localhost:8090/api/v1/users/current": dial tcp [::1]:8090: connect: connection refused
... (repeated for every endpoint) ...

    Ran: 0
 Failed: 118
Skipped: 0
Error: tests failed
❌ Emberfall tests failed
```

## Environment

- macOS 15.3.1 (Apple Silicon / arm64), Docker Desktop with Compose v2, `network_mode: host` (no experimental host-networking feature enabled).
- Reproduced against PR #383 (`chore/emberfall-install-os-aware`); `emberfall` verified as a native `Mach-O arm64` binary reporting `emberfall version 0.4.1`.
- Unaffected on Linux/CI, where host networking exposes container ports on the host directly.

## Proposed Fix

Stop relying on host networking for the Mac path. Put `test-db` and `test-api` on a compose-managed network and publish the ports the host tooling needs:

- `test-db`: `ports: ["54399:54399"]` (already listens on `-p 54399`), reachable from host `go test`.
- `test-api`: `ports: ["8090:8090"]`, reachable from host emberfall.
- Rewire the api→db address off host loopback: `test-api` currently connects to `DB_ENDPOINT: localhost` / `DB_PORT: 54399`, which only works under host networking. On a bridge network it should reach the DB by service name (`DB_ENDPOINT: test-db`, port `54399`).

Keep the behavior identical on Linux/CI (published ports work there too, so the `network_mode: host` "like CI" note can likely be dropped). Confirm the `test-e2e` port-rewrite step (`sed 's/localhost:8080/localhost:8090/g'`) still targets the published host port.

## Acceptance Criteria

- On an Apple Silicon Mac with Docker running, a fresh checkout can `make install-emberfall` then `make test-e2e`, and the suite actually executes (non-zero `Ran:` count reflecting real pass/fail) rather than failing on `connection refused`.
- `make test-integration` connects to the isolated test DB and runs on macOS.
- The isolated stack continues to work unchanged on Linux/CI.

## Affected Component

`backend/compose-test.yml` (`test-db` and `test-api` services, `network_mode: host`; `test-api` DB endpoint wiring). Consumed by the `test-e2e` and `test-integration` targets in `Makefile`.

Surfaced while testing PR #383 on an Apple Silicon Mac.

voidspooks:
The isolated test stack in `backend/compose-test.yml` ran both `test-db` and `test-api` with `network_mode: host`. Host networking bridges container ports onto the host only on native Linux (CI). On Docker Desktop for Mac the containers attach to the Docker VM's network and their ports are never exposed to the macOS host, so both local test targets silently broke on Apple Silicon:

- `make test-e2e`: the host-side emberfall runner could not reach the API at `localhost:8090` — 0 tests ran and every request failed with `connection refused`.
- `make test-integration`: the host-side `go test` could not reach the test DB at `localhost:54399`.

This is the follow-up to #383, which fixed the emberfall *installer*: the runner now installs and runs as a native `Darwin_arm64` binary, but the suite it drives still couldn't connect until this networking issue was resolved.

## Changes

- Drop `network_mode: host` from `test-db` and `test-api`; run them on Compose's default bridge network with published ports (`54399:54399`, `8090:8090`) so the host-side runners reach them on every platform.
- Point `test-api` at the DB by service name (`DB_ENDPOINT: test-db`) instead of `localhost`, since the two services no longer share host loopback.
- Rewrite the header comment (the old "matches GitHub Actions behavior / host network like CI" note was also inaccurate — CI uses `actions-compose.yml` + the emberfall action, not this file).

Published ports behave identically on Linux, and CI is unaffected because it does not use `compose-test.yml`. The seed flow, fresh-DB-per-run, healthcheck, and port numbers are all unchanged.

## Test plan

Verified on an Apple Silicon Mac (macOS 15.3.1, Docker Desktop):

- [x] `make test-e2e` runs the full suite — **Ran: 118, Failed: 0** (was 0 run / 118 failed on `connection refused` before the change)
- [x] Integration tests pass against the seeded test DB reachable at `localhost:54399`
- [x] `docker compose -f compose-test.yml config` validates; host can reach both `localhost:8090` and `localhost:54399` with the stack up
- [x] CI path unchanged — `compose-test.yml` is not referenced by any workflow

Fixes #384

### DIFF
--- a/backend/compose-test.yml
+++ b/backend/compose-test.yml
@@ -1,5 +1,11 @@
-# Isolated test environment - matches GitHub Actions behavior
-# Uses host networking like CI, fresh database each run, no volumes
+# Isolated test environment - fresh database each run, no volumes.
+# Uses a bridge network with published ports (not host networking) so the
+# host-side test runners work on every platform: emberfall (make test-e2e)
+# reaches the API at localhost:8090 and `go test` (make test-integration)
+# reaches the DB at localhost:54399. Docker Desktop on macOS does not expose
+# `network_mode: host` container ports to the host, so host networking silently
+# broke both targets on Apple Silicon; published ports behave identically on
+# Linux/CI. See CMS-Enterprise/ztmf#384.
 
 services:
   test-db:
@@ -8,7 +14,8 @@ services:
       POSTGRES_DB: ztmf
       POSTGRES_USER: admin
       POSTGRES_PASSWORD: testpass
-    network_mode: host  # Use host network like CI
+    ports:
+      - "54399:54399"
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U admin -d ztmf -p 54399"]
       interval: 2s
@@ -23,7 +30,7 @@ services:
       dockerfile: Dockerfile
     environment:
       PORT: 8090
-      DB_ENDPOINT: localhost
+      DB_ENDPOINT: test-db  # reach the DB by service name over the compose network
       DB_PORT: 54399
       DB_NAME: ztmf
       DB_USER: admin
@@ -32,7 +39,8 @@ services:
       AUTH_HS256_SECRET: zeroTrust
       AUTH_HEADER_FIELD: Authorization
       ENVIRONMENT: test
-    network_mode: host  # Use host network like CI
+    ports:
+      - "8090:8090"
     volumes:
       - ./_test_data_empire.sql:/app/_test_data_empire.sql:ro
     depends_on:


jsos3-cms:
@danielbowne @voidspooks - I wonder, is it possible to retarget this to the branch in 383 rather than main? Might be able to inherit import and approval for the wf.

MackOverflow:
Moving this to #387 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @voidspooks!

danielbowne (approved):
Nice - this is the other half of the macOS local-test story. #383 got the emberfall runner installing and running as a native binary; this fixes the stack it drives.

The root cause is right: `network_mode: host` only maps container ports to the host on native Linux, so on Docker Desktop for Mac the host-side runners (emberfall to `:8090`, `go test` to `:54399`) had nothing to connect to. Moving to a bridge network with published ports fixes it, and pointing `test-api` at the DB by service name (`DB_ENDPOINT: test-db`) is the necessary companion once they stop sharing host loopback.

Verified two things:
- CI is genuinely unaffected: `compose-test.yml` is referenced by no workflow (CI uses `backend.yml`), so removing the inaccurate "matches CI" comment is the right call too.
- No Linux regression: ran `make test-e2e` on Linux against this branch, 118 ran and 0 failed. Published ports behave the same as host networking there.

Small, focused, correct. Approving.

danielbowne:
Rehomed to #388 so CI and Snyk can run against an in-repo branch (fork PRs can't reach the org secrets Snyk needs). Your commit and authorship are preserved on the rehomed branch. Continuing from #388; leaving this as the review and credit record. Thanks @voidspooks.

PR #394, opened 2026-07-07
Map data-center environments to a shared scoring vocabulary (#392)

### LINKED ISSUE #392: HHS systems score 0.00: normalize data-center environments to a shared scoring vocabulary (mapping table)
## Problem
The scoring engine matches a system to its ZT functions by `datacenterenvironment` (`functions.datacenterenvironment = fismasystems.datacenterenvironment`, `scores.go:685`). The function catalog is keyed to a fixed vocabulary. HHS inventory values (`Gov-Owned, Multi-tenant`, `AWS (incl GovCloud)`, etc.) are not in it, so the join returns no functions and those systems score **0.00** - about 69% of HHS systems (762 of 1,108). Confirmed against a full local load. Same root mechanism as #367 (answers keyed to data center); this is the bulk / vocabulary-mismatch angle.

## Approach (Option B - agreed by team)
- Existing `fismasystems.datacenterenvironment` is **left untouched** - every system keeps its raw value for its own reporting; nothing lost.
- A **mapping table** turns the raw value into a scoring **category** the app scores/reports on. All agency-specifics (raw names, mappings) live in **data, not code**, so another OpDiv/agency just adds rows.

## Category set - DECIDED by Elizabeth (business owner), 2026-07-07
Hybrid: some CMS-specific kept, the rest generic. Categories and their question set:

| category | question set |
| --- | --- |
| `CMS-Cloud-AWS` | existing |
| `CMS-Cloud-MAG` | existing |
| `CMSDC` | existing |
| `AWS` (aka `Cloud-AWS`) | existing AWS |
| `MAG` (aka `Cloud-MAG`) | generalize CMS-MAG: keep the MAG parts, drop CMS-specificity |
| `SaaS` | existing (question list from Martha) |
| `Other` | existing |
| `data-center-gov` | **use the OPDC question set** |
| `data-center-contractor` | **rename OPDC to this**; uses OPDC's questions |

Key insight from Elizabeth: **OPDC = "Other People's Data Center" = effectively contractor-owned, single-tenant.** So OPDC's questions are the base for both new `data-center-*` categories. Naming is type-first to match the cloud ones. These are the HHS ScoreCard questions.

### Raw value -> category mapping
| raw value | category |
| --- | --- |
| CMS-Cloud-AWS, Cloud-CMS-AWS | CMS-Cloud-AWS |
| CMS-Cloud-MAG, Cloud-CMS-Azure | CMS-Cloud-MAG |
| Azure (Commercial or MAG) | MAG |
| AWS, AWS (incl GovCloud) | AWS |
| CMSDC | CMSDC |
| OPDC | data-center-contractor |
| Contractor-Owned, Single-tenant | data-center-contractor |
| Gov-Owned, Multi-tenant | data-center-gov |
| SaaS | SaaS |
| Other | Other |

(`DECOMMISSIONED` excluded - legacy marker, see follow-up.)

## Implementation notes (to complete the work)
1. **Mapping**: add a lookup table `raw datacenterenvironment -> category`, seeded with the rows above. Agency-neutral - a CMS deployment's mappings are data, not code.
2. **Function/question sets**: rename OPDC functions -> `data-center-contractor`; create `data-center-gov` from the OPDC question set (point at or copy); generalize `CMS-Cloud-MAG` into a de-CMS'd `MAG`. Confirm AWS/SaaS/Other/CMS-Cloud-AWS/CMSDC keep their existing sets.
3. **Scoring join** (`scores.go:685`): join `functions` via the resolved **category** (through the mapping), not the raw `datacenterenvironment`.
4. **Load/seed**: HHS load writes raw `datacenterenvironment` as-is; the mapping resolves the category. No agency literals in code.
5. **Frontend**: raw value shown for reporting, category used for scoring (coordinate with #449).

## Open sub-decisions (from Elizabeth, non-blocking)
- Gov/contractor naming: `data-center-gov` + `data-center-contractor`, OR collapse to one `data-center`? She leaned to the two type-first names.
- She offered an inheritance diagram / working session if useful.

## Acceptance
- Raw `datacenterenvironment` preserved untouched for reporting.
- Mapping table drives the scoring join via category; nothing agency-specific hardcoded.
- HHS systems score correctly (0.00 cleared for the mapped categories).
- Each category has (or points at) a question set; OPDC renamed, `data-center-gov` seeded from OPDC questions.

## Related follow-ups (separate, not blocked)
- Data-call handling (sort by deadline, current/latest chip, consolidate year-paired calls, drop fake FY2026): tracked in #393.
- Legacy `DECOMMISSIONED` data center: 7 systems carry `datacenterenvironment = 'DECOMMISSIONED'`, left over from pre-endpoint decommissioning. All 7 are already `decommissioned = true`. Clean up their `datacenterenvironment`: restore the original from the events audit trail where present (6 of 7 have it) or null it otherwise.

danielbowne:
## What and why

HHS inventory systems were scoring 0.00. The scoring engine matches a system to its Zero Trust functions by an exact `datacenterenvironment` string (`functions.datacenterenvironment = fismasystems.datacenterenvironment`). The functions catalog is keyed to a small fixed vocabulary, but real inventory values (for example `Data Center: Gov-Owned, Multi-tenant`, `AWS (incl GovCloud)`) are free text and absent from it, so the join returned no functions and roughly 69% of HHS systems scored 0.00.

Closes #392.

## Approach

Option B from the issue: leave each system's raw `datacenterenvironment` untouched (every OpDiv keeps its own reporting label) and add a reference table that maps the raw value to the scoring vocabulary.

New table `public.datacenterenvironments`:

- `datacenterenvironment` (PK) - the raw value stored on a system
- `category` - the reporting / dropdown bucket
- `scoring_key` - the `functions.datacenterenvironment` set to score against (NULL = not scored, e.g. the legacy DECOMMISSIONED marker)
- `selectable` - offered in the new/edit-system dropdown
- `ordr` - dropdown order

The scoring aggregate and the answer-form query both resolve a system's raw value through this table via `scoring_key` instead of matching functions directly, so a value that is not itself a catalog key still scores.

Every scored environment is a first-class function set with its own key, so `scoring_key` is an identity mapping today. The migration:

- renames the cryptic `OPDC` ("Other People's Data Center") set to `data-center-contractor` - a pure key rename, so its functionids and every score referencing them are unchanged.
- adds `MAG` (a copy of the CMS Azure/MAG set) and `data-center-gov` (a copy of the OPDC set) as first-class sets, so those environments score independently and can diverge later by editing data - de-CMS the MAG wording, drop SaaS's inapplicable questions, gov-vs-contractor tweaks - with no further migration.

`scoring_key` stays a distinct column so a future divergence can still point several environments at one set by changing data, not code.

## Category set

Matches the business owner's decision: keep `CMS-Cloud-AWS`, `CMS-Cloud-MAG`, `CMSDC`; generic `AWS`, `MAG`, `SaaS`, `Other`; and `data-center-gov` + `data-center-contractor` (both built on the former OPDC/HHS ScoreCard question base today).

## Changes

- `migrations/0045datacenterenvironments.go` - create + seed the table, rename OPDC, add the MAG and data-center-gov sets
- `scores.go`, `questions.go` - route both env joins through `scoring_key`
- `controller/datacenterenvironments.go` + router - `GET /api/v1/datacenterenvironments` so the UI environment dropdown is backend-driven (`selectable_only`)
- `datacenterenvironments.go` model + DB-backed validation in `fismasystems.Save` / `functions.Save`
- `validations.go` - removed the hardcoded vocabulary map
- `openapi.yaml` regenerated

## Existing data (companion cleanup, not in this PR)

This PR is the schema + code. On an environment that already has scores, the schema change alone is not enough: measured on a full local CMS+HHS dataset, ~78% of score rows reference a different env's functionids than the system now scores against (a pre-existing scatter, present in CMS data too), so they would render as "Not Assessed" until realigned. A separate, idempotent re-point script moves each such score to the matching option in its system's own set (same function, same score - lossless, because the option sets are byte-identical across envs). It is run manually per environment after a snapshot, matching the HHS load's prod discipline.

Validated end to end on real local data:
- re-point: 120,935 rows realigned, 0 ambiguous, 0 unmatchable
- HCAS (an HHS system that showed 0.00) now returns 3.47 "Advanced" with a full pillar breakdown through the live API
- 988 HHS systems in one data call now spread across all four tiers (Traditional through Optimal) instead of collapsing to Not Assessed

## Testing

- `make test-unit`, `make test-integration`, `make test-e2e` all pass (emberfall 132/132).
- Integration test proves an aliased environment (a raw value that is not a functions key) scores identically to its real environment through the mapping.
- The all-aggregate invariant test iterates every seeded system through the rewritten join, confirming no scoring regression.

## Follow-ups (tracked separately)

- The 7 new FY25 HHS questions need functions authored in each scoring env before they score.
- `DECOMMISSIONED` (7 systems) is a legacy marker, not an environment - small env-value cleanup.
- Frontend (ztmf-ui#449): backend-driven dropdown + re-key pillar filtering to the category.

## Also in this PR: ISSO name for the systems table

The main systems table read `issoemail` for its ISSO column, so HHS systems (which carry `isso_name`, not email) rendered blank. The list query now returns `COALESCE(isso_name, the ISSO user's fullname resolved by issoemail)`, giving one populated name column for both CMS and HHS in a single query with no per-row lookup. Read-only in the list; the single-system GET and the write path are unchanged, so no resolved value is persisted back.

Follow-up ztmf#395 covers turning the other HHS system attributes (FIPS, system type, cloud fields, etc.) into canonical validated select boxes; they ship as free-text fields for now.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -360,6 +360,20 @@ INSERT INTO public.questions VALUES (8016, 'How does your system authenticate an
 INSERT INTO public.questions VALUES (8017, 'What measures detect Force-sensitive individuals or Rebel infiltrators attempting system access?', 'Detail identity verification processes, behavioral analysis, and security screening protocols to identify potential security threats among personnel.', 6, 0) ON CONFLICT DO NOTHING;
 INSERT INTO public.questions VALUES (8018, 'How does your system manage Imperial identity lifecycle from recruitment to retirement?', 'Describe automated identity provisioning, access reviews, and deprovisioning processes for Imperial officers, contractors, and service accounts.', 6, 0) ON CONFLICT DO NOTHING;
 
+-- datacenterenvironments mapping rows for the Imperial (test) environments
+-- (ztmf#392). The migration seeds the real CMS/HHS vocabulary; these identity
+-- rows register the Star Wars environments so the scoring join and the answer
+-- form (both route fismasystems.datacenterenvironment -> scoring_key) resolve.
+-- Not selectable, so the /datacenterenvironments dropdown stays the canonical set.
+INSERT INTO public.datacenterenvironments (datacenterenvironment, category, scoring_key, selectable, ordr) VALUES
+    ('Imperial-Fleet',   'Imperial-Fleet',   'Imperial-Fleet',   FALSE, 0),
+    ('Space-Station',    'Space-Station',    'Space-Station',    FALSE, 0),
+    ('Forest-Moon',      'Forest-Moon',      'Forest-Moon',      FALSE, 0),
+    ('Ground-Assault',   'Ground-Assault',   'Ground-Assault',   FALSE, 0),
+    ('Surveillance-Net', 'Surveillance-Net', 'Surveillance-Net', FALSE, 0),
+    ('Shipyard-RnD',     'Shipyard-RnD',     'Shipyard-RnD',     FALSE, 0)
+ON CONFLICT DO NOTHING;
+
 -- Sample Functions (Imperial Zero Trust Functions) - MUST come before functionoptions
 -- Each datacenterenvironment needs functions with questionid set so FindQuestionsByFismaSystem works
 -- (INNER JOIN functions ON functions.questionid=questions.questionid)
--- a/backend/cmd/api/internal/controller/datacenterenvironments.go
+++ b/backend/cmd/api/internal/controller/datacenterenvironments.go
@@ -1,0 +1,35 @@
+package controller
+
+import (
+	"log"
+	"net/http"
+
+	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+)
+
+// ListDataCenterEnvironments returns the datacenterenvironment reference list.
+// Open to any authenticated user because it carries no sensitive data and the
+// frontend needs it to build the system-environment dropdown from the backend
+// (ztmf#392) instead of a hardcoded list. Pass selectable_only=true to get only
+// the values offered for new/edited systems.
+//
+//	@Summary	List datacenterenvironment mappings
+//	@Tags		datacenterenvironments
+//	@Produce	json
+//	@Security	bearerAuth
+//	@Param		selectable_only	query		bool	false	"Only environments offered in the system dropdown"
+//	@Success	200				{object}	apiResponse[[]model.DataCenterEnvironment]
+//	@Failure	500				{object}	apiResponse[any]
+//	@Router		/datacenterenvironments [get]
+func ListDataCenterEnvironments(w http.ResponseWriter, r *http.Request) {
+	input := model.FindDataCenterEnvironmentsInput{}
+
+	if err := decoder.Decode(&input, r.URL.Query()); err != nil {
+		log.Println(err)
+		respond(w, r, nil, ErrMalformed)
+		return
+	}
+
+	envs, err := model.FindDataCenterEnvironments(r.Context(), input)
+	respond(w, r, envs, err)
+}
--- a/backend/cmd/api/internal/migrations/0045datacenterenvironments.go
+++ b/backend/cmd/api/internal/migrations/0045datacenterenvironments.go
@@ -1,0 +1,128 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		"create datacenterenvironments mapping table, seed it, rename OPDC, add MAG and data-center-gov function sets",
+		`
+-- ZTMF scores a system by matching its datacenterenvironment against the
+-- functions catalog (functions.datacenterenvironment). That catalog is keyed to
+-- a small, fixed vocabulary. Real inventory values (HHS ScoreCard exports and
+-- other OpDivs) use free-text environment names that are not in it, so the
+-- scoring join returned no functions and those systems scored 0.00 (ztmf#392).
+--
+-- This reference table decouples the value stored on a system (kept untouched so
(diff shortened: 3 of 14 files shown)


MackOverflow (approved):
Super strong backend fix. went short on review, will need a full stitching with SDL and Docker in the AM. Reviewed the mapping design and both rewritten joins (scores + answer-form) — the reference-table indirection is sound and I confirmed neither join fans out rows. Unit suite green locally. Approving the code. Before deploy: confirm the MAG/data-center-gov function copies match source counts, that no unseeded functions.datacenterenvironment values exist in the target env, and that the re-point script runs after snapshot+migrate per the runbook.

danielbowne:
Thanks Mack. Ran your three pre-deploy checks against the applied migration on the local dev-scale data:

1. **Copies match source exactly** - MAG = CMS-Cloud-MAG (48 functions / 192 options), data-center-gov = data-center-contractor (48 / 192).
2. **No unseeded envs** - 0 orphan function sets, 0 scoring_keys pointing at a missing function env, 0 `fismasystems` env values without a mapping row.
3. **Ordering** is in the runbook now: snapshot -> migrate (through 0045) -> run these checks -> seed -> re-point -> verify. The re-point script has a fan-out guard that aborts before writing if any score is ambiguous, and prints `misaligned_before`/`misaligned_after` (after must be 0). It's idempotent.

One nuance I made an explicit step: the "unmapped env" check is data-dependent, so it re-runs AFTER the seed too - a refreshed HHS package could introduce a raw env value we haven't mapped yet, which just needs a new row in `datacenterenvironments` before the re-point.

Issue #393, opened 2026-07-07
Data call handling: sort by deadline, deadline-driven current/latest chip, consolidate year-paired calls


danielbowne:
## Summary
How ZTMF picks the "current"/default data call, and how the year-paired historical calls get consolidated. Surfaced during the HHS historical load. Not blocked on the scoring-vocabulary decision (#392).

## Background
The app treats "current" as the highest `datacallid` - backend `FindLatestDataCall` does `ORDER BY datacallid DESC`, and the frontend (`Title.tsx`) sorts datacalls by id descending. Once historical HHS calls are loaded, that breaks: a re-imported historical year can out-id the real current call. Going forward it is a single annual call every OpDiv answers at the same time (CMS is just an OpDiv under HHS), so "current" is the same for everyone.

## Scope
1. **Sort by `deadline`, not `datacallid`** - in both backend `FindLatestDataCall` and the frontend datacall sort.
2. **Deadline-driven chip** - show "current" only while a call is open (deadline not passed); flip to "latest" once closed.
3. **Consolidate the year-paired calls** into one annual call each: 2023 = `FY2023 Q4` + `FY23 ZTM`; 2025 = `FY2025 Q3` + `FY25 ZTM`; 2024 = `FY24 ZTM`. The score data is structurally identical (same functions/questions/options), so reassign scores + junction rows to the surviving call, delete the redundant one, rename to the agreed annual name.
4. **Drop the fake `FY2026`** test data call.

## Delivery
- Consolidation (3 + 4) as an **idempotent, transactional post-load reconciliation script**, run manually per environment (dev, verify, then prod), versioned in the repo. Not a migration (environment-specific data surgery) and not baked into the seed (one-time historical).
- Sort + chip (1 + 2) are normal code changes (backend + ztmf-ui).

## Notes
- No per-OpDiv "current" for now - all OpDivs share one annual cadence, so it would be complexity with no use case. Good pattern if cadences ever split, but YAGNI.
- Exact merged call names are pending the annual-call naming decision.
- Related: #392 (scoring vocabulary, blocked on ES).

## Acceptance
- Current/default resolves by deadline and is correct with historical calls present.
- Chip reads "current" only for an open call, "latest" otherwise.
- One annual call per year after consolidation; no fake FY2026; nothing hardcoded that would break another deployment.

voidspooks:
Working this. Thanks for the write up @danielbowne

jsos3-cms:
[...]

PR #396, opened 2026-07-08
feat(datacalls): resolve current/latest by deadline, not datacallid (#393)

voidspooks:
Part of #393 (the backend half; paired with the ztmf-ui PR for the picker/chip). The data-consolidation script (items 3-4 of the ticket) is a separate follow-up.

## Problem
"Current"/"latest" resolved by `datacallid`: `FindLatestDataCall` and the `FindDataCalls` list ordered by `datacallid DESC`. A re-imported historical data call can be assigned a higher serial id than the real current call, so once the HHS historical load lands it would hijack the default/current view.

## Changes
- `FindLatestDataCall`: order by `deadline DESC` (`datacallid DESC` only as a tiebreak). This also fixes `GET /api/v1/datacalls/latest`.
- `FindDataCalls`: same ordering, so the list the UI renders already agrees with "latest".
- `findPreviousDataCall`: same ordering, so `copyPreviousScores`' rollover points at the true prior annual call rather than a high-id backfill. (Scope-adjacent, but the same bug in the rollover path — called out in the ticket discussion.)
- Integration test asserting a higher-`datacallid`/earlier-`deadline` call is **not** treated as latest, and updated the `copyPreviousScores` fixtures (which leaned on id ordering) to pin deadline ordering.

No schema change — `deadline` already exists on `datacalls`.

## Testing
- `go build ./...`, `go vet`, `go test -short ./...` green (in the pinned `golang:1.25.11` container; host has no Go).
- Full `internal/model` integration suite green against a real Postgres, including the new `TestFindLatestDataCallByDeadlineIntegration`.
- Rebuilt the dev API from this branch and verified end-to-end: injected a high-`datacallid` call with a past deadline; `GET /datacalls/latest` still returns the open furthest-out call, and `GET /datacalls` sorts the injected call into its deadline position rather than the top.

Refs #393

### DIFF
--- a/backend/internal/model/datacalls.go
+++ b/backend/internal/model/datacalls.go
@@ -53,7 +53,7 @@ func (d *DataCall) Save(ctx context.Context) (*DataCall, error) {
 func FindDataCalls(ctx context.Context) ([]*DataCall, error) {
 	sqlb := stmntBuilder.Select(dataCallColumns...).
 		From("datacalls").
-		OrderBy("datecreated DESC")
+		OrderBy("deadline DESC", "datacallid DESC")
 
 	return query(ctx, sqlb, pgx.RowToAddrOfStructByName[DataCall])
 }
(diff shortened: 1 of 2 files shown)


voidspooks:
Paired frontend PR: CMS-Enterprise/ztmf-ui#461 (picker sort + deadline-driven Current/Latest chip). Both are the code half of #393; the data-consolidation script (items 3-4) is a separate follow-up.

jsos3-cms:
Rehomed in-repo as #397 so the org-secret CI gates (Snyk, Infrastructure/Deploy) can run — fork PRs don't receive those secrets, so nothing ran here. Your commit was cherry-picked unchanged with authorship preserved, and is signed/verified on #397. Thanks @voidspooks — closing in favor of #397.

jsos3-cms:
Reopening and moving this to **draft** rather than leaving it closed — the convention is to keep the original fork PR open (as draft) until its in-repo rehome (#397) is merged, then close it. Apologies for closing prematurely.

Your follow-up commit `4aa72aa` (e2e: update emberfall `datacalls/latest` for deadline-based current) has also been rehomed onto #397, signed with authorship preserved. Both your commits now ride on #397; this PR stays open (draft) to track it until #397 lands. Thanks @voidspooks.

jsos3-cms:
Correction to my previous note: I can't actually reopen this. GitHub blocks it — `state cannot be changed: the feat/393-datacall-sort-by-deadline branch was force-pushed or recreated` — because the fork branch was rebased after this PR was closed, so the original head is unreachable. It therefore stays closed.

The live PR is **#397**, which now carries both commits (the deadline change and `4aa72aa`, the e2e update), signed with authorship preserved. Going forward we'll keep the original open as a draft until the rehome merges rather than closing it, to avoid exactly this. Thanks @voidspooks.

PR #397, opened 2026-07-08
feat(datacalls): resolve current/latest by deadline, not datacallid

jsos3-cms:
In-repo rehome of #396 by @voidspooks (fork PR), moved so the org-secret CI gates (Snyk, Infrastructure/Deploy) can run — they can't on a fork PR. Commit pushed unchanged; authorship preserved.

Part of #393 (the backend half; paired with the ztmf-ui PR for the picker/chip). The data-consolidation script (items 3-4 of the ticket) is a separate follow-up.

## Problem
"Current"/"latest" resolved by `datacallid`: `FindLatestDataCall` and the `FindDataCalls` list ordered by creation/id. A re-imported historical data call can be assigned a higher serial id than the real current call, so once the HHS historical load lands it would hijack the default/current view.

## Changes
- `FindLatestDataCall`: order by `deadline DESC` (`datacallid DESC` only as a tiebreak). This also fixes `GET /api/v1/datacalls/latest`.
- `FindDataCalls`: same ordering, so the list the UI renders already agrees with "latest".
- `findPreviousDataCall`: same ordering, so `copyPreviousScores`' rollover points at the true prior annual call rather than a high-id backfill. (Scope-adjacent, but the same bug in the rollover path — called out in the ticket discussion.)
- Integration test asserting a higher-`datacallid`/earlier-`deadline` call is **not** treated as latest, and updated the `copyPreviousScores` fixtures (which leaned on id ordering) to pin deadline ordering.

No schema change — `deadline` already exists on `datacalls` (and is `NOT NULL`, so `ORDER BY deadline DESC` has no NULL-ordering hazard).

## Testing
- `go build ./...`, `go vet`, `go test -short ./...` green (in the pinned `golang:1.25.11` container; host has no Go).
- Full `internal/model` integration suite green against a real Postgres, including the new `TestFindLatestDataCallByDeadlineIntegration`. The deadline tie-break and the backfill-conflict scenario were independently re-verified against a live Postgres.

## Follow-up note for #393 items 3-4 (historical-load consolidation)
`findPreviousDataCall` now resolves to the global latest by deadline excluding the given id. The consolidation/historical-load script must therefore **not** insert backfilled rows via `DataCall.Save()`, which fires `copyPreviousScores` and would roll current scores backward into a historical row. Reassign scores/junction rows directly instead.

Refs #393. Original contribution by @voidspooks.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -22,14 +22,18 @@ opDivAdminHeaders: &opDivAdminHeaders
 opDivReadonlyHeaders: &opDivReadonlyHeaders
   authorization: "eyJhbGciOiJIUzI1NiJ9.eyJlbWFpbCI6Im9wZGl2LnJlYWRvbmx5QGVtcGlyZS50ZXN0In0.enezNI1lnuNFcXzcxZqm0CTalZfbOV40dolVjoFxnII"
 
-# Data definitions using YAML anchors
+# Data definitions using YAML anchors.
+# The deadline is intentionally far-future — beyond the empire seed's
+# furthest-out call (Audit Fields Smoke Cycle, 2099-12-31) — so this
+# test-created call is the one GET /datacalls/latest returns. As of #393,
+# "latest"/"current" is the max-deadline call, not the max datacallid.
 dataCallData: &dataCallData
-  datacall: "FY2025 Q1"
-  deadline: "2025-03-31T17:59:59Z"
+  datacall: "FY2100 ZTM"
+  deadline: "2100-03-31T17:59:59Z"
 
 updatedDataCallData: &updatedDataCallData
-  datacall: "FY2025 Q1"
-  deadline: "2025-03-31T17:59:59Z"
+  datacall: "FY2100 ZTM"
+  deadline: "2100-03-31T17:59:59Z"
 
 fismaSystemData: &fismaSystemData
   fismauid: "12345678-ABCD-4321-AFAB-123456789ABC"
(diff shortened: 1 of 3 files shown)


jsos3-cms:
@voidspooks - Can you please take a look at the emberfall tests and include an update in your branch? I'll rehome the changes.
```
FAIL : GET http://localhost:8080/api/v1/datacalls/latest
       expected body.json.data.datacall == FY2025 Q1 got Audit Fields Smoke Cycle
       expected body.json.data.deadline == 2025-03-31T17:59:59Z got 2099-12-31T23:59:59Z
       {"data":{"datacallid":5,"datacall":"Audit Fields Smoke Cycle","datecreated":"2026-01-01T00:00:00Z","deadline":"2099-12-31T23:59:59Z"}}
```

voidspooks:
Thanks @jsos3-cms — emberfall updated on my branch `voidspooks:feat/393-datacall-sort-by-deadline` (commit `4aa72aa`), ready to pull in.

The failing `/datacalls/latest` assertion was encoding the old max-`datacallid` behavior: the emberfall flow creates a data call and asserts it's "latest," which only holds if that call has the furthest-out deadline. I set the created call's deadline beyond the empire seed's furthest-out call (Audit Fields Smoke Cycle, 2099-12-31) so it's genuinely the latest under the new deadline ordering — the single-file change is in `backend/emberfall_tests.yml`.

Ran the full suite locally against latest `main` (rebased onto it): `make test-unit`, `make test-integration`, and `make test-e2e` all green — emberfall **132 ran / 0 failed**, including `GET /datacalls/latest`.

Paired frontend change is on `voidspooks/ztmf-ui:feat/393-datacall-sort-by-deadline` (picker sort + Current/Latest chip) if you're re-homing that one too.

MackOverflow (approved):
looks great Jono, no nits!

PR #400, opened 2026-07-08
feat(fismasystems): per-system target maturity tier + justification (#398 MVP)

voidspooks:
Backend half of #398 (MVP scope per the ticket discussion with @MackOverflow — system-level only; pillar/function targets, gap views, and audit-history UI stay tracked by the epic). Paired frontend PR on ztmf-ui.

## What

Each system gets a risk-based **target maturity tier** plus a **required one-sentence justification**, settable by the assigned ISSO/ISSM or an admin, stored per system, audited, and included in the export.

## Design decisions (from the review of the draft AC)

- **Tier stored by NAME** (`Initial`/`Advanced`/`Optimal`, CHECK-constrained), not a CISA stage integer. The app's internal score scale is 1–5 and `Tier()` maps 3.10–4.09 → Advanced, so a stored `3` invites an invalid raw-number comparison against `systemscore`. Storing the name makes tier-vs-tier the only expressible comparison; the UI renders the stage numerals as display labels ("3 — Advanced").
- **No backfill; NULL = "no target asserted."** Backfilling Advanced would fabricate a GAO-auditable assertion nobody made. The UI/export present the Advanced default explicitly labeled as such.
- **Justification required on every save** — the risk rationale is the deliverable.
- **Dedicated endpoint** `PUT /fismasystems/{id}/target-maturity` instead of widening the admin-only system PUT: the assigned ISSO/ISSM owns exactly this field pair. Gate mirrors the scores write path — read-only tiers 403, non-admins must be assigned, admin tiers OpDiv-scoped via `guardManageFismaSystem`. Writes are audited through the standard `events` hook.

## Changes

- Migration `0046`: `target_maturity_tier` (varchar, CHECK) + `target_maturity_justification` (varchar 1000), nullable.
- Model: fields flow through all reads; `SaveTargetMaturity` validates tier/justification and records the audit event.
- Controller/route + swag annotations; `openapi.yaml` regenerated.
- Export: **Target Maturity Level** / **Target Justification** columns; NULL renders `Advanced (default)`.
- Emberfall: 8 new cases (ISSO-assigned 200, unassigned ISSO 403, read-only tiers 403, out-of-scope OPDIV_ADMIN 403, bad tier 400, missing justification 400, read-through).

## Test plan

- [x] `make test-unit` — incl. new validation tests (run pre-DB, `-short`-safe)
- [x] `make test-integration` — write → read-back, trim behavior, live CHECK-constraint rejection
[...]

Issue #398, opened 2026-07-08
Record the Expected Maturity Level


eschwein-usds:
### Feature Type

General Feature

### Workflow Description

Each system should have an "expected maturity level" that we can compare answers to.  ISSOs should be able to set this at a System, Pillar, and Function level.

Note: this is a request from Evan Miller in response to a finding from a GAO audit.

### Acceptance Criteria

- [ ] Expected maturity level is recorded in the data base for each System, Pillars, and Functions
- [ ] Expected maturity level is displayed and modifiable in the UI
- [ ] Expected maturity level is included in the export

### Additional Context

The default expected maturity level is Advanced, and teams can move it up or down from there.

MackOverflow:
## MVP scope (per Evan Miller)

Feedback from Evan: a minimum viable version is a single per-system control to state the risk-based target maturity level appropriate for that system. The ISSO selects **2, 3, or 4** and writes one sentence about why. This is scoped to the **system level only** — Pillar and Function targets are deferred (see below).

### Workflow Description (MVP)
Each system gets one risk-based **target maturity level**. An ISSO selects a target of 2, 3, or 4 and enters a short (one-sentence) justification for why that level is appropriate for the system.

Values follow the CISA ZTMM 2.0 scale. Assuming the mapping 1=Traditional, 2=Initial, 3=Advanced, 4=Optimal, the selectable options are: **2 — Initial**, **3 — Advanced (default)**, **4 — Optimal**. (Please confirm the numeric-to-stage mapping and that Advanced remains the default.)

### Acceptance Criteria (MVP)
- [ ] `target_maturity_level` (integer, 2–4) stored per system in the database
- [ ] `target_maturity_justification` (short text) stored per system
- [ ] System detail view displays the current target level and justification
- [ ] ISSO can select the target level (2/3/4) and edit the justification in the UI
- [ ] Defaults to 3 (Advanced) when not yet set
- [ ] Target level + justification included in the system export

### UI Details
- Add a **Target Maturity Level** control (dropdown or radio group) to the system detail/settings view, with each option labeled by number and CISA stage name.
- Add a **Justification** field beneath it with helper text: "In one sentence, explain why this target level is appropriate for this system."
- Preselect the default (Advanced) for systems with no target set.
- Restrict edit access to ISSOs/admins per existing role permissions.

### Technical Notes
[...]

PR #401, opened 2026-07-08
feat(fismasystems): per-system target maturity tier + justification (#398 MVP)

MackOverflow:
> **Note:** In-repo copy of #400 by @voidspooks. Moved from the fork so Snyk
>
> Refs #400

---

Backend half of #398 (MVP scope per the ticket discussion with @MackOverflow — system-level only; pillar/function targets, gap views, and audit-history UI stay tracked by the epic). Paired frontend PR on ztmf-ui. MERGE THIS PR BEFORE THE REHOME OF 398

## What

Each system gets a risk-based **target maturity tier** plus a **required one-sentence justification**, settable by the assigned ISSO/ISSM or an admin, stored per system, audited, and included in the export.

## Design decisions (from the review of the draft AC)

- **Tier stored by NAME** (`Initial`/`Advanced`/`Optimal`, CHECK-constrained), not a CISA stage integer. The app's internal score scale is 1–5 and `Tier()` maps 3.10–4.09 → Advanced, so a stored `3` invites an invalid raw-number comparison against `systemscore`. Storing the name makes tier-vs-tier the only expressible comparison; the UI renders the stage numerals as display labels ("3 — Advanced").
- **No backfill; NULL = "no target asserted."** Backfilling Advanced would fabricate a GAO-auditable assertion nobody made. The UI/export present the Advanced default explicitly labeled as such.
- **Justification required on every save** — the risk rationale is the deliverable.
- **Dedicated endpoint** `PUT /fismasystems/{id}/target-maturity` instead of widening the admin-only system PUT: the assigned ISSO/ISSM owns exactly this field pair. Gate mirrors the scores write path — read-only tiers 403, non-admins must be assigned, admin tiers OpDiv-scoped via `guardManageFismaSystem`. Writes are audited through the standard `events` hook.

## Changes

- Migration `0046`: `target_maturity_tier` (varchar, CHECK) + `target_maturity_justification` (varchar 1000), nullable.
- Model: fields flow through all reads; `SaveTargetMaturity` validates tier/justification and records the audit event.
- Controller/route + swag annotations; `openapi.yaml` regenerated.
- Export: **Target Maturity Level** / **Target Justification** columns; NULL renders `Advanced (default)`.
- Emberfall: 8 new cases (ISSO-assigned 200, unassigned ISSO 403, read-only tiers 403, out-of-scope OPDIV_ADMIN 403, bad tier 400, missing justification 400, read-through).

## Test plan

- [x] `make test-unit` — incl. new validation tests (run pre-DB, `-short`-safe)
- [x] `make test-integration` — write → read-back, trim behavior, live CHECK-constraint rejection
- [x] `make test-e2e` — **141 ran / 0 failed** incl. the 8 new target-maturity cases
- [x] Full flow validated in the local stack (admin + ISSO roles, export, audit events)

## Screenshots

Screenshot 2026-07-08 at 4 53 24 PM
Screenshot 2026-07-08 at 4 54 41 PM

Refs #398

### DIFF
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -233,6 +233,38 @@ func TestSaveScore_ISSONotAssignedForbidden(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
 
+// --- SaveFismaSystemTargetMaturity (#398) ---
+
+func TestSaveFismaSystemTargetMaturity_ReadonlyAdminForbidden(t *testing.T) {
+	body := jsonBody(t, map[string]any{
+		"target_maturity_tier":          "Advanced",
+		"target_maturity_justification": "should never write",
+	})
+	r := httptest.NewRequest("PUT", "/api/v1/fismasystems/1/target-maturity", body)
+	r.Header.Set("Content-Type", "application/json")
+	r = mux.SetURLVars(r, map[string]string{"fismasystemid": "1"})
+	r = withUser(r, readonlyAdmin)
+	w := httptest.NewRecorder()
+
+	SaveFismaSystemTargetMaturity(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
+func TestSaveFismaSystemTargetMaturity_ISSONotAssignedForbidden(t *testing.T) {
+	body := jsonBody(t, map[string]any{
+		"target_maturity_tier":          "Advanced",
+		"target_maturity_justification": "should never write",
+	})
+	r := httptest.NewRequest("PUT", "/api/v1/fismasystems/999/target-maturity", body)
+	r.Header.Set("Content-Type", "application/json")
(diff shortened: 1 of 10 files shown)


jsos3-cms (dismissed):
Sent along some non-blocking nits. Will provide approval and review if it gets clear, but looks good so far!

MackOverflow:
Two good fixes pointed out by @jsos3-cms 


Fix 1 — Justification length check. It was counting bytes (len()) against a varchar(1000) column that Postgres counts in characters, so non-ASCII text (accented names, CJK, emoji) could be wrongly rejected under the real limit. Switched to utf8.RuneCountInString on the trimmed value, matching the existing pattern in scores.go.

Fix 2 — Missing auth tests. The new /target-maturity endpoint had the same read-only/assigned permission gate as SaveScore, SaveFismaSystem, etc., but no fast unit test guarding it — only the slower Docker-based e2e suite covered it. Added two httptest-based cases (readonly-admin forbidden, unassigned ISSO forbidden) so a future gate reorder fails fast in make test-unit.

jsos3-cms (approved):
Thank you for the fast turn around!

PR #404, opened 2026-07-09
Identifiable system questionnaire updates

tarratsco:
### Summary

Adds `GET /api/v1/scores/progress`, a per-system questionnaire-progress endpoint that lets OpDiv Admins (and ISSOs, for their own systems) see which systems have actually updated their questionnaires during a data call — the core of `ztmf#299`. This is the backend half; the dashboard column ships in a companion `ztmf-ui` PR: https://github.com/CMS-Enterprise/ztmf-ui/pull/480.

### The problem this solves

When a data call opens, the backend pre-populates every system's answers by copying the previous cycle's responses (`copyPreviousScores`). So on day one, every system already has a full set of `scores` rows — which means the obvious "count the answer rows" metric reads **100% for everyone** and answers the wrong question. It measures "was this ever answered," not "did anyone touch it this cycle."

### The mechanic

The pre-population copy **deliberately records no audit events**; genuine writes through the normal save path do. That asymmetry already draws the line we need:

| Row | Has a `scores` row? | Has an edit event this cycle? |
| --- | --- | --- |
| Carried over, untouched | yes | **no** |
| Edited this cycle | yes | **yes** |

So **"updated" = the answer has an edit event in this data call.** No schema change, no new tracking — the endpoint counts answers that have an event (via the same `events` lateral join `FindScores` already uses for `last_edited_at`) against the questions applicable to the system (via the `datacenterenvironments` scoring-vocabulary, so SaaS systems get their smaller denominator).

Response, per system: `questionsexpected`, `questionsupdated`, `lastupdatedat`, `updatedsincestart`.

### Scoping

Identical tier rules to `/scores/aggregate`, applied after query decode so a client cannot widen scope via params: OWNER and HHS admins see all systems, while OpDiv admins fail closed to their granted OpDiv systems. Decommissioned systems are excluded from the response.

### File changes

| Component | Files |
| --- | --- |
| Database model | `backend/internal/model/scoreprogress.go `, `scoreprogress_test.go`, `scoreprogress_integration_test.go` |
| Handler + route | `cmd/api/internal/controller/scores.go`, `router/router.go` |
| Controller tests | `cmd/api/internal/controller/authorization_test.go` |
| Spec | `openapi.yaml` (regenerated) |

6 commits, +683 / 7 files. No migration, no schema change.

### Notes for reviewers

* Additive only, meaning no existing endpoint or write path is touched.
* `updatedsincestart` is derivable from `questionsupdated > 0`. It is kept because it names the ticket's literal question in the response and the "since start" anchor is real, since an event can only postdate the data call's creation.
* Deferred to follow-ups: a downloadable progress report with contacts (can ride the existing `/datacalls/{id}/export` CSV) and "viewed but not answered" tracking (which needs a new event type and NIH product input).

### DIFF
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -595,3 +595,92 @@ func TestDeleteUser_MissingIDIsNotFound(t *testing.T) {
 
 	assert.Equal(t, http.StatusNotFound, w.Code)
 }
+
+// --- GetScoresProgress ---
+
+// int32PtrAuthz keeps the OpDiv fixture below readable without pulling in a
+// shared helper from another test file.
+func int32PtrAuthz(v int32) *int32 { return &v }
+
+// TestScopeScoreProgressInput pins the role matrix the progress endpoint
+// applies to its query input. This is the security boundary for /scores/
+// progress: the SQL builder is tested separately, so what matters here is
+// that each tier populates (or leaves empty) exactly the scope fields the
+// builder keys on.
+func TestScopeScoreProgressInput(t *testing.T) {
+	t.Run("OwnerUnrestricted", func(t *testing.T) {
+		input := model.FindScoreProgressInput{}
+		scopeScoreProgressInput(adminUser, &input)
+
+		assert.False(t, input.RestrictToOpDivIDs, "OWNER must not be OpDiv-restricted")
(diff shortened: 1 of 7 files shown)


MackOverflow:
Moving this to #407 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates. Your commits and authorship carry over unchanged. Thanks @tarratsco!

PR #408, opened 2026-07-09
feat(auth): logout endpoint clearing app + ALB OIDC sessions (#402)

### LINKED ISSUE #402: Backend: logout endpoint to clear the app session and ALB OIDC cookies
### Feature Type

General Feature

### Workflow Description

ZTMF has no way to log out. Once a user authenticates there is no way to end the session — the `ztmf_session` cookie (minted by the #334 session layer) and the ALB's own OIDC session simply live until their TTLs expire. This issue adds the backend half: a logout endpoint that tears the session down. The UI button that calls it is the frontend partner issue CMS-Enterprise/ztmf-ui#477.

The plumbing already half-exists. `auth.ClearSessionCookie` (`backend/cmd/api/internal/auth/session.go`) writes the expired-cookie response with the matching `Domain`/`Path` attributes and its doc comment says "used on logout or when a session is rejected" — but nothing calls it. Add an unauthenticated logout route (e.g. `POST /api/v1/auth/logout`) registered on the `root` router in `router.go` beside `/auth/lookup` and `/login`, so it stays reachable even when the session is already expired. The handler calls `ClearSessionCookie` and returns 204 (or 302 to `/`).

Clearing `ztmf_session` alone is not a complete logout, because the ALB keeps its own OIDC session cookies (`AWSELBAuthSessionCookie` / `AWSELBAuthSessionCookie-Entra`, plus `-0`/`-1` shards when large):

- **Dual-IdP mode** (`entra_enabled=true`): `/api/*` is gated by `ztmf_session`, so clearing it does block API access — but a still-valid ALB session will silently re-authenticate the user on their next `/login` click with no credential prompt.
- **Single-IdP mode**: `/api/*` is gated by the ALB OIDC session itself, so expiring the ALB cookies is *required*, not best-effort.

The ALB cookies are on the same domain, so the logout handler should expire them too (AWS's documented ALB logout pattern). Full IdP-side SSO logout (Okta/Entra `end_session` redirect) is **out of scope** — no end-session endpoint is configured anywhere in code or Terraform today, so that's a separate config + infra addition; file a follow-up if we want it.

New endpoint means swag annotations (copy the style on `LookupIdP` in `controller/auth.go`) and a regenerated `openapi.yaml` (`make generate-openapi` + `make lint-openapi`), plus an emberfall case so the contract can't regress.

### Acceptance Criteria

- [ ] Unauthenticated logout endpoint clears `ztmf_session` (cookie attributes match mint, so the browser actually deletes it); subsequent `/api/*` requests return 401
- [ ] Endpoint also expires the ALB `AWSELBAuthSessionCookie[-Entra]` cookies (including shards) so logout is real in single-IdP mode and doesn't silently re-auth in dual-IdP mode
- [ ] Endpoint is reachable with an expired or missing session (registered outside `auth.Middleware`)
- [ ] After logout, hitting `/login` again prompts a fresh IdP login rather than silently re-authenticating (verified on dev in both IdP modes)
- [ ] `openapi.yaml` regenerated via `make generate-openapi` and lint-clean; emberfall coverage for the endpoint
- [ ] Unit coverage for the handler (cookie attributes, status code)

### Additional Context

Builds on the #334 session layer (session mint, `ClearSessionCookie`, cookie config: `AUTH_SESSION_COOKIE_NAME`, `AUTH_COOKIE_DOMAIN`, `AUTH_SESSION_TTL`). Delivery order: this endpoint lands first; the frontend button (CMS-Enterprise/ztmf-ui#477) points at it once it's on dev. IdP end-session (true SSO logout at Okta/Entra) is intentionally deferred to a follow-up.

MackOverflow:
> **Note:** In-repo copy of #403 by @voidspooks. Moved from the fork so Snyk
>
> Refs #403

---

Closes #402.

## What

Adds `POST /api/v1/auth/logout`, registered on the unauthenticated `root` router (alongside `/auth/lookup` and `/login`) so it can clear a session that is already expired or invalid — it must not sit behind `auth.Middleware`.

The handler:
- Clears the `ztmf_session` app session cookie (via the existing `ClearSessionCookie`).
- Expires the ALB OIDC session cookies — `AWSELBAuthSessionCookie` (Okta) and `AWSELBAuthSessionCookie-Entra` (Entra), plus their `-0`/`-1` shards — as **host-only** deletes (no `Domain`) to match how the ALB sets them.
- Runs the same same-origin CSRF check the middleware applies to state-changing session requests (it's outside the middleware, so it does this itself), returning the standard `FORBIDDEN_ORIGIN` error shape on a cross-origin request.
- Returns `204 No Content`.

## Why the ALB cookies matter

Clearing `ztmf_session` alone is not a complete logout:
- **Dual-IdP mode** (`entra_enabled=true`): `/api/*` is gated by `ztmf_session`, so clearing it blocks API access — but a still-valid ALB session silently re-authenticates the user on their next `/login` with no credential prompt.
- **Single-IdP mode**: `/api/*` is gated by the ALB OIDC session itself, so expiring the ALB cookies is required, not best-effort.

The cookie names mirror the `session_cookie_name` values in `infrastructure/alb-internal.tf` (documented in a comment kept in sync by hand).

## Out of scope

IdP-side single logout (Okta/Entra `end_session` redirect). No end-session endpoint is configured anywhere in code or Terraform today; that's a separate config + infra change. This ends the ZTMF and ALB sessions only.

## Tests

- **Unit** (`session_test.go`): cookie clearing (session + ALB families, host-only attributes, expired), and CSRF rejection on a foreign origin returning `FORBIDDEN_ORIGIN`.
- **E2E** (`emberfall_tests.yml`): `204` on same-origin POST, `403` + `FORBIDDEN_ORIGIN` on a cross-origin POST.
- Regenerated `openapi.yaml`; the deliberately-public logout route is added to the redocly `security-defined` ignore list next to `/auth/lookup` (the rule stays a hard error for every other handler).

Ran locally and passing: `make test-unit`, `make test-integration`, `make test-e2e` (143 emberfall, 0 failed), `make lint-openapi`.

The frontend logout button is the partner ticket, CMS-Enterprise/ztmf-ui#477.

### DIFF
--- a/backend/.redocly.lint-ignore.yaml
+++ b/backend/.redocly.lint-ignore.yaml
@@ -4,14 +4,18 @@
 # `redocly lint` in CI, while every other instance of these rules stays a hard
 # error. Keep it minimal — do not add an entry to silence a real problem.
 #
-# security-defined on GET /auth/lookup: this is the single deliberately public,
-# pre-auth route (per-user IdP lookup for the login landing page). It has no
-# `@Security bearerAuth` by design; the auth boundary is the ALB/OIDC handshake
-# downstream, not this endpoint. Every other operation still carries bearerAuth,
-# and the rule stays an error for them so a new handler that forgets `@Security`
-# fails the build. The shared-handler `path-parameters-defined` warnings and the
-# `info-license` warning are intentionally NOT ignored here — they stay visible
-# as warnings (see redocly.yaml) rather than silenced.
+# security-defined on the deliberately public, pre-auth auth routes:
+#   - GET  /auth/lookup: per-user IdP lookup for the login landing page.
+#   - POST /auth/logout: clears the session, and must work when the session is
+#     already expired, so it cannot sit behind auth either.
+# Both have no `@Security bearerAuth` by design; the auth boundary is the
+# ALB/OIDC handshake downstream, not these endpoints (logout additionally runs
+# its own same-origin CSRF check). Every other operation still carries
+# bearerAuth, and the rule stays an error for them so a new handler that forgets
+# `@Security` fails the build. The shared-handler `path-parameters-defined`
+# warnings and the `info-license` warning are intentionally NOT ignored here —
+# they stay visible as warnings (see redocly.yaml) rather than silenced.
 openapi.yaml:
   security-defined:
     - '#/paths/~1auth~1lookup/get'
+    - '#/paths/~1auth~1logout/post'
--- a/backend/cmd/api/internal/auth/session.go
+++ b/backend/cmd/api/internal/auth/session.go
@@ -212,3 +212,72 @@ func ClearSessionCookie(w http.ResponseWriter) {
 		SameSite: http.SameSiteStrictMode,
 	})
 }
+
+// albSessionCookieNames are the OIDC session cookies the ALB sets, mirrored by
+// hand from the authenticate-oidc rules in infrastructure/alb-internal.tf: the
+// AWS-default AWSELBAuthSessionCookie used by the Okta login rule, and the
+// explicit AWSELBAuthSessionCookie-Entra on the Entra rule. Clearing only
+// ztmf_session ends the app session, but the ALB keeps its own session in these
+// cookies; without expiring them too, a still-valid ALB session silently
+// re-authenticates on the next /login in dual-IdP mode, and in single-IdP mode
+// the ALB session is what gates /api/* in the first place. Keep in sync with the
+// terraform session_cookie_name values.
+var albSessionCookieNames = []string{
+	"AWSELBAuthSessionCookie",
+	"AWSELBAuthSessionCookie-Entra",
+}
+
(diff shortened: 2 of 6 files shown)


danielbowne (approved):
LGTM. Ran the local E2E on the isolated suite (143 pass / 0 fail, and the logout 204 + cross-origin 403 CSRF cases are in the run) - green. Cookie clearing matches how each family is set: app session domain-scoped, ALB cookies host-only to match the ALB, so the browser actually deletes them. CSRF reuses the same `sameOrigin` the auth middleware uses, so there's no divergent path to bypass, and the unauth placement is safe (POST-only, no side effects, just Set-Cookie + 204).

Two optional low-sev notes for later, neither blocking: the ALB shard clearing is hardcoded to {base, -0, -1} so a future `-2` shard would survive logout (low risk - dropping -0/-1 already breaks the set); and no rate limiter on /logout unlike /auth/lookup (fine here since it does zero DB work). 🚀

PR #390, opened 2026-07-06
add isso_name to fismasystems (migration 0044)

MackOverflow:
## Summary

Fixes #389. Migration 0043 landed 11 of the 12 HHS inventory columns; `isso_name` (ISSO display name, distinct from `issoemail`) was omitted. This PR adds it with the same RBAC-gated treatment as the other 11 HHS fields.

- **`0044isssoname.go`** — `ALTER TABLE public.fismasystems ADD COLUMN IF NOT EXISTS isso_name varchar(255)` (nullable, safe retry)
- **`fismasystems.go` (model)** — `ISSOName *string` added to `FismaSystem` struct and `fismaSystemColumns`; `isso_name` added to `Save()` INSERT cols/vals and UPDATE `hhsCols` map
- **`fismasystems.go` (controller)** — `clearHHSMetadata` and `copyHHSMetadata` updated to cover all 12 HHS fields
- **`openapi.yaml`** — `isso_name: type: string` added to `FismaSystem` schema (alphabetically between `issoemail` and `legacy`)
- **`emberfall_tests.yml`** — `hhsExpected`, `fismaSystemWithHHSData`, `system1002WithHHSData` anchors include `isso_name: "Darth Vader"`; scoped-admin clear-gate assertion adds `isso_name: null`
- **`_test_data_empire.sql`** — `isso_name` column added to the INSERT for all 15 HHS fixture rows (2001–2015); ~99% populated, rows with null `isso_name` left null to reflect real data sparsity

## Test plan

- [ ] `make generate-openapi` (if applicable to this repo) and confirm no diff beyond the new field
- [ ] Run emberfall tests locally against a migrated DB — all existing HHS RBAC tests should pass; the clear-gate assertion now also checks `isso_name: null`
- [ ] Manually verify: OWNER can set `isso_name` on create/update; OPDIV_ADMIN create strips it to null; OPDIV_ADMIN update preserves it via `copyHHSMetadata`

PR #386, opened 2026-07-06
HHS schema integration (rehome of #380)

danielbowne:
Rehomes #380 by @costacalvin so CI and Snyk can run against an in-repo branch. Fork-head PRs cannot reach the org secrets Snyk needs, which gates every job that depends on the analysis stage. No history was rewritten; all commits are authored by @costacalvin.

Full review history and approval are on #380.

## Summary

HHS schema integration. Migration 0043 adds 11 nullable HHS inventory columns to `fismasystems` and a `notes_is_ai_summary` boolean to `scores`. Model and RBAC changes gate the 11 HHS field writes to unscoped admins (owner / HHS admin); scoped admins see them read-only, and a partial update no longer wipes stored values. Includes Emberfall E2E coverage for the write gate and the flag semantics, plus local mock data.

Reviewed and approved on #380; e2e verified green locally.

PR #410, opened 2026-07-09
Refresh impl from main (HHS schema, isso_name, datacenter scoring, target maturity, logout)

danielbowne:
Brings the impl environment up to date with main. impl carries no bespoke logic - this is main plus impl env scaffolding, per the refresh flow.

## What lands on impl
- HHS schema integration (#386)
- isso_name on fismasystems, migration 0044 (#390)
- Data-center environments mapped to a shared scoring vocabulary (#392/#394)
- Per-system target maturity tier + justification (#398 MVP / #401)
- Identifiable system questionnaire updates (#407)
- Datacalls resolve current/latest by deadline, not datacallid (#397)
- Logout endpoint clearing app + ALB OIDC sessions (#402/#408)
- Emberfall e2e runner made OS-aware for macOS (#383)

## Sync mechanics
- Merged origin/main into impl (merge, not rebase, per impl policy). Clean except one conflict.
- Sole conflict: backend/compose-test.yml. Resolved to main's bridge-networking model (service-name test-db, published ports; the macOS fix in #384), with only the host-published ports kept parametrized (TEST_DB_PORT / TEST_API_PORT) so the impl test stack can run alongside main's without clobbering host ports. Container ports and service-name wiring are main's.
- Migrations 0043-0046 are all merged on main - no shared-DB strand risk relative to main.
- impl scaffolding preserved (Makefile env-include header, Makefile.impl/main, impl.tfvars, backend-impl.tf, orchestration-impl.yml).

## CI note
Emberfall e2e is green. A separate, pre-existing main issue is being fixed on main in parallel: the new integration test TestFindScoreProgressIntegration (added in #407) is red because findPreviousDataCall now resolves the previous cycle by deadline across all datacalls (#397), so its synthetic datacall collides with seed datacalls and the score copy FK-aborts. That test is not run by any CI job today (only the impl pre-push hook runs go integration tests), so it does not affect this PR's checks or the deployed artifact, which is already live in dev/prod. The fix and a CI gate to run integration tests on main are tracked separately.

## Deploy
Squash-merge triggers orchestration-impl.yml -> impl stack in the dev account -> impl.ztmf.cms.gov.


MackOverflow (approved):
looks good once this is resolved (which is the plan): TestFindScoreProgressIntegration is currently failing on main due to an FK collision introduced by the deadline-based datacall resolution change ([#397](https://github.com/MackOverflow/Zero_Trust/issues/397)), and states no CI job runs Go integration tests today (only a local pre-push hook does). Since this claim is load-bearing for "this PR doesn't make CI redder than it already is," it's worth a quick independent check — confirm no .github/workflows/*.yml actually invokes go test ./... -tags=integration (or similar) before merging, since a workflow file wouldn't necessarily show up in this diff if it already existed on main.

Issue #411, opened 2026-07-10
copyPreviousScores silently swallows errors, can open a new data call with zero carried-over answers


danielbowne:
### Description

`copyPreviousScores` (`backend/internal/model/scores.go`) rolls the previous cycle's answers into a newly created data call. It swallows every failure:

- On `findPreviousDataCall` error: `log.Println(err); return` (nil).
- On the bulk `INSERT ... SELECT` error: `log.Println(err)` and returns without surfacing it.

It is also launched as `go copyPreviousScores(...)` from `datacalls.go`, so even returning an error could not propagate to the caller.

Net effect: if the previous cycle contains a single FK-inconsistent (or otherwise rejectable) score row, the whole `INSERT ... SELECT` aborts, gets logged, and the new data call opens with **zero carried-over answers** - with no error to the API caller and no operator signal. Every system in the new cycle would appear to have made no progress.

### Steps to Reproduce

Not easily reproducible in normal operation (real previous cycles are consistent). The risk surfaces if a previous cycle ever holds a score row that violates a copy-time constraint. Surfaced during review of #<this PR>: the score-progress integration test exercises the happy path of this copy; the sad path has no coverage or signal.

### Expected Behavior

A failed or partial rollover is surfaced - the copy errors loudly (metric / event / alert / non-silent log), so an empty new cycle is detected rather than shipped silently.

### Actual Behavior

Both error paths are logged and swallowed; the goroutine launch means no error can propagate. An empty rollover looks identical to a successful one from the outside.

### Environment

ztmf backend, data call creation path.

### Acceptance Criteria

- A rollover that copies zero rows (or errors) is distinguishable from a successful one at runtime (surfaced via an event/metric/alert, not just an anonymous log line).
- Decide whether `copyPreviousScores` should run synchronously (so failures can propagate) or remain async with explicit failure signaling.
- Consider whether the bulk copy should be resilient to individual bad rows rather than all-or-nothing.

### Severity

Medium - latent data-integrity foot-gun; no current known trigger, but silent when it fires.

danielbowne:
I will grab this one.

PR #412, opened 2026-07-10
Fix score-progress integration test and gate integration tests in CI

danielbowne:
Two deliverables, one PR.

## 1. Fix `TestFindScoreProgressIntegration`

The test (added in #407) has been red on main since it merged, because no CI job on main runs the Go integration tests (see deliverable 2). Root cause:

`copyPreviousScores` resolves "previous" via `findPreviousDataCall`, which post-#397 returns the datacall with the **globally furthest deadline** other than the target (correct in prod, where the new cycle always has the furthest deadline). The empire seed carries a far-future cycle - the "Audit Fields Smoke Cycle", datacallid 5, deadline `2099-12-31`. So `findPreviousDataCall(newDC)` returned that 2099 seed datacall, `copyPreviousScores` rolled its scores rather than the test's marker into `newDC`, and the marker lookup found no rows.

Fix: anchor the two synthetic datacalls' deadlines above the current `MAX(deadline)` (`prevDC` at max+24h, `newDC` at max+48h), so `newDC` is the global latest and `prevDC` is the one `findPreviousDataCall` rolls forward - independent of whatever the seed holds, and self-healing across reruns. The copy-then-edit semantics under test are unchanged (progress is computed from event presence, not deadlines).

## 2. Gate integration tests in CI

`go test -run Integration` ran nowhere on main - only impl's pre-push hook executed it. Added an `integration-tests` job to `backend.yml` that runs `make test-integration` (isolated, freshly-seeded compose-test DB, brought up and torn down). It is a **peer of `build`**, so it gates the merge (mark it a required check in branch protection); prod, which deploys post-merge, therefore never runs on a red integration test. The comment documents that `build`'s deploy runs in parallel and how to gate the deploy itself if wanted later.

## Verification

- `make test-integration` green (was red before the fix).
- `make test-full` green (unit + 143 e2e, 0 failures).
- `gofmt` / `go vet` clean.

## Follow-ups (out of scope, filed)

- **#411** - `copyPreviousScores` silently swallows both its errors (and runs as a goroutine), so a bad previous cycle can open a new data call with zero carried-over answers with no signal. The happy path is now covered by this test; the sad path deserves loud failure.
- Minor: `make test-integration` uses a fixed `sleep 15` rather than a readiness poll; fine locally, could flake on a cold CI runner now that it is on the critical path.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -9,6 +9,37 @@ on:
         required: true
         
 jobs:
+  integration-tests:
+    name: Integration Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          # Read the required Go from go.mod (setup-go v6 sets GOTOOLCHAIN=local,
+          # so a pinned version below go.mod's would not auto-download).
+          go-version-file: backend/go.mod
+          cache-dependency-path: backend/go.sum
+
+      # Runs `go test -run Integration ./...` against an isolated, freshly-seeded
+      # DB brought up and torn down by backend/compose-test.yml. This is the only
+      # place on main that executes the Go integration tests; without it, an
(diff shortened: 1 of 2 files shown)


jsos3-cms (approved):
LGTM! Clean work

MackOverflow (approved):
lgtm Danny, although Palp only has read only and REBELLION is disabled?!? For shame...

PR #415, opened 2026-07-10
Skip local seed when the database already has data

### LINKED ISSUE #414: Local dev api crash-loops on restart: non-idempotent seed re-runs against persisted volume
Restarting the local backend with `make dev-down && make dev-up` (or any `make dev-up` after the first) crashes the api container on boot once the dev database volume already contains data.

## Symptom

The UI shows "We are experiencing technical difficulties," and the api container is `Exited (1)`:

```
executing migrations...
Populating database with /app/_test_data_empire.sql
ERROR: duplicate key value violates unique constraint "datacall_key" (SQLSTATE 23505)
```

## Root cause

Two things combine:

1. `make dev-up` regenerates `backend/dev.compose.env` on every run with `DB_POPULATE=/app/_test_data_empire.sql` set, so the seed step runs on every boot.
2. The seed script (`_test_data_empire.sql`) is not idempotent - it INSERTs fixed-key rows. The local docker volume persists across `docker compose down` (no `-v`), so the second and later boots re-run the seed against an already-populated database and hit a duplicate-key error, which `log.Fatal`s the process.

This bites anyone who has loaded real data into the local volume (e.g. a dev/prod DB sync) and then restarts the backend.

## Expected

On local/test, seeding should preserve an already-populated database rather than crashing. A fresh (empty) volume should still seed normally.

## Fix

Gate `populate()` on an empty-database check (row count on `fismasystems`). If the database already holds data, skip the seed and leave it untouched. Ephemeral E2E databases start empty each run, so they continue to seed.

danielbowne:
Fixes #414.

## Problem

Restarting the local backend with `make dev-up` after the first run crash-loops the api container. The seed script (`_test_data_empire.sql`) is not idempotent, and the local docker volume persists across `docker compose down`, so every boot after the first re-runs the seed against an already-populated database and dies with:

```
ERROR: duplicate key value violates unique constraint "datacall_key" (SQLSTATE 23505)
```

This is especially painful after loading real data into the local volume (e.g. a dev/prod sync), because the only workaround was to wipe the volume - destroying exactly the data you wanted to keep.

## Change

Gate `populate()` on an empty-database check. Before loading the seed, it counts rows in `fismasystems` (a core table created by migrations and populated by the seed). If the database already holds data, seeding is skipped and the existing data is left untouched. A fresh, empty volume still seeds as before.

The check runs before any connection is acquired for the load, and only reads the filesystem when it actually intends to seed, so an already-seeded restart neither touches the seed file nor holds a pooled connection unnecessarily.

## Why this is safe for CI

The Emberfall E2E stack and the Go integration stack both spin up an ephemeral database with no persistent volume, so they start empty every run and continue to seed normally. Only a persisted local dev volume is affected.

## Testing

- Added `TestPopulateSkipsWhenSeededIntegration`, which asserts the gate reports an already-seeded DB and that a second `populate()` is a no-op (it passes a nonexistent path to prove the gate returns before any file read).
- `make test-full` passes locally: 143 Emberfall E2E checks, 0 failures, plus the integration suite.
- Manually verified `make dev-down && make dev-up` on a persisted volume now logs `Skipping populate: database already contains data` and boots cleanly, with the loaded data (1369 systems) intact.

### DIFF
--- a/backend/cmd/api/internal/migrations/populate.go
+++ b/backend/cmd/api/internal/migrations/populate.go
@@ -9,19 +9,53 @@ import (
 )
 
 func populate(path string) error {
-	log.Printf("Populating database with %s\n", path)
+	ctx := context.TODO()
+
+	// The seed script is not idempotent (it INSERTs fixed-key rows and fails with a
+	// duplicate-key error on a second run). A local dev volume persists across restarts,
+	// so on every `make dev-up` after the first the seed would otherwise re-run against
+	// an already-populated database and crash the api on boot. Skip seeding when the
+	// database already holds data, preserving whatever the developer has loaded (e.g. a
+	// dev/prod DB sync). Ephemeral test databases (Emberfall E2E) start empty each run,
+	// so they still seed normally.
+	seeded, err := alreadySeeded(ctx)
+	if err != nil {
+		return err
+	}
+	if seeded {
+		log.Print("Skipping populate: database already contains data")
+		return nil
+	}
+
+	log.Printf("Populating database with %s", path)
 	sql, err := os.ReadFile(path)
 	if err != nil {
 		return err
 	}
-
-	ctx := context.TODO()
 
 	conn, err := db.Conn(ctx)
 	if err != nil {
 		return err
 	}
 	defer conn.Release()
+
 	_, err = conn.Exec(ctx, string(sql))
 	return err
+}
+
+// alreadySeeded reports whether the database already holds seed data. It checks
+// fismasystems, a core table created by migrations and populated by the seed script.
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
Not a blocker, but there's no test for the "empty database" path — i.e., that populate() still seeds normally when fismasystems is empty. The PR description says this was verified manually (fresh volume, 1369 systems), but the automated suite only covers the skip branch. Since the integration DB is always pre-seeded per make test-integration, this may be hard to test without a schema-reset step, which could explain the omission — but it's the one path most likely to regress silently in the future (e.g., someone renames the check column or breaks the empty-count query).

PR #417, opened 2026-07-10
Add system_insights table (JSONB seam for ZTMF Insights)

### LINKED ISSUE #416: Add system_insights table (JSONB seam for ZTMF Insights)
## What
Add `public.system_insights` via a numbered tern migration in `backend/cmd/api/internal/migrations/`, following the exact pattern of `0038systemenrichment.go`. Use the **next available** number - `0047` as of this writing (highest is `0046targetmaturity.go`), but confirm the highest in `backend/cmd/api/internal/migrations/` at implementation time; migrations are landing fast. File: `00XXsysteminsights.go` (e.g. `0047systeminsights.go`).

```go
package migrations

func init() {
	getMigrator().AppendMigration(
		"add generic system_insights extension table",
		// UP: Generic, insight-agnostic per-question extension point owned by ztmf
		// core. The ztmf-insights insights-sync lambda populates the jsonb payload
		// from Snowflake (CORE.ZTMF_ENRICHMENT_INSIGHTS), keyed on
		// (fismasystemid, questionid). Adding or removing insight fields (Kion,
		// SecurityHub, Hardenize, CFACTS, ARS scores/evidence) is a payload-key
		// change in that pipeline and requires no change here. Keyed on
		// fismasystemid (the app PK, always populated for every OpDiv and matching
		// the API param + users_fismasystems RBAC) rather than fisma_uuid. No FK:
		// this is a disposable TRUNCATE+INSERT sync cache, mirroring
		// system_enrichment; fisma_uuid is kept inside the payload for display.
		`CREATE TABLE IF NOT EXISTS public.system_insights (
			fismasystemid INTEGER     NOT NULL,
			questionid    INTEGER     NOT NULL,
			payload       JSONB       NOT NULL,
			synced_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
			PRIMARY KEY (fismasystemid, questionid)
		);`,
		// DOWN
		`DROP TABLE IF EXISTS public.system_insights;`)
}
```

## Why
Carries per-system x question ZTMF Insights scoring (Kion / SecurityHub / Hardenize / CFACTS / ARS) from Snowflake into Aurora so the API can serve it for the 8/1 HHS data-call user testing. This is a distinct grain from `system_enrichment`, which carries per-**system** CFACTS metadata keyed on `fisma_uuid`; `system_insights` is the per-**question** insights seam.

## Design decisions (locked)
- **Generic JSONB seam, not typed columns.** All scoring/evidence lives in `payload`; adding a field later is a payload-key change with no further ztmf migration - same philosophy as `system_enrichment` (0038).
- **Key = `(fismasystemid, questionid)`.** `fismasystemid` (the app's own PK) is the natural key: it matches the API param (`?fismasystemid=`) and the RBAC scoping (`users_fismasystems`) directly, for both reads and writes. `fisma_uuid` is kept inside the payload for display, not as the key. The choice is about API/RBAC alignment, not uuid availability (see scope below - every row here has a uuid anyway).
- **CMS-only data scope.** The insights SDL (Snowflake) holds CMS data only - HHS is not in this SDL and will not be. So `system_insights` will only ever be populated for CMS-OpDiv systems, all of which have a `fisma_uuid`. The schema stays generic and OpDiv-neutral (ztmf core owns it), but the *data* is CMS-scoped, which is why the downstream `/insights` surface is CMS-conditional.
- **No foreign key (deliberate).** Mirrors `system_enrichment`: this is a disposable TRUNCATE+INSERT sync cache written by the lambda, so a FK to `fismasystems` / `questions` would only risk sync failures on a system/question mid-change. Referential integrity is the sync's responsibility, not the schema's.
- **No datacall dimension** - grain is current-state per system x question.
- **Table owned by ztmf core.** The migration lives here; the ztmf-insights `insights-sync` lambda only writes to it (never creates or migrates it).

## Downstream (both block on this table)
- `GET /api/v1/insights?fismasystemid=&questionid=` (this repo). Its model (`FindSystemInsights`) will scope reads with the same predicates as `FindScores` / `FindScoreProgress` - `users_fismasystems` + `opdiv_id`, fail-closed on empty grants - which the `fismasystemid` key supports directly. No new authz model needed.
- `insights-sync` lambda (ztmf-insights repo): `CORE.ZTMF_ENRICHMENT_INSIGHTS` -> this table.

Target: table + migration merged Week 1 (7/13-7/17).

## Acceptance
- [ ] Migration `00XXsysteminsights.go` (next available number; `0047` at time of writing) adds `public.system_insights` with the DDL above (`CREATE TABLE IF NOT EXISTS`, `func init()` + `getMigrator().AppendMigration`, matching 0038)
- [ ] Down migration drops it cleanly
- [ ] `make dev-setup` applies it clean; `\d system_insights` shows the composite PK
- [ ] Applied to dev

danielbowne:
## What

Adds `public.system_insights` via migration `0047systeminsights.go`, following the pattern of `0038systemenrichment.go`.

```
public.system_insights (
  fismasystemid INTEGER     NOT NULL,
  questionid    INTEGER     NOT NULL,
  payload       JSONB       NOT NULL,
  synced_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  PRIMARY KEY (fismasystemid, questionid)
)
```

## Why

This carries per-system x question ZTMF Insights scoring (Kion / SecurityHub / Hardenize / CFACTS / ARS) from Snowflake into Aurora so the API can serve it for the HHS data-call user testing. It is a distinct grain from `system_enrichment`, which carries per-system CFACTS metadata keyed on `fisma_uuid`; `system_insights` is the per-question seam.

## Design decisions

- **Generic JSONB payload, not typed columns.** All scoring and evidence live in `payload`, so adding a field later is a pipeline payload-key change with no further migration - same approach as `system_enrichment`.
- **Key is `(fismasystemid, questionid)`.** `fismasystemid` is the app's own PK and matches both the API param and the `users_fismasystems` RBAC scoping directly, for reads and writes. `fisma_uuid` stays inside the payload for display.
- **No foreign key (deliberate).** Mirrors `system_enrichment`: a disposable TRUNCATE+INSERT sync cache written by the lambda. A FK to `fismasystems` / `questions` would only risk sync failures on a system or question mid-change; referential integrity is the sync's responsibility.
- **No datacall dimension.** Grain is current-state per system x question.
- **Table owned by ztmf core.** The migration lives here; the ztmf-insights sync lambda only writes to it.

## Testing

- `make dev-setup` applies the migration clean on a fresh database.
- `\d public.system_insights` shows the composite primary key.
- Backend compiles.

Closes #416

### DIFF
--- a/backend/cmd/api/internal/migrations/0047systeminsights.go
+++ b/backend/cmd/api/internal/migrations/0047systeminsights.go
@@ -1,0 +1,24 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		"add generic system_insights extension table",
+		// UP: Generic, insight-agnostic per-question extension point owned by ztmf
+		// core. The ztmf-insights sync lambda populates the jsonb payload from
+		// Snowflake, keyed on (fismasystemid, questionid). Adding or removing
+		// insight fields (Kion, SecurityHub, Hardenize, CFACTS, ARS scores or
+		// evidence) is a payload-key change in that pipeline and requires no change
+		// here. Keyed on fismasystemid (the app PK, always populated and matching
+		// the API param + users_fismasystems RBAC) rather than fisma_uuid, which is
+		// kept inside the payload for display. No FK: this is a disposable
+		// TRUNCATE+INSERT sync cache, mirroring system_enrichment.
+		`CREATE TABLE IF NOT EXISTS public.system_insights (
+			fismasystemid INTEGER     NOT NULL,
+			questionid    INTEGER     NOT NULL,
+			payload       JSONB       NOT NULL,
+			synced_at     TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+			PRIMARY KEY (fismasystemid, questionid)
+		);`,
+		// DOWN
+		`DROP TABLE IF EXISTS public.system_insights;`)
+}


jsos3-cms (approved):
Approving the easiest PR ever, a non-functional schema change for a future endpoint :P

PR #419, opened 2026-07-10
Add GET /api/v1/insights reader for per-system per-question ZTMF Insights

danielbowne:
## What

Adds `GET /api/v1/insights`, the read side of the `system_insights` seam (table added in #417). Serves per-system x question insight records - the Kion / SecurityHub / Hardenize / CFACTS / ARS scoring and evidence written from Snowflake by the ztmf-insights sync - so the UI can surface a suggested maturity score and its supporting findings alongside each questionnaire item.

New `model.FindSystemInsights`, `controller.ListSystemInsights`, and the route. Optional `fismasystemid` and `questionid` query filters.

## Access control (same model as the score pages)

1. **Authenticated** - behind `auth.Middleware`; no token is a 401.
2. **Scoped by tier**, identical to `FindScores`: unscoped admin tiers see everything; per-OpDiv admin tiers fail closed to their granted OpDivs' systems; ISSO/ISSM/OWNER are scoped to their `users_fismasystems` grants. Scope is applied server-side after query decode, so a client cannot widen it via params.
3. **OpDiv gate** - on top of the above, a row is visible only if its owning OpDiv has `insights_enabled = TRUE`. Insights data exists for CMS OpDivs only, so a system in a non-enabled OpDiv returns no rows.

Net: only people assigned to a system (or admins) can see its insights.

## Payload is opaque passthrough

`payload` is returned as-is (generic JSONB), same approach as `system_enrichment`. When the upstream Snowflake view adds a field, it appears in the API on the next sync with no change here - verified locally by adding a nested `findings` object (Kion/SecurityHub/Hardenize arrays with id, NIST control, description, remediation) to the view and watching it flow straight through.

## Testing

- **Integration** (`systeminsights_integration_test.go`): the RBAC proof against real SQL + the empire seed - users_fismasystems scope, fail-closed OpDiv restriction, the `insights_enabled` gate hiding a REBELLION system's row from an unscoped admin, and jsonb passthrough. Passes under `make test-integration`.
- **Emberfall**: status + array-body smoke for the admin and ISSO paths, plus the empty-array behavior for the gated and unassigned cases (the list-endpoint counterpart to systemenrichment's 403/404). Passes under `make test-e2e`.
- **Fixtures**: Star Wars themed `system_insights` rows in `_test_data_empire.sql` (no real data).
- OpenAPI regenerated (`make generate-openapi`).

## Notes

- `payload` is typed as a generic object in OpenAPI (`swaggertype:"object"`) on purpose - typing all ~75 keys would kill the additive property.
- Depends on #417 (merged).

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -571,6 +571,36 @@ INSERT INTO public.system_enrichment (fisma_uuid, payload, synced_at) VALUES (
     '2026-05-20 00:00:00+00'
 ) ON CONFLICT (fisma_uuid) DO NOTHING;
 
+-- Generic system_insights rows (issue #416), keyed on (fismasystemid, questionid).
+-- The payload is opaque to ztmf core (owned by the insights pipeline); these are
+-- Star Wars themed mocks that mirror the real payload shape (per-source suggested
+-- score/label, evidence_sources, and a nested findings object with id / nist /
+-- description / remediation). They give /insights E2E + integration coverage:
+--   1003 (Shield Gen)  - EMPIRE (insights_enabled), assigned to the Emberfall ISSO
+--   1001 (Death Star)  - EMPIRE (insights_enabled), NOT assigned to that ISSO,
+--                        so an ISSO read must return 1003 but never 1001 (scope)
+INSERT INTO public.system_insights (fismasystemid, questionid, payload, synced_at) VALUES
+  (1003, 1,
+   '{"pillar":"Identity","question_text":"How does the system authorize access?","suggested_score":2,"suggested_label":"Initial","evidence_sources":"Kion, SecurityHub","score_floor_source":"Kion (failing checks)","findings":{"kion":[{"id":"shield-relay-without-mfa","nist_controls":"IA-2","description":"Shield generator control relays must enforce multi-factor authentication for operator access.","remediation":"Enable Force-signature MFA on all shield relay operator accounts."}],"sechub":[{"id":"SG.1","title":"Shield frequency should rotate on a fixed schedule","severity":"MEDIUM","description":"Static shield frequencies are predictable and can be mapped by an adversary.","remediation":"Rotate deflector shield frequency per operational window."}]}}',
+   '2026-05-20 00:00:00+00'),
+  (1003, 2,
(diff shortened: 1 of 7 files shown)


MackOverflow (approved):
lgtm no blockers, some things we can refine over time

MackOverflow:
Query-param decode errors likely return 500, not 400. If a client sends ?fismasystemid=abc, decoder.Decode errors and gets passed straight to respond. Worth confirming that surfaces as a 4xx client error rather than a server error — a malformed param is the caller's fault.
Redundant filtering path. The EXISTS OpDiv gate and the si.fismasystemid IN (SELECT ... WHERE opdiv_id = ANY(?)) clause both touch fismasystems. Correct, but a reviewer might ask whether they could be unified. Not worth changing for a read endpoint at this scale.
Auth middleware wiring isn't visible in this diff — the route is registered but the auth.Middleware wrapping happens elsewhere in router.go. The PR body claims it's covered; the E2E tests exercising 401/scope behavior are the real proof, so this is fine, just not self-evident from the diff alone.

danielbowne:
Thanks for the review. Dispositions:

1. **Malformed query param returns 500, not 400** - real, but pre-existing and codebase-wide, not introduced here. `GET /api/v1/scores?fismasystemid=abc` returns 500 identically; this endpoint mirrors `ListScores`' `decoder.Decode` + `respond()` path exactly, and `sanitizeErr` has no case for a schema decode error. Fixing it only here would make the insights endpoint inconsistent with the rest of the list endpoints. Tracked as a standalone follow-up in #420 (map schema decode errors to 400 in `sanitizeErr`, fixing all list endpoints at once). Merging as-is for consistency.

2. **Redundant filtering path (EXISTS gate + `opdiv_id = ANY` both touch fismasystems)** - intentional; it mirrors `FindScores` and keeps the two concerns (data-availability gate vs OpDiv scope) legible. Unifying them is a micro-optimization on a read at this scale. Leaving as-is.

3. **Auth wiring not visible in the diff** - correct that it's not in the diff. The route is registered on the `router` subrouter, which has `router.Use(auth.Middleware)` (router.go, outside the diff hunk). The E2E 401 + ISSO scope tests are the real proof. No change needed.

Merging on the strength of the approval + green CI (integration + E2E covering the scope/gate matrix).

PR #425, opened 2026-07-13
Parametrize compose postgres host port so dev and impl run side by side (main port of #422)

danielbowne:
Ports the impl-branch fix for #422 to `main`.

## Problem

The `compose-dev.yml` generator in the Makefile hardcoded the postgres host-port mapping (and the `-p` flag and healthcheck) as `54321`. On `main`, `DB_PORT` is 54321 so it happened to match and went unnoticed — but a worktree that overrides the port (impl uses 54322 via `Makefile.impl`) regenerated a mapping that still published host **54321**, colliding with the dev worktree's postgres and stopping dev + impl from running at the same time.

## Why this isn't just cherry-picking a39b3cc

The impl fix (`a39b3cc`) is one line — `"54321:54321"` → `"$(DB_PORT):$(DB_PORT)"` — because the impl branch already defines `DB_PORT` as a make variable and already uses it in the `-p` flag and healthcheck. **main defines only `API_PORT`, not `DB_PORT`**, so that one-liner would expand to `":"` and break the generator here.

So this adds the missing infra to match: `DB_PORT ?= 54321` (mirroring `API_PORT`), then threads it through every hardcoded db-port spot in the generator — the host-port mapping, the postgres `-p` flag, the healthcheck, the `dev.compose.env` `DB_PORT` line, and the info echoes.

## Behavior

- **main unchanged** — `DB_PORT` defaults to 54321, so the generated compose is byte-identical to before.
- A per-worktree override now works end to end: `make DB_PORT=54322 …` renders `54322:54322` / `-p 54322` / healthcheck 54322 / env `DB_PORT=54322`.
- Container/volume/network isolation is already handled by `COMPOSE_PROJECT_NAME`.

## Verification

- `DB_PORT=54322` renders 54322 throughout the generated `compose-dev.yml`; default renders 54321 (confirmed both).

Closes the main-side gap of #422.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,10 @@ SHELL := /bin/bash
 
 # Single source of truth for the local API port
 API_PORT ?= 8080
+# Postgres host port for the local dev compose stack. Overridable per worktree
+# (e.g. impl sets 54322) so dev and impl can run side by side without colliding
+# on the published host port. Container isolation is handled by COMPOSE_PROJECT_NAME.
+DB_PORT ?= 54321
 
 # Emberfall E2E runner install, resolved per platform. Releases are published as
 # emberfall_{Darwin,Linux}_{arm64,x86_64,i386}.tar.gz, so a hardcoded Linux URL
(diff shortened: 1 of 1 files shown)


MackOverflow (approved):
looks good to me, super clean and straightforward

PR #426, opened 2026-07-14
fix(api): return 400 not 500 for malformed list-endpoint query params (#420)

voidspooks:
## Summary

Fixes #420. GET list endpoints that decode query params via `gorilla/schema` returned **HTTP 500** when a param failed to parse. The decode error from `decoder.Decode(&input, r.URL.Query())` was passed straight into `respond()`, whose `sanitizeErr` had no case for a schema decode error, so it fell through to a 500. A malformed client-supplied param is the caller's fault and should surface as a **4xx** — a 500 inflates 5xx error metrics/alerts and gives the caller a misleading status.

This is pre-existing and codebase-wide: it affected every GET list handler using the shared `decoder.Decode` + `respond()` pattern (scores, insights, and others), not one endpoint.

## Fix

Took the smaller, consistent option from the issue: map `gorilla/schema` decode errors to **400** in the shared `sanitizeErr`, which fixes all list endpoints at once.

- `sanitizeErr` now matches `schema.MultiError` and the `ConversionError` / `UnknownKeyError` / `EmptyFieldError` it wraps, returning **400** with a dedicated `ErrInvalidQueryParam` sentinel. The sentinel keeps the response message clean instead of leaking the library's internal parse detail (e.g. `strconv.ParseInt: ... invalid syntax`).

## Before / after

```
GET /api/v1/scores?fismasystemid=abc     500 -> 400
GET /api/v1/insights?fismasystemid=abc   500 -> 400
GET /api/v1/insights?questionid=xyz      500 -> 400
```

Valid requests are unaffected — the mapping only triggers on a decode error, which a well-formed request never produces.

## Tests

- `controller_test.go` — unit coverage for `sanitizeErr`: drives the real `decoder` so the concrete error type (`schema.MultiError` wrapping a `ConversionError`) matches what handlers pass in, covers the unknown-key path, and adds a table pinning the existing status mappings so the new branch can't shadow them.
- `emberfall_tests.yml` — 400 cases for `scores?fismasystemid=abc` and `insights` with a non-numeric `fismasystemid` / `questionid` (the decode error short-circuits before the insights-enabled gate, so 400 regardless of enablement).
- Added `@Failure 400` to the `ListScores` / `ListSystemInsights` swag annotations and regenerated `openapi.yaml` (redocly lint clean).

Full suite green locally: `make test-unit`, `make test-integration`, `make test-e2e` (emberfall 150/0).

Closes #420.

PR #430, opened 2026-07-14
fix(api): return 400 not 500 for malformed list-endpoint query params (#420)

### LINKED ISSUE #420: GET list endpoints return 500 instead of 400 on malformed query params
## Describe the bug

GET list endpoints that decode query params via gorilla/schema return **HTTP 500** when a param fails to parse. The decode error from `decoder.Decode(&input, r.URL.Query())` is passed straight into `respond()`, whose `sanitizeErr` has no case for a schema decode error, so it falls through to a 500. A malformed client-supplied param is the caller's fault and should surface as a **4xx**, not a server error.

## To reproduce

```
GET /api/v1/scores?fismasystemid=abc     -> 500
GET /api/v1/insights?fismasystemid=abc   -> 500
GET /api/v1/insights?questionid=xyz      -> 500
```

## Expected behavior

`400 Bad Request` (client error), so malformed input does not inflate 5xx error metrics / alerts and the caller gets an actionable status.

## Scope

Pre-existing and codebase-wide: it affects every GET list handler using the shared `decoder.Decode` + `respond()` pattern (scores, insights, and others), not one endpoint. Surfaced during review of #419 (the insights reader), which mirrors `ListScores` exactly - both return 500 identically.

## Fix sketch

Either:
- map schema decode errors (`schema.MultiError` / `schema.ConversionError`) to 400 inside `sanitizeErr` (fixes all list endpoints at once), or
- wrap the `decoder.Decode` error as `*model.InvalidInputError` (already maps to 400) at the controller layer (explicit per-handler).

The first is the smaller, consistent change.

jsos3-cms:
## Summary

In-repo rehome of #426 by @voidspooks (Cameron Testerman) so the repository CI gates (which fork PRs cannot run) execute against this commit. Original authorship is preserved on the cherry-picked commit; the original PR #426 stays open as a draft until this merges.

Fixes #420. GET list endpoints that decode query params via `gorilla/schema` returned **HTTP 500** when a param failed to parse. The decode error from `decoder.Decode(&input, r.URL.Query())` was passed straight into `respond()`, whose `sanitizeErr` had no case for a schema decode error, so it fell through to a 500. A malformed client-supplied param is the caller's fault and should surface as a **4xx** — a 500 inflates 5xx error metrics/alerts and gives the caller a misleading status. This is pre-existing and codebase-wide: it affected every GET list handler using the shared `decoder.Decode` + `respond()` pattern (scores, insights, and others), not one endpoint.

## Fix

Map `gorilla/schema` decode errors to **400** in the shared `sanitizeErr`, which fixes all list endpoints at once. `sanitizeErr` now matches `schema.MultiError` (and the `ConversionError` / `UnknownKeyError` / `EmptyFieldError` it wraps) and returns **400** with a dedicated `ErrInvalidQueryParam` sentinel. The sentinel keeps the response message clean instead of leaking the library's internal parse detail (e.g. `strconv.ParseInt: ... invalid syntax`).

Valid requests are unaffected: `gorilla/schema` `Decode` returns an untyped `nil` on success and only returns a non-empty `MultiError` on an actual decode failure, so the new case can never trip on a well-formed request. The case returns early, ahead of the existing `errors.Is` chain, so it does not shadow any current status mapping.

## Before / after

```
GET /api/v1/scores?fismasystemid=abc     500 -> 400
GET /api/v1/insights?fismasystemid=abc   500 -> 400
GET /api/v1/insights?questionid=xyz      500 -> 400
```

## Tests

- `controller_test.go` — unit coverage for `sanitizeErr`: drives the real `decoder` so the concrete error type (`schema.MultiError` wrapping a `ConversionError`) matches what handlers pass in, covers the unknown-key path, and adds a table pinning the existing status mappings so the new branch can't shadow them.
- `emberfall_tests.yml` — 400 cases for `scores?fismasystemid=abc` and `insights` with a non-numeric `fismasystemid` / `questionid` (the decode error short-circuits before the insights-enabled gate, so 400 regardless of enablement).
- Added `@Failure 400` to the `ListScores` / `ListSystemInsights` swag annotations and regenerated `openapi.yaml`.

## Provenance / verification

Commit cherry-picked verbatim from #426 (`voidspooks/ztmf:fix/list-endpoints-400-malformed-params`), authorship preserved. On this branch atop current `main`: `go build ./...`, `go vet`, and the `controller` unit-test package are green; `make generate-openapi` reproduces the committed `openapi.yaml` byte-for-byte (no drift). The `gorilla/schema` v1.4.1 error-type behavior above was confirmed against the library source.

### DIFF
--- a/backend/cmd/api/internal/controller/controller.go
+++ b/backend/cmd/api/internal/controller/controller.go
@@ -97,6 +97,15 @@ func sanitizeErr(err error) (int, string, error) {
 	switch err.(type) {
 	case *model.InvalidInputError:
 		return 400, "", err
+	case schema.MultiError, schema.ConversionError, *schema.ConversionError,
+		schema.UnknownKeyError, *schema.UnknownKeyError,
+		schema.EmptyFieldError, *schema.EmptyFieldError:
+		// A gorilla/schema decode failure means the caller sent a query param
+		// that can't be parsed into the input struct (e.g. fismasystemid=abc).
+		// That's a client error across every GET list handler using the shared
+		// decoder.Decode + respond() pattern, so surface a clean 400 rather than
+		// falling through to a 500 that inflates 5xx metrics. (#420)
+		return 400, "", ErrInvalidQueryParam
 	}
 
 	var (
--- a/backend/cmd/api/internal/controller/controller_test.go
+++ b/backend/cmd/api/internal/controller/controller_test.go
@@ -1,0 +1,67 @@
+package controller
+
+import (
+	"errors"
+	"net/url"
+	"testing"
+
+	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+	"github.com/stretchr/testify/assert"
+)
+
+// A malformed query param decoded via gorilla/schema must surface as a 400
+// (client error), not a 500. Regression for #420: the decode error fell through
+// sanitizeErr to the default 500 branch, inflating 5xx metrics for what is the
+// caller's mistake. Exercises the real decoder so the concrete error type
+// (schema.MultiError wrapping a ConversionError) matches what handlers pass in.
+func TestSanitizeErrMapsSchemaDecodeErrorTo400(t *testing.T) {
+	var input struct {
+		FismaSystemID int `schema:"fismasystemid"`
+	}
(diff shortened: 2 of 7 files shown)


MackOverflow (approved):
lgtm

PR #432, opened 2026-07-15
fix(model): make users_fismasystems Save idempotent on duplicate assignment (#429)

### LINKED ISSUE #429: users_fismasystems Save is not idempotent: re-assigning an existing system errors instead of upserting
### Type
Bug (flagged by an existing TODO)

### Description
`UserFismaSystem.Save` is meant to be idempotent but errors when the assignment already exists. The marker:

```go
// TODO: fix issue where inserting the same (userid, fismasystemid) twice returns error. It should effectively upsert
```

`backend/internal/model/usersfismasystems.go:40`

```go
sqlb := stmntBuilder.Insert("userid, fismasystemid").
    Into("users_fismasystems").
    Values(uf.UserID, uf.FismaSystemID).
    Suffix("ON CONFLICT DO NOTHING RETURNING userid, fismasystemid")

return queryRow(ctx, sqlb, pgx.RowToStructByName[UserFismaSystem])
```

### Root cause
`ON CONFLICT DO NOTHING RETURNING` returns **no rows** when a conflict occurs (nothing is inserted, so there's nothing to return). `queryRow` then surfaces that as an error (`pgx.ErrNoRows` → `model.ErrNoData`). So re-assigning a system a user already has fails instead of being a no-op — the opposite of the intended upsert.

The table supports a proper conflict target:

```sql
CREATE TABLE public.users_fismasystems (
  userid uuid REFERENCES users (userid) ON DELETE CASCADE,
  fismasystemid INT REFERENCES fismasystems (fismasystemid) ON DELETE CASCADE,
  PRIMARY KEY (userid, fismasystemid)
);
```

### Reproduce
`POST /api/v1/users/{userid}/assignedfismasystems` with a `fismasystemid` the user already has → error, where a second identical assignment should succeed idempotently.

### Proposed fix
Make the conflict path return the existing row:

```go
Suffix("ON CONFLICT (userid, fismasystemid) DO UPDATE SET userid = EXCLUDED.userid RETURNING userid, fismasystemid")
```

(`DO UPDATE` with a no-op SET so `RETURNING` yields the row on conflict.) Alternatively keep `DO NOTHING` and treat the no-rows result as success in `Save`. The `DO UPDATE` form is simpler and keeps the single-row return contract.

### Acceptance criteria
- Assigning a system a user already has succeeds idempotently (no error).
- Integration coverage: a duplicate `Save` returns the existing `(userid, fismasystemid)` without error; a first-time `Save` still inserts and returns the row.
- Emberfall: a repeated `POST .../assignedfismasystems` returns success rather than 4xx/5xx.

voidspooks:
Closes #429.

## Problem

`UserFismaSystem.Save` is meant to be idempotent, but re-assigning a system a user already had returned a 404 instead of succeeding.

The insert used `ON CONFLICT DO NOTHING RETURNING`. On a duplicate `(userid, fismasystemid)`, `DO NOTHING` suppresses the insert, so `RETURNING` yields zero rows; `queryRow` surfaces that as `pgx.ErrNoRows` → `model.ErrNoData`, which the API layer maps to a 404. This was the long-standing `// TODO: fix issue where inserting the same (userid, fismasystemid) twice returns error` marker.

## Fix

Switch to a no-op `DO UPDATE` so the conflicting row is visible to `RETURNING`:

```
ON CONFLICT (userid, fismasystemid) DO UPDATE SET userid = EXCLUDED.userid RETURNING userid, fismasystemid
```

`DO UPDATE SET userid = EXCLUDED.userid` writes the row's own value back to itself — a no-op to the data — but marks the row as affected so `RETURNING` returns it. This preserves `queryRow`'s single-row contract on both paths: a fresh insert returns the new row, a duplicate returns the existing one.

## Tests

- **Integration** (`usersfismasystems_integration_test.go`): first `Save` inserts and returns the row (exactly 1 row); duplicate `Save` is a no-op that still returns the existing row (still exactly 1 row — no duplication).
- **Emberfall e2e**: POST `.../assignedfismasystems` twice returns 201 with the same row both times (was 404 on the second), then DELETE restores seed state.

Ran locally and passing: `make test-unit`, `make test-integration`, `make test-e2e` (153 emberfall cases, 0 failed).

> Note: the target user in the e2e/integration tests must have a version-4 UUID — both `validate()` and `FindUserByID` gate on the v4-strict `isValidUUID`, so the all-digits empire placeholder users 404 before reaching `Save`. Thrawn (`aa000088-…`) is v4 and carries no seed assignments.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -1341,6 +1341,56 @@ tests:
       headers:
         content-type: "application/json"
 
+  # System assignment is idempotent (#429). Save uses ON CONFLICT ... DO UPDATE
+  # ... RETURNING so a duplicate POST returns the existing row; the old
+  # DO NOTHING RETURNING produced zero rows on conflict, which surfaced as a
+  # 404. Target user must have a version-4 UUID: FindUserByID and validate()
+  # gate on the v4-strict isValidUUID, so the all-digits placeholder users
+  # (11111111-..., 22222222-...) 404 before reaching Save. Thrawn (aa000088-...)
+  # is v4 and carries no seed assignments.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems"
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        fismasystemid: 1003
+    expect:
+      status: 201
+      body:
+        json:
+          data:
+            userid: "aa000088-8888-4888-8888-888888888888"
+            fismasystemid: 1003
+
+  # Identical POST again - the regression target. Must return the same row,
+  # not 404/500.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems"
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        fismasystemid: 1003
+    expect:
+      status: 201
+      body:
+        json:
+          data:
+            userid: "aa000088-8888-4888-8888-888888888888"
+            fismasystemid: 1003
+
+  # Remove the assignment so Thrawn's scope stays as seeded for any tests
+  # that follow.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems/1003"
+    method: DELETE
+    headers:
+      <<: *commonHeaders
+    expect:
+      status: 200
+
   # PUT returns 204 (No Content) per controller.respond; assert the write
   # path completes cleanly. Audit-on-write is covered by the POST assertion
   # above.
--- a/backend/internal/model/usersfismasystems.go
+++ b/backend/internal/model/usersfismasystems.go
@@ -37,11 +37,15 @@ func (uf *UserFismaSystem) Save(ctx context.Context) (*UserFismaSystem, error) {
 		return nil, err
 	}
 
-	// TODO: fix issue where inserting the same (userid, fismasystemid) twice returns error. It should effectively upsert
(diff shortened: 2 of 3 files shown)


jsos3-cms:
Thanks @voidspooks. Fork PRs can't run our repository CI gates (Snyk / secret-dependent checks), so I've rehomed your commit in-repo as #438 with your authorship preserved — cherry-picked verbatim atop current `main`. This PR stays as-is (draft) while #438 is in flight, and `Fixes #429` now lives on #438; I'll close this once #438 merges. Nothing needed from you to land the fix.

The `DO UPDATE` idempotency fix itself is correct and race-safe (the conflict target matches the `PRIMARY KEY (userid, fismasystemid)`, and the no-op `SET` can't clobber anything since the table is only the composite key).

One thing worth a follow-up (non-blocking, your call): because `DO UPDATE` always returns a row, `queryRow` now reaches `recordEvent` on an idempotent duplicate POST and logs it as `action: "created"`. So a repeated "assign system X to user Y" adds phantom `created` entries to the `events` audit trail — whereas before, the duplicate errored before any event was recorded. It doesn't affect this fix's API contract or any data, but it does slightly muddy the audit trail (which the app treats as a real admin-facing surface).

`usersopdivs.go`'s `Save` already solves this exact case a different way — `ON CONFLICT (userid, opdiv_id) DO NOTHING RETURNING …`, catch `ErrNoData` from `queryRow`, substitute the input struct as the result, and never call `recordEvent` on the duplicate path — so it's idempotent *without* the phantom audit event. The #429 issue mentions the `DO NOTHING` option was considered; flagging the audit-log consequence in case you'd prefer to align with the `usersopdivs` idiom. Happy either way — just didn't want it to slip by silently.

jsos3-cms:
Closed in https://github.com/CMS-Enterprise/ztmf/pull/438

PR #434, opened 2026-07-15
ci: skip Go lint and Snyk Go scan on infra-only PRs

### LINKED ISSUE #301: CI: skip Go lint and Snyk Go scan on infra-only PR pushes
### Type

Chore

### Description

The Analysis stage in `orchestration-dev.yml` (and the `analysis.yml` workflow it calls) runs `lint go` and `snyk test` on every PR push regardless of which files changed. Pure infrastructure-only changes (Terraform, tfvars, IAM, CloudFront, ALB, monitoring) trigger the full Go pipeline even though no `.go` / `.sum` / `Dockerfile` files were touched.

The repository already computes a backend-changed signal in the `Check for Changes` job to gate the Backend deploy:

```yaml
- name: Check for changes in backend/
  id: backend
  run: echo "backend=$(git --no-pager diff --name-only origin/main backend/ | grep ".go\|.sum\|Dockerfile" | head -n 1)" >> $GITHUB_OUTPUT
```

That same signal could gate the Go-specific Analysis jobs:

- `Analysis / lint go`
- `Analysis / snyk test` (the Go module Snyk scan, distinct from `snyk iac test` and `snyk code test`)

The Terraform / IAC scans should still run on every PR since they evaluate the full infrastructure regardless of which file changed.

### Why

- Wall-clock savings on infra-only PRs (~1 to 2 minutes per push, more on larger Go workspaces).
- Cleaner PR check status for infra reviewers.
- Reduces noise from Go-related Snyk findings on PRs that are not modifying any Go code.

### Acceptance criteria

- `Analysis / lint go` and `Analysis / snyk test` run only when the same backend diff signal that gates the Backend deploy job is non-empty.
- `Analysis / lint terraform`, `Analysis / snyk iac test`, and `Analysis / snyk code test` continue to run on every PR push.
- Pure Terraform / tfvars PR shows lint go and snyk test as `skipping`, not `pending` then `success`.
- A PR that touches both Go and Terraform still runs all Analysis jobs.

### Related

Surfaced while landing the cert-rotation Lambda activation PR (#300). The branch had two infra-only commits where Go lint and Snyk Go ran for no reason.

danielbowne:
## Summary

Closes #301. The Analysis stage ran `lint go` and `snyk test` (the Go module Snyk scan) on every PR push, even when no Go files changed - so pure infra/Terraform PRs paid for the full Go pipeline for nothing.

This adds a `changes` job to `analysis.yml` that computes the same backend diff signal the orchestration workflows already use to gate the Backend deploy, and gates the two Go-specific jobs on it. The Terraform, IaC, code, and OpenAPI scans are unchanged and still run on every PR.

## Behavior

- Infra-only PR push -> `lint go` and `snyk test` show as skipped; `lint terraform`, `snyk iac test`, `snyk code test`, and the OpenAPI check still run.
- PR touching Go (`.go`, `go.mod`, `go.sum`, `Dockerfile`) or the backend/analysis workflows -> all Analysis jobs run.
- Outside the PR-open context (a merge to main/prod, or any push) the signal is forced non-empty, so the Go scans always run there - prod and impl behavior is unchanged.
- Draft PRs are unaffected (the Analysis workflow still runs; only the Go jobs are conditional).

## Trade-off

`snyk test` is a point-in-time SCA scan. Gating it means an infra-only PR no longer re-surfaces a newly-disclosed CVE in an otherwise-unchanged Go dependency. Acceptable because the merge-to-main/prod and impl paths always re-run it, and `snyk code test` (Go SAST) stays always-on.

## Why it helps

- Cuts ~1-2 minutes of compute per infra-only PR push.
- Cleaner check status for infra reviewers.
- Partial rehome relief: the Go Snyk scan needs the org `SNYK_TOKEN` that fork PRs can't access, so skipping it on infra-only changes removes one of the token-gated jobs a fork would otherwise fail on. Note `snyk code test` and `snyk iac test` also need the token and are not gated, so this reduces but does not eliminate the rehoming need for infra-only fork PRs.

## Testing

`analysis.yml` validated (parses; only `lint go` and `snyk test` carry the `needs: changes` gate). Reviewed by Codex - no critical issues; the deploy-implies-scan invariant holds across dev/impl/prod/push paths. `go.mod` added to the signal per that review.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -4,8 +4,37 @@ on:
   workflow_call:
 
 jobs:
+  # Detect whether this change touches Go/backend files so the Go-specific
+  # Analysis jobs (lint go, snyk test) can be skipped on infra-only PRs. Mirrors
+  # the backend diff signal the orchestration workflows use to gate the Backend
+  # deploy. Outside the PR-open context (a merge to main/prod, or any push) the
+  # signal is forced non-empty so the Go scans always run there.
+  changes:
+    name: detect backend changes
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      backend: ${{ steps.backend.outputs.backend }}
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for changes in backend/
+        id: backend
+        run: |
+          if [ "${{ github.event_name }}" != "pull_request" ] || [ "${{ github.event.pull_request.merged }}" = "true" ]; then
+            echo "backend=always" >> $GITHUB_OUTPUT
+          else
+            echo "backend=$(git --no-pager diff --name-only "origin/${{ github.base_ref }}" -- backend/ .github/workflows/backend.yml .github/workflows/analysis.yml | grep -E '\.go$|go\.mod$|\.sum$|Dockerfile|workflows/(backend|analysis)\.yml' | head -n 1)" >> $GITHUB_OUTPUT
+          fi
+
   Lint-Go:
     name: lint go
+    needs: changes
+    if: ${{ needs.changes.outputs.backend != '' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -81,6 +110,8 @@ jobs:
 
   Snyk-Test:
     name: snyk test
+    needs: changes
+    if: ${{ needs.changes.outputs.backend != '' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


jsos3-cms (commented):
This is Clean, and I like how it's building out. I want to keep this "changes" pattern, I think it's slick

MackOverflow (dismissed):
clean, one small nit for future readiness: 

Two copies of near-identical diff logic will drift. analysis.yml's changes and orchestration-dev.yml's diff are now intentionally divergent supersets. That's correct today, but the next person editing one grep won't know to update the other. Consider a shared composite action or a comment in each pointing at the other. This is the highest-value maintainability note.

MackOverflow (approved):
lgtm

Issue #435, opened 2026-07-15
Persist questionnaire answer state (simple state machine) instead of deriving "updated this cycle" from the events audit log


voidspooks:
## Problem

The crux — what this should give us:

- Past calls persist their real state, permanently, as a first-class fact — not reconstructed by lateral-joining an audit table at read time.
- The progress query becomes a trivial `COUNT(*) WHERE state = 'finished'` instead of the `INNER JOIN LATERAL … events … LIMIT 1` gymnastics.
- Correctness stops depending on a best-effort, non-transactional, user-context-gated event write. The thing I confirmed is genuinely fragile; a state column set transactionally with the answer can't silently disagree with reality.

The dashboard's "Data Call Progress" metric answers "which systems have genuinely updated their questionnaire *this cycle*." Today that's derived at read time by lateral-joining the **`events` audit table**: `FindScoreProgress` (`backend/internal/model/scoreprogress.go`) counts a function as updated only if its `scores` row has ≥1 edit event.

The reason it keys off events at all: when a new data call opens, `copyPreviousScores` (`scores.go:731`) pre-populates every answer from the prior call via a raw `Exec` that deliberately records **no** event. So "has a `scores` row" reads 100% answered on day one, and event-presence is used as a stand-in for "touched since the copy."

**Using an audit log as a source of business truth is the wrong coupling, and it's fragile.** Concretely:

1. **Event writes are fire-and-forget.** `queryRow` (`model.go:94`) calls `recordEvent` and ignores its result — a failed event insert is silently swallowed after the score write already succeeded.
2. **Not transactional.** `recordEvent` (`events.go:49`) runs on a *separate* connection, so the answer and its event are never atomic.
3. **Context-gated.** `recordEvent` early-returns if there's no authenticated user in context (`events.go:76`) — some write paths persist the answer but no event.

When any of those drop an event, the progress metric silently under-counts. The metric's correctness depends on an audit log that is explicitly best-effort and lossy.

## Impact

Surfaced concretely in **ztmf-ui#537**: past data calls show `0/40 Not updated` for systems that were actually complete, because those calls were completed via carry-forward/seed/migration with no edit events. The frontend fix there had to lean on a *second* proxy — "has a `systemscore` ⇒ complete" — because `ScoreProgress` has no field that actually records answer state. So the UI is currently stacking two inferences (event-presence ≈ touched, score-presence ≈ complete), neither of which is a persisted fact.

[...]

Issue #299, opened 2026-04-30
Systems that have updated are identifiable


eschwein-usds:
### Feature Type

General Feature

### Workflow Description

OpDiv Admins would like to know which systems have not updated their Questionnaires since the start of the Data Call.

Optional -- know who has not viewed the Questionnaires at all

Possibly a report that can be downloaded by the Admins, with some system contact information?
Can talk to NIH for user input

### Acceptance Criteria

- [ ] There is a method for OpDiv Admins to obtain this information

jsos3-cms:
Onix has requested to do this work

jsos3-cms:
Closed by https://github.com/CMS-Enterprise/ztmf-ui/issues/489

PR #437, opened 2026-07-15
Persist per-answer status on scores instead of deriving progress from events

MackOverflow:
## Summary

Implements #435: persist each questionnaire answer's review state as a first-class `scores.status` column, and derive "Data Call Progress" from it instead of lateral-joining the `events` audit log.

The progress metric used to infer "updated this cycle" from the presence of an edit event. That coupling is fragile — event writes are fire-and-forget (`queryRow` ignores `recordEvent`'s result), non-transactional (separate connection), and context-gated (skipped when there's no authenticated user). Any dropped event silently under-counts progress. This moves the truth onto the `scores` row itself, written in the **same statement** as the answer, so it can never disagree with the row it describes.

Opening as a **draft for @voidspooks to review** since they authored the issue and own the design intent.

## What changed

- **`scores.status`** (`not_started` | `done`), `varchar + CHECK` (matches the `target_maturity_tier` convention, migration `0046`).
- **`scores.Save`** sets `status = 'done'` on a genuine create/change, in the same INSERT/UPDATE as the answer — no transaction needed. A read-through PUT hits the existing no-op guard and never touches status, so a carried-forward answer stays `not_started` (preserves the #299 "carry-forward is not updated" rule).
- **`copyPreviousScores`** seeds carried rows as `'not_started'`.
- **`FindScoreProgress`** counts `status = 'done'` and drops the events INNER lateral. The events lateral stays as a LEFT join, feeding **only** `lastupdatedat` (a legitimately observational audit-timeline use).
- **`ScoreProgress.QuestionsAnswered`** (new): distinct applicable functions with an answer row, any status. This is the answered/total a *closed* call needs — it lets **ztmf-ui#537** retire its "score present ⇒ complete" proxy. `QuestionsUpdated` stays the "touched this cycle" numerator.
- **Migration `0048`** backfills existing rows (edit event → `done`, else `not_started`), mirroring the old derivation exactly so dashboards don't move at cutover. Seed data mirrors the same backfill (it loads after migrations).
- **`openapi.yaml`** regenerated.

## Design decision (see #435 open questions)

Shipped the **two-state** machine (`not_started → done`), not three. A per-answer `complete` state has no trigger in the app today — completion is tracked at the *system* level via `datacalls_fismasystems`, and events can't distinguish "finalized" from "touched" for a backfill. Two states solve the whole stated problem and backfill losslessly; a third can layer on later if product wants per-answer finalization. Happy to revisit if you'd prefer three.

## Acceptance criteria (#435)

- [x] Progress counts derive from persisted `scores.status`, not the `events` table
- [x] A dropped/failed event write no longer affects progress numbers
- [x] `copyPreviousScores` resets carried rows to `not_started`; #299 behavior preserved
- [x] Past data calls report a real answered/total (`QuestionsAnswered`)
- [x] Backfill leaves existing dashboards unchanged at cutover
- [x] `events` is no longer read by progress/completion logic (only the audit-timeline `lastupdatedat`)

## Test plan

- [x] `go build`, `go vet`, `gofmt` clean
- [x] Unit tests updated + passing (`scoreprogress_test.go` — status filter, `QuestionsAnswered`, SQL shape)
- [x] Migration DDL + backfill verified against real Postgres (correct done/not_started split)
- [x] Integration tests (`scoreprogress_integration_test.go`) — run in CI against a migrated DB
- [x] Reviewer to confirm the two-state choice matches intended design

Refs #435

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -906,6 +906,22 @@ SELECT DISTINCT ON (s.scoreid)
  WHERE s.scoreid >= 9100
  ORDER BY s.scoreid, uf.userid;
 
+-- Seed scores.status to match the just-seeded audit trail (ztmf#435). This
+-- runs after migrations, so the migration's one-time backfill has already
+-- executed against an empty table and does NOT see these seed rows; mirror
+-- its exact logic here so the seed DB's progress numbers agree with what the
+-- events above imply. A score with a seeded 'updated' event is 'done'; the
+-- rest keep the not_started default. Keeps this consistent automatically as
+-- the events seed above changes (it currently events only scoreid >= 9100).
+UPDATE public.scores s
+SET status = 'done'
+WHERE EXISTS (
+    SELECT 1
+    FROM public.events e
+    WHERE e.resource = 'public.scores'
+      AND (e.payload->>'scoreid')::int = s.scoreid
+);
+
 -- Reset every SERIAL sequence to its current table max. Use
 -- pg_get_serial_sequence() so the right name is resolved at runtime: some
 -- environments have historically renamed tables (e.g. functionscores -> scores)
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -177,6 +177,13 @@ INSERT INTO public.datacalls VALUES (4, 'FY2025 Death Star Assessment', '2025-01
 -- "datacallid: 5" references and the "?datacallid=5" query string
 -- in that file in lockstep.
 INSERT INTO public.datacalls VALUES (5, 'Audit Fields Smoke Cycle', '2026-01-01T00:00:00Z', '2099-12-31T23:59:59Z') ON CONFLICT DO NOTHING;
+-- Closed, IMPORTED cycle (ztmf#435): mirrors production history loaded from
+-- outside the app - its scores are seeded below with NO events rows, so every
+-- answer keeps status 'not_started' and carries no editor/timestamp. This is
+-- the shape that makes a closed call read 0/N under the "updated" lens while
+-- QuestionsAnswered still reports it complete; keep it event-free on purpose
+-- so that behavior stays reproducible locally.
+INSERT INTO public.datacalls VALUES (6, 'FY2020 Imperial Archives Import', '2020-01-01T00:00:00Z', '2020-12-31T23:59:59Z') ON CONFLICT DO NOTHING;
 
 -- Test FISMA Systems (Imperial Systems)
 -- Use explicit column names to work with initial schema
@@ -879,6 +886,22 @@ BEGIN
         END LOOP;
     END LOOP;
 END $$;
+
+-- Imported-history scores for the FY2020 archives cycle (datacallid 6). Clone
+-- the FY2024 (datacallid 3) answer set so every (system, functionoption) pair
+-- is known-applicable to its system's environment, offset scoreids by +49 to
+-- stay below 9100 - the events seed below only covers scoreid >= 9100, so
+-- these rows get NO events and keep the 'not_started' status default. That is
+-- the point: an imported closed call has answers (QuestionsAnswered > 0) but
+-- no edit provenance (QuestionsUpdated = 0, LastUpdatedAt null), reproducing
+-- the production imported-history shape locally without any real data.
+INSERT INTO public.scores (scoreid, fismasystemid, datecalculated, notes, functionoptionid, datacallid)
+SELECT s.scoreid + 49, s.fismasystemid, '2020-06-01 00:00:00+00',
+       'Recovered from the Imperial Archives - assessor records lost',
+       s.functionoptionid, 6
+  FROM public.scores s
+ WHERE s.datacallid = 3 AND s.scoreid < 9100
+ON CONFLICT DO NOTHING;
 
 -- Audit trail for the seeded scores. last_edited_at / last_edited_by are NOT
 -- stored on scores; FindScores derives them from the events table (the most
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -929,19 +929,44 @@ SELECT DISTINCT ON (s.scoreid)
  WHERE s.scoreid >= 9100
  ORDER BY s.scoreid, uf.userid;
 
+-- Import provenance for the FY2020 archives cycle (datacallid 6): a service
+-- account plus one 'imported' event per archived score, mirroring how
+-- externally-loaded history is attributed in real environments (ztmf#435). These
+-- give the archived rows a who/when (last-updated shows the import instead of
+-- blank) but use action 'imported', which the status-sync below deliberately
+-- excludes - so the archived answers keep status 'not_started' despite carrying
+-- events. This is the shape that proves imported != updated: answers present,
+-- provenance present, never 'done'.
+INSERT INTO public.users (userid, email, fullname, role, deleted, identity_provider)
+VALUES ('00000000-0000-4000-8000-000000000001', 'svc-importer@empire.test',
+        'Imperial Archives Import Service', 'HHS_READONLY_ADMIN', false, 'okta')
+ON CONFLICT DO NOTHING;
+
+INSERT INTO public.events (userid, action, resource, createdat, payload)
+SELECT '00000000-0000-4000-8000-000000000001', 'imported', 'public.scores',
+       '2020-06-01 00:00:00+00',
+       jsonb_build_object('scoreid', s.scoreid, 'fismasystemid', s.fismasystemid,
+                          'functionoptionid', s.functionoptionid,
+                          'datacallid', s.datacallid, 'notes', s.notes)
+  FROM public.scores s
+ WHERE s.datacallid = 6;
+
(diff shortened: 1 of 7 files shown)


voidspooks:
LGTM  🚀

jsos3-cms:
Reviewing now

jsos3-cms:
Reviewed the state-machine change — the core is correct and I'd be comfortable with it once the one test gap below is closed.

**What holds up under trace:**

- **Atomicity**: `status` is set in the *same* INSERT/UPDATE as the answer (`scores.go` Insert ~L109, Update `SetMap` ~L117) — there's no second statement, so status can't disagree with the row it describes. The event write stays decoupled and best-effort, which is the point.
- **#299 carry-forward (code)**: `copyPreviousScores` seeds `'not_started'` in the same `INSERT...SELECT`, and `Save`'s no-op guard returns before any write when `scoresEqualForUpdate` is true, so an unchanged re-save leaves status untouched. Only a genuine create (`ScoreID==0`) or a real change reaches `status='done'`.
- **Migration 0048 backfill**: the `EXISTS (… events … resource='public.scores' …)` predicate mirrors the old `FindScoreProgress` events-lateral exactly, so dashboards won't move at cutover. DDL matches the `target_maturity_tier` (0046) varchar+CHECK convention; `ADD COLUMN … DEFAULT <literal>` is metadata-only on PG16, and `IF [NOT] EXISTS` makes it idempotent.
- **Progress query**: events is now a `LEFT JOIN LATERAL` feeding only `lastupdatedat`; the count is `COUNT(DISTINCT …) FILTER (WHERE status='done')` — no fan-out, no NULL issue. `QuestionsAnswered` is a strict subset of `QuestionsExpected`, so it can't exceed it. Two states satisfy all six acceptance criteria.

**Please add before un-drafting (test coverage — the highest-risk path):** nothing exercises "carried-forward row + identical no-op re-save → status stays `not_started`," which is exactly what the no-op guard protects and what #299 requires. `TestFindScoreProgressIntegration` resaves with *changed* notes (only proves `not_started → done`), and the pre-existing `TestScoreSaveNoOpPreservesPriorEditorIntegration` starts from a freshly-`Save`d row (already `done`) and asserts on events / `LastEditedBy`, never on `scores.status`. Suggested test: `copyPreviousScores` → assert `status='not_started'` → `Save` with identical fields → re-assert still `not_started` (direct DB read + `FindScoreProgress` still reporting `QuestionsUpdated==0`).

**Minor:**

- Migration 0048 runs `ALTER TABLE` + the backfill `UPDATE` in one transaction, so `scores` holds an ACCESS EXCLUSIVE lock through the backfill scan (not just the fast metadata-only ALTER). Fine at current scale; worth knowing before this runs against a large `scores` table.
- The brand-new-answer INSERT path (`ScoreID==0 → 'done'`) isn't asserted end-to-end via the progress endpoint for an *applicable* function (it's only incidentally exercised by the inapplicable-function test, where both counts are 0). Low priority — add for symmetry.

For what it's worth: local `go build`/`go vet`/model unit tests pass, and `make generate-openapi` reproduces the committed `openapi.yaml` with no drift, so the spec is in sync with the new `QuestionsAnswered` field.

danielbowne:
Good review, and agreed on the part that matters: that no-op-preserves-`not_started` test is the one to land before this comes out of draft. It's sitting right on top of the bug we're actually trying to kill here (a carried-forward row sneaking to `done` and over-counting progress), so having a test pinned directly to it is worth it.

Few notes on the migration so nobody loses sleep over the lock:

- The lock worry is softer than it looks. `ADD COLUMN ... DEFAULT 'not_started'` is metadata-only on PG16, so there's no table rewrite, and the backfill `UPDATE` only touches rows that actually have a matching event, riding the `events_score_audit_idx` partial index from 0037. So it's an index probe per row, not a seq scan of the whole table. On our `scores` table that's a short window at cutover, nothing I'd reach for a fancier pattern over at this size.
- Down migration is clean, it drops the column, so a rollback puts `scores` back exactly how it started. Easy to miss but good that it's there.
- The `DEFAULT` + `CHECK` combo is a nice safety net too: any insert path that ever forgets to set `status` just lands on `not_started` instead of blowing up, and can't write a value outside the two states.

So the only thing I'd want before un-drafting is that test. Everything else reads correct to me, once voidspooks signs off on the two-state call.

danielbowne:
@MackOverflow if this is time-sensitive, I can pick up these changes in the morning (the no-op test plus the small migration notes) and get it shipped. Just say the word and I'll take it off your plate.

danielbowne:
Picked this up. Added the no-op-preserves-`not_started` integration test we talked about (carry a row forward, re-save it with identical fields, assert status stays `not_started` and QuestionsUpdated stays 0), and merged latest main into the branch. Integration + unit green locally.

That closes the one code gate from review. The only thing left before this comes out of draft is @voidspooks' call on two-state vs three-state, since that's a design decision on your issue. Once you're good with two states I'll un-draft and let CI take it.

danielbowne:
Since the last look:

- Merged latest main into the branch.
- Added the no-op integration test from review: carry an answer forward, re-save it with identical fields, assert status stays `not_started` and QuestionsUpdated stays 0. It can only pass if the guard actually fires - if the guard regressed, the write would flip status to done and the assertion fails.
- Fixed the ScoreProgress doc comment: updated counts aren't zero for every closed cycle - cycles genuinely worked in-app keep their backfilled counts; it's imported history that reads zero. Answered/total stays the only completion signal for past calls.
- Added a closed 'FY2020 Imperial Archives Import' datacall to the empire seed whose scores intentionally have no events - answers exist but carry no editor or timestamp, so it renders 0/N under the updated lens while questionsanswered reports it complete. @voidspooks this one's for you: it reproduces the imported-history shape locally (it's in the seed the test stack loads), so you can see the exact closed-call behavior this PR fixes without needing access to any real data.

Unit, integration, and the full e2e suite are green locally. On the two-state question: with the seed fixture in front of you the case is easier to see - answers-without-provenance is already representable (row exists, status not_started), so a third persisted state has nothing to encode today; the varchar+CHECK leaves room to add one later without a rewrite if a review workflow ever lands.

Staying in draft until review lands - this PR carries a migration, and deploying an unmerged migration to the shared dev DB strands it ahead of main. Once approved we'll un-draft and merge in one motion.

jsos3-cms (approved):
**Approving.** Correct and safe to un-draft on the engineering merits — the earlier test-coverage gap is closed and the new work holds up under trace and under a live integration run.

## Verified
- `go build` / `go vet` / unit tests clean; `make generate-openapi` reproduces the committed `openapi.yaml` (no drift); migration `0048` self-registers via `init()` and doesn't collide (main tops out at `0047`); the `8e17313` catch-up merge is clean (no conflict markers, no overlap with `scores.go`/`scoreprogress.go`/migrations).
- Full `internal/model` integration suite passes against an isolated seeded Postgres — including `TestScoreNoOpReSavePreservesNotStartedIntegration` and `TestImportedScoresKeepProvenanceWithoutDoneStatusIntegration`. Direct DB check matches intent: FY2020 archives cycle (datacallid 6) = 12 rows, all `not_started`, each carrying one `imported` event and zero `created`/`updated`.

## Correctness confirmed
- **No-op re-save (#299 guard):** `f15b7f8` starts from a genuinely carried-forward (`not_started`) row, re-saves identical fields, and asserts status stays `not_started` via both a direct DB read and `QuestionsUpdated == 0`. That's the exact path a broken no-op guard would regress — good.
- **Imported-provenance exclusion (`e2edee8`):** the backfill/seed filter `action IN ('created','updated')` is an allow-list, so it's a behavioral no-op against all real data (`recordEvent` only ever emits `created`/`updated`/`deleted`); it only changes the outcome for the new `imported` fixture. Mixed rows (an `imported` event *and* a real edit) still flip to `done` correctly via `EXISTS`, so there's no under-counting. The seed status-sync uses the identical predicate, so backfill and seed can't drift.
- **Original design intact atop merged main:** status is set in the same INSERT/UPDATE as the answer (atomic); `events` is now only `LEFT JOIN`ed for `lastupdatedat`; `QuestionsAnswered ⊆ QuestionsExpected`; `copyPreviousScores` seeds `not_started` in the same `INSERT…SELECT`. Migration 0048 (varchar+CHECK, metadata-only ADD COLUMN on PG16, index-assisted backfill) follows the 0046 precedent.

## One non-blocking note
The `imported`-provenance convention is new — it isn't in #435's text or the earlier thread, and it introduces a third de-facto event category alongside the two persisted states. It's implemented correctly, so it doesn't block, but it's worth a quick design confirm with the issue owner (@voidspooks) so the provenance scheme is blessed the same way the two-state-vs-three-state call was.

## Nits (non-blocking, follow-up)
- The action strings (`created`/`updated`/`imported`) are bare literals duplicated across the migration SQL and the seed, with no shared constant (`recordEvent` already uses inline literals, so this is a pre-existing pattern). A future rename would silently desync the backfill.
- `scores.Save` hardcodes `status='done'` with no override, so a future real importer would have to bypass `Save` with raw SQL (as the seed/backfill do). Worth a one-line note or a tracking issue for whoever builds the importer.

@danielbowne — cleared from my side. Good to **undraft and merge** whenever you're ready (optionally after the quick @voidspooks design confirm on the imported-provenance convention). Reminder that main's ruleset needs the DEV deployment green and thread resolution alongside this approval.

PR #438, opened 2026-07-15
fix(model): make users_fismasystems Save idempotent on duplicate assignment (#429)

### LINKED ISSUE #429: users_fismasystems Save is not idempotent: re-assigning an existing system errors instead of upserting

jsos3-cms:
## Summary

In-repo rehome of #432 by @voidspooks (Cameron Testerman) so the repository CI gates (which fork PRs cannot run) execute against this commit. Original authorship is preserved on the cherry-picked commit; the original PR #432 stays open as a draft until this merges.

Fixes #429. `users_fismasystems.Save` was not idempotent: re-assigning a user to a FISMA system they were already assigned to returned an error (HTTP 404) instead of succeeding. The old `ON CONFLICT (userid, fismasystemid) DO NOTHING RETURNING …` returned zero rows on a duplicate, which `queryRow` surfaced as `pgx.ErrNoRows` → `ErrNoData` → 404.

## Fix

Switch to `ON CONFLICT (userid, fismasystemid) DO UPDATE SET userid = EXCLUDED.userid RETURNING …` so the conflict path yields a row without changing any data (the table is exactly the composite-key columns, so the no-op `SET` writes nothing new). A duplicate assignment now returns 201 with the existing row; the normal new-insert path is unchanged. The conflict target matches the table's `PRIMARY KEY (userid, fismasystemid)` (migration 0001), and `ON CONFLICT … DO UPDATE` is a single atomic, race-safe statement.

## Tests

- `usersfismasystems_integration_test.go` — first assignment inserts and returns the row; a duplicate re-assignment succeeds idempotently (no error, correct row, row count stays 1).
- `emberfall_tests.yml` — e2e: a second identical POST expects 201 with the same `userid`/`fismasystemid`.

## Known follow-up (non-blocking)

Because `DO UPDATE` always returns a row, `recordEvent` now fires on an idempotent duplicate POST and logs it as `action: "created"` — so repeated identical assignments add phantom "created" entries to the `events` audit trail (previously the duplicate errored before reaching `recordEvent`). This does not affect the API contract or data, but slightly degrades audit-trail precision. The codebase already has an idiom that avoids this — `usersopdivs.go`'s `Save` uses `DO NOTHING RETURNING`, catches `ErrNoData`, and skips `recordEvent` on the duplicate path. Flagged to the author on #432 for a follow-up decision (keep `DO UPDATE` vs. adopt the `usersopdivs` pattern); not a blocker for this fix.

## Provenance / verification

Commit cherry-picked verbatim from #432 (`voidspooks/ztmf:fix/users-fismasystems-idempotent-save-429`), authorship preserved, atop current `main` (the fork branch's base was stale — the phantom CI-workflow diffs in the fork view are just newer `main` commits it predated). `go build ./...` and `go vet` are green on this branch.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -1341,6 +1341,56 @@ tests:
       headers:
         content-type: "application/json"
 
+  # System assignment is idempotent (#429). Save uses ON CONFLICT ... DO UPDATE
+  # ... RETURNING so a duplicate POST returns the existing row; the old
+  # DO NOTHING RETURNING produced zero rows on conflict, which surfaced as a
+  # 404. Target user must have a version-4 UUID: FindUserByID and validate()
+  # gate on the v4-strict isValidUUID, so the all-digits placeholder users
+  # (11111111-..., 22222222-...) 404 before reaching Save. Thrawn (aa000088-...)
+  # is v4 and carries no seed assignments.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems"
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        fismasystemid: 1003
+    expect:
+      status: 201
+      body:
+        json:
+          data:
+            userid: "aa000088-8888-4888-8888-888888888888"
+            fismasystemid: 1003
+
+  # Identical POST again - the regression target. Must return the same row,
+  # not 404/500.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems"
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        fismasystemid: 1003
+    expect:
+      status: 201
+      body:
+        json:
+          data:
+            userid: "aa000088-8888-4888-8888-888888888888"
+            fismasystemid: 1003
+
+  # Remove the assignment so Thrawn's scope stays as seeded for any tests
+  # that follow.
+  - url: "http://localhost:8080/api/v1/users/aa000088-8888-4888-8888-888888888888/assignedfismasystems/1003"
+    method: DELETE
+    headers:
+      <<: *commonHeaders
+    expect:
+      status: 200
+
   # PUT returns 204 (No Content) per controller.respond; assert the write
   # path completes cleanly. Audit-on-write is covered by the POST assertion
   # above.
--- a/backend/internal/model/usersfismasystems.go
+++ b/backend/internal/model/usersfismasystems.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"context"
+	"errors"
 
 	"github.com/jackc/pgx/v5"
 )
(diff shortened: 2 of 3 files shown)


danielbowne:
Pushed a follow-up commit resolving the audit-trail item this PR flagged as an open follow-up. Given the deadline and the upcoming code freeze, I pushed the fix directly rather than requesting a change, so this isn't blocked on another review round-trip.

The `DO UPDATE` form fixes the 404 but has a side effect: `queryRow` fires `recordEvent` whenever a row is returned, and the action is derived from the builder, so an `Insert` always logs `"created"`. Under the old `DO NOTHING` a duplicate returned zero rows and errored before `recordEvent` ran, so nothing was logged. `DO UPDATE` always returns a row, so every idempotent re-POST would write a false `"created"` event into the `events` table. For a tool whose `events` table is the audit trail, that trades the 404 for an audit-precision bug.

The commit switches to the existing `UserOpDiv.Save` idiom: `DO NOTHING RETURNING`, catch `ErrNoData`, return the in-memory row. Since `users_fismasystems` is exactly its two primary-key columns, the returned row is identical to the stored row, so the API response and the 201 are unchanged. It also drops the no-op UPDATE's dead-tuple write and unifies both junction-table `Save` methods on one pattern. Integration and emberfall pass unchanged (153 e2e, 0 failed).

One note for #429: the issue's proposed fix suggested `DO UPDATE` for simplicity, which is what steered this. Worth correcting there so it doesn't recur.

danielbowne (approved):
With my additional changes LGTM

MackOverflow (approved):
@danielbowne two nits: 

1. Update the PR description so ## Fix describes DO NOTHING + ErrNoData (not DO UPDATE), and remove/rewrite ## Known follow-up since the phantom-event concern is already resolved.
2. Fix the stale DO UPDATE reference in the emberfall_tests.yml.

PR #439, opened 2026-07-17
chore(deps): bump golang.org/x/crypto 0.51.0 -> 0.52.0 (backend)

danielbowne:
Rehomes #413 onto an in-repo branch so CI and the Snyk scans actually run. Dependabot PRs don't get the org secrets, so their checks can't complete, which is why we move the bump onto a branch here.

Bumps `golang.org/x/crypto` 0.51.0 -> 0.52.0 in `/backend` (indirect). The release is mostly `x/ssh` hardening (source-address critical-option bypass, DSA/RSA key parameter limits, a couple of deadlock/overflow fixes). We don't use `x/ssh` directly, so this is low-risk maintenance, but worth taking to stay current.

Verified locally: `go build ./...` clean, `go mod verify` passes, `go mod tidy` produces no drift (only the two lines dependabot changed).

Supersedes #413 - close that one out once this merges.


MackOverflow (approved):
lgtm!

PR #441, opened 2026-07-17
fix(mail): skip NULL/blank recipients so one contactless system can't abort the send

### LINKED ISSUE #440: Email Users fails entirely when any recipient has a NULL/blank email
## Description

The "Email Users" feature fails with a generic `An error occurred, please try again later` for any recipient group that includes a system or user with a NULL (or blank) email. `MassEmail.Recipients()` (`backend/internal/model/massemails.go`) scans every recipient with `pgx.RowTo[string]`, which cannot scan a SQL NULL. Recipient sources include nullable columns — `fismasystems.issoemail` and `fismasystems.datacallcontact` — so a single system without a contact email aborts the entire send before any mail is dialed.

Surfaced in dev after the HHS import brought in systems with NULL `issoemail` / `datacallcontact`. It is not an SMTP or infrastructure problem — the failure happens while building the recipient list, well before the mail code runs.

## Steps to Reproduce

1. Have at least one FISMA system with a NULL `issoemail` or `datacallcontact` (e.g. an imported HHS system).
2. Open the dashboard and click "Email Users".
3. Select group "ALL" (also reproduces for ISSO / DCC), enter a subject and body, and send.
4. See "An error occurred, please try again later"; no email is sent to anyone.

## Expected Behavior

The email sends to every recipient that has a valid email. Systems or users without an email are skipped, not treated as a fatal error that blocks the whole group.

## Actual Behavior

The recipient query returns a NULL email; `pgx.RowTo[string]` fails with `cannot scan NULL into *string`; the request 500s and nothing is sent.

## Environment

Deployed dev (ECS/Fargate), backend API. Reproduces in any environment whose recipient set contains a NULL or blank email.

## Acceptance Criteria

- [ ] "Email ALL" succeeds when some systems/users have NULL or blank emails
- [ ] NULL and empty-string recipients are skipped; valid recipients still receive
- [ ] No regression for groups whose recipients all have valid emails
- [ ] Regression test covering a NULL recipient in the set

## Severity

High - Significantly impacts usage (any single NULL recipient blocks the entire "Email Users" feature; currently blocking the code-freeze announcement to testers).

## Affected Component

Mass email / "Email Users" — `backend/internal/model/massemails.go` (`Recipients`)

## Logs / Error Messages

```
can't scan into dest[0] (col: email): cannot scan NULL into *string
SELECT DISTINCT email as email FROM (
  SELECT email FROM users WHERE role='ISSM'
  UNION ALL SELECT DISTINCT string_to_table(datacallcontact,';') AS email FROM fismasystems
  UNION ALL SELECT email FROM users WHERE role='ISSO'
  UNION ALL SELECT issoemail AS email FROM fismasystems)
```

## Workaround

None in-app — any NULL recipient in the selected group blocks the entire send.

danielbowne:
## Summary

Fixes #440. The "Email Users" feature returned an opaque `An error occurred, please try again later` whenever a recipient group included a system or user with a NULL email. `MassEmail.Recipients()` scanned every recipient with `pgx.RowTo[string]`, which cannot scan a SQL NULL — so a single system with a NULL `issoemail`/`datacallcontact` aborted the entire send **before any mail was dialed**. Surfaced in dev after the HHS import brought in systems without contact emails.

This is not an SMTP or infrastructure change — the failure was entirely in building the recipient list, ahead of the mail code.

## Change

- Scan recipients into `*string` instead of `string`, then filter in a new `dedupeRecipients` helper:
  - drop NULL addresses (nullable sources: `fismasystems.issoemail`, `datacallcontact`)
  - drop blank/whitespace-only addresses (from `string_to_table` splitting a trailing or doubled `;`)
  - de-duplicate case-insensitively, so a person listed via both a `users` row and a system's `issoemail` is emailed once, not twice. (In the CMS/Okta identity context, `A@x` and `a@x` are always the same mailbox.)
- Skip the `mail.Send` goroutine entirely when there are no deliverable recipients (a fully-contactless group, now reachable post-import), rather than dialing the SMTP relay for zero recipients.

## Testing

- Unit: `TestDedupeRecipients` — NULL handling (the crash case), blank/whitespace segments, case-insensitive dedup, trimming, all-filtered.
- Integration: `TestMassEmailRecipientsSkipsNullIntegration` — inserts a system with NULL `issoemail`/`datacallcontact` and asserts the ISSO/DCC/ALL recipient queries no longer crash on the NULL (drives the real scan path, pins the regression end to end).
- `make test-unit` and `make test-integration` green, `go vet` clean.

## Out of scope / follow-ups

- Populating the missing contact emails on imported systems is intentionally left out — the fix skips contactless systems rather than fabricating addresses.
- Worth filing separately: per-recipient send-result capture (pass/fail), and confirming the dev VPC can actually reach the CMS internal SMTP relay (`smtp.cloud.internal.cms.gov`) — this crash previously masked whether it's reachable from dev.

### DIFF
--- a/backend/cmd/api/internal/controller/massemails.go
+++ b/backend/cmd/api/internal/controller/massemails.go
@@ -56,6 +56,15 @@ func SaveMassEmail(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// No deliverable recipients: a group where every system/user is contactless
+	// (far more reachable now that imported systems carry NULL emails). Skip the
+	// goroutine so we don't dial/auth/quit the SMTP relay for zero recipients;
+	// return the empty list so the caller sees nobody was targeted.
+	if len(recipients) == 0 {
+		respond(w, r, recipients, nil)
+		return
+	}
+
 	go mail.Send(m.Subject, m.Body, recipients)
 
 	respond(w, r, recipients, nil)
--- a/backend/internal/model/massemails.go
+++ b/backend/internal/model/massemails.go
@@ -3,6 +3,7 @@ package model
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/Masterminds/squirrel"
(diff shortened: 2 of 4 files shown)


MackOverflow (dismissed):
Correct, well-tested, minimal, and the diagnosis matches the code. The follow-ups @danielbowne  calls out (populating missing contact emails, per-recipient send-result capture, confirming dev→SMTP reachability) are the right things to defer. Only truly optional nit worth considering before merge: 

Empty-list still dials SMTP. When everything filters out, Recipients() returns [], and the controller still fires go mail.Send(...), which dials + authenticates + sends 0 emails. Harmless and pre-existing, but a if len(recipients) == 0 { … } short-circuit in the controller would avoid a pointless SMTP round-trip. Out of scope for this PR.

not a blocker, but maybe something for clean up later on

MackOverflow (approved):
lgtm

danielbowne:
Thanks @MackOverflow! Went ahead and folded in the one nit you flagged: the empty-recipient case now short-circuits in the controller (`if len(recipients) == 0`), so a fully-contactless group no longer dials/auths the SMTP relay for zero sends. Also added an integration test that inserts a system with NULL issoemail/datacallcontact and asserts the ISSO/DCC/ALL recipient queries no longer crash — pins the regression at the actual scan path, not just the filter.

Both landed in 0d5f6c4, which dismissed your review. Re-requesting whenever you have a sec.

Issue #431, opened 2026-07-15
Standardize ZTMF system metadata - retire free-text fields for typed, validated inputs


danielbowne:
## Epic Summary

System metadata on `fismasystems` is stored as free text today. HHS maintained these values in Excel, where they drift over time - typos, inconsistent casing, synonyms, and blanks. This epic makes the application enforce a canonical vocabulary at the point of entry (system add/edit) so the data stays clean going forward.

A one-time import normalizes the existing values separately; this epic is about keeping them normalized in the app - the standardization a spreadsheet could never enforce.

## Field taxonomy - what each field becomes

**Booleans (Yes / No / Unknown - null means unknown, do NOT coerce to No):**
- `hva`
- `cloud_system`
- `legacy`

**Single-select enums (canonical, case-sensitive):**
- `fips` -> Low / Moderate / High
- `system_type` -> Major Application / Minor Application / Minor Standalone / General Support System / Enterprise / Local / Other
- `system_operator` -> Agency / Contractor
- `goco_coco_gogo` -> GOCO / COCO / GOGO

**Multi-select (checkboxes; store decomposed, not one combined string):**
- `cloud_service_model` -> IaaS / PaaS / SaaS / Other

**Structured free text (stays text, normalize on display):**
- `system_owner`, `system_owner_email`, `isso_name` - format "First Last", multiple people joined with "; ".

## Business rules to enforce (backend + UI)

- `cloud_system = No` => `cloud_service_model` and `cloud_vendor` are NULL (clear/hide in UI, reject otherwise in the API).
- Booleans: `null` = unknown (distinct from No).
- Enums: reject anything off the canonical list.

## Success Criteria

- Every field above is entered through a typed, validated control in system add/edit - no free-text entry for the boolean/enum/multi-select fields.
- The API rejects off-canon values and enforces the `cloud_system` dependency rule.
- The frontend renders options from a single backend-provided vocabulary, not hardcoded lists.
- Existing values are migrated in place and backfilled.
- Booleans preserve the unknown (null) state end to end.

## Sub-Issues

> These are intentionally coarse - they need refinement and further breakdown before work starts.

- [ ] Backend - typed columns, vocabulary endpoint, validation, migration (CMS-Enterprise/ztmf)
- [ ] Frontend - typed inputs sourced from the vocabulary endpoint, conditional UI (CMS-Enterprise/ztmf-ui)

## Out of scope (separate future issues)

[...]

PR #443, opened 2026-07-17
fix(fismasystems): clear blank optional fields to NULL, stop blank-email 400 (#442)

### LINKED ISSUE #442: System edit save mishandles empty values: blank email 400s, fields can't be cleared, blanks stored as "" not NULL
## Summary

The admin system add/edit save path (`POST/PUT /fismasystems`) mishandles empty and null values on the optional metadata fields. Three symptoms, one root cause. This blocks epic #431's `cloud_system = No => clear cloud_service_model/cloud_vendor` rule and its "preserve unknown (null) end to end" criterion, so it should land before that work.

## Symptoms (reproduced against a fresh local backend, empire seed, admin JWT)

**1. Blanking an email field 400s the whole save.**
`validate()` (`backend/internal/model/fismasystems.go:518-524`) rejects `datacallcontact`/`issoemail` when non-nil and not a valid email, and `isValidEmail("")` is false (`validations.go`). The frontend sends a cleared input as `""`, not null.
- `PUT issoemail:""` -> `{"data":{"issoemail":""},"error":"invalid input"}` HTTP 400. Baseline unchanged round-trip -> 204.

**2. You can't clear a set metadata field.**
The frontend sends extended fields as `editedSystem[key] ?? null` (`SystemDetailPage.tsx:357-359`); the backend UPDATE writes only non-nil fields (`fismasystems.go:253-271`, `if val != nil`). So null = leave unchanged, and there's no way to null out a value that already exists.
- set `hva="Yes"` (204) -> `PUT hva:null` (204) -> GET still `"Yes"`.

**3. Clearing stores `""`, not NULL.**
A cleared input is `""`, and the backend writes it verbatim. So "unknown" (NULL) and "blank" (`""`) diverge.
- `PUT hva:""` (204) -> GET returns `""` (not null).

## Root cause

The frontend sends the full system object on every save, but the backend write path was built for partial PUTs (the non-nil guard in `Save`'s UPDATE branch), and nothing normalizes `""` -> NULL. `validate()` also treats an empty-string email as invalid rather than unset.

## Fix direction (backend-only)

- Normalize `""` -> nil for the optional string fields before `validate()`.
- Write the metadata fields authoritatively on UPDATE (drop the non-nil guard). `Save` has a single caller (the admin endpoint); the frontend sends the full object with stored values (null only when a field is already null), so authoritative write is safe and idempotent. The scoped-admin `copyHHSMetadata` / `clearHHSMetadata` paths are unaffected (they set the fields before Save).

## Acceptance criteria

- Blanking `datacallcontact` / `issoemail` clears them to NULL, no 400.
- A metadata field can be cleared to NULL through the edit form.
- Cleared optional fields store NULL, not `""`.
- A normal save does not wipe importer / scoped-admin data.
- emberfall e2e updated if it pins partial-PUT behavior.

Refs #431.

voidspooks:
Closes #442.

## Problem

The admin system add/edit save (`POST/PUT /fismasystems`) mishandled empty values on the optional metadata fields. Three symptoms, one root cause:

1. **Blanking an email field 400'd the whole save** — a cleared `datacallcontact`/`issoemail` arrives as `""`, which reached `isValidEmail("")` (false) and failed validation.
2. **A set metadata field couldn't be cleared** — the UPDATE path wrote only non-nil fields, so `null` meant "leave unchanged" and there was no way to un-set a value.
3. **Clearing stored `""`, not NULL** — a cleared input was written verbatim, so a column's "unknown" (NULL) and "blank" (`""`) states diverged.

Root cause: nothing normalized `""`, and the write path couldn't tell "clear this" from "leave unchanged."

## Fix (backend-only)

A three-state write contract via a small `blankToNil` helper:

- **`""` (a cleared form input) → write NULL** (clears the value)
- **`nil` (field omitted or `null`) → leave the stored value unchanged** — so a partial PUT never wipes importer data (preserves the existing emberfall Test A guard), and a scoped-admin edit whose HHS fields `copyHHSMetadata` set to stored values is a no-op
- **a value → write it**

Contact emails collapse `""` → nil before validation (so a blanked email is *unset*, not invalid); INSERT and UPDATE run the metadata fields through `blankToNil`.

No frontend change required — normalizing on the backend makes the existing form behavior correct.

## Why this shape

The obvious "write everything authoritatively" approach would have broken **emberfall Test A**, which deliberately pins that an OWNER partial PUT with HHS fields *omitted* must not wipe stored values. The `nil = leave unchanged` / `"" = clear` split fixes all three bugs while keeping that contract.

## Testing

- New unit tests: `TestBlankToNil`, `TestFismaSystem_Validate_BlankEmails`.
- New emberfall case (Test A2): clear-via-`""` writes NULL, a blanked email saves (204, cleared), and omitted HHS fields stay preserved.
- `make test-integration` green; `make test-e2e` green (155/0). Pre-commit + pre-push hooks green.
- Verified live against the dev API: blank email → 204 + NULL; `hva:""` → NULL; partial PUT (hva omitted) → preserved; `hva:null` → unchanged.

## Context

Surfaced while reviewing the metadata-standardization epic (#431); this blocks that epic's `cloud_system = No ⇒ clear cloud_service_model/cloud_vendor` rule and its "preserve unknown (null) end to end" criterion, so it lands first.

> Snyk / secret-dependent CI will show red on this fork PR (no repo secrets) — expected; maintainers re-run after pulling.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -612,6 +612,39 @@ tests:
           data:
             <<: *hhsExpected
 
+  # Test A2 — clearing values (ztmf#442). On the same system (still holding the
+  # HHS values Test A preserved), an OWNER PUT that sends hva as "" clears it to
+  # NULL, and a blanked issoemail no longer 400s the whole save. HHS fields still
+  # omitted from the body stay untouched (partial-PUT contract from Test A).
+  - url: "http://localhost:8080/api/v1/fismasystems/{{.createFismaSystemWithHHS.Response.data.fismasystemid}}"
+    method: PUT
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        <<: *updatedFismaSystemData
+        issoemail: "" # blanked email must not 400 (it clears to NULL)
+        hva: "" # cleared metadata field -> NULL
+    expect:
+      status: 204
+
+  - url: "http://localhost:8080/api/v1/fismasystems/{{.createFismaSystemWithHHS.Response.data.fismasystemid}}"
+    method: GET
+    headers:
+      <<: *commonHeaders
+    expect:
+      status: 200
+      headers:
+        content-type: "application/json"
+      body:
+        json:
+          data:
+            hva: null # "" cleared the column to NULL, not "" or unchanged
+            issoemail: null # blanked email cleared to NULL, no 400
+            system_owner: "Grand Moff Tarkin" # omitted HHS field preserved
+            cloud_system: "Yes" # omitted HHS field preserved
+
   # Test B — Scoped admin RBAC gate.
   #
   # Part 1: opDivAdmin POSTs a system with HHS fields → clearHHSMetadata
--- a/backend/internal/model/fismasystems.go
+++ b/backend/internal/model/fismasystems.go
@@ -182,6 +182,15 @@ func (f *FismaSystem) Save(ctx context.Context) (*FismaSystem, error) {
 
 	var sqlb SqlBuilder
 
+	// A blanked contact email arrives as "" from the edit form (a cleared input
+	// is "", not null). These are always written, so collapse "" to nil now:
+	// otherwise validate() below runs isValidEmail("") and 400s the whole save,
+	// and the column would store "" instead of NULL (ztmf#442). The metadata
+	// fields are handled per-path below, where "" (clear) must stay
+	// distinguishable from nil (omitted → leave unchanged).
+	f.DataCallContact = blankToNil(f.DataCallContact)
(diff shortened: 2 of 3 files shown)


jsos3-cms:
I can pick this one up and rehome it :)

jsos3-cms:
Rehomed to #444 so the Snyk/Deploy CI gates run — they can't on a fork PR. The commit was carried over unchanged with authorship preserved, and review is requested there from @danielbowne and @MackOverflow. Keeping this PR open as a draft for now so it can't accidentally merge; it'll be closed once #444 is merged. Thanks @voidspooks!

jsos3-cms:
Merged in 444

PR #446, opened 2026-07-18
Nightly sync: main -> impl (2026-07-18)

jsos3-cms:
Automated nightly sync of `main` into `impl` — **main wins every overlap**.

This branch is `impl` with `main` merged in using `-X theirs`, so the result is deterministic: main's version wins any conflict, and impl-only files that don't conflict are preserved. There should be nothing to hand-resolve.

**Squash-merge this PR** — `impl` is squash-only by policy (linear history). Because `main` and `impl` share no squash ancestry the diff is computed against a stale base, but `-X theirs` makes it deterministic and the content settles (a no-change night opens no PR). Unlike ztmf-ui (which merges as a merge commit), this stays in content-sync rather than closing the ancestry gap.

- Review the net change, then **squash-merge**. Merging pushes to `impl` and triggers the impl deploy (orchestration-impl: backend image build + terraform apply).
- To smoke-test in the impl environment first, run **Manual Deploy to IMPL** with `--ref automation/sync-main-to-impl`.

_Opened by the nightly-impl-sync workflow._


danielbowne (approved):
lgtm

PR #447, opened 2026-07-21
Surface silent score-rollover failures on data call creation

### LINKED ISSUE #411: copyPreviousScores silently swallows errors, can open a new data call with zero carried-over answers

danielbowne:
## Summary

`copyPreviousScores` rolls the previous cycle's answers forward into a newly created data call. It was a fire-and-forget goroutine that swallowed every error and never checked how many rows it copied, so a single FK-inconsistent row in the previous cycle could abort the whole `INSERT ... SELECT`, get logged anonymously, and open the new cycle with **zero carried-over answers** - indistinguishable from success to the API caller and to operators. Every system in the new cycle would appear to have made no progress.

Fixes #411.

## Changes

- **Surface failures.** `copyPreviousScores` now returns `(int64, error)` and runs synchronously. The "no previous cycle" case (first-ever data call) is treated as benign; any real error, a zero copy against a non-empty previous cycle, or a partial copy now emits a distinctive `ROLLOVER_ANOMALY` log line with the expected/copied counts.
- **Resilient copy.** The source `SELECT` now `INNER JOIN`s the `fismasystems` and `functionoptions` FK parents, so a score whose parent no longer resolves is dropped from the batch instead of aborting the entire rollover. The row count actually copied is compared against the count of candidate rows to detect a partial or empty result.
- **Insert-only gate.** The copy now runs only on the insert path (`d.DataCallID == 0`). Previously it fired after both the INSERT and UPDATE branches of `DataCall.Save`, so editing an existing data call re-ran the full copy and duplicated every carried-over score. The create is not failed on a copy error - the data call row is already committed and is valid without a rollover - but the outcome is now observable.
- **Request context.** The copy now runs on the caller's `context.Context` rather than `context.TODO()`, so a cancelled create request cancels the copy and releases its pooled connection.
- **Alerting.** A CloudWatch log metric filter + alarm on the `ROLLOVER_ANOMALY` token, routed to the existing `ztmf_alarms` SNS topic (same pattern as the login/OIDC alarms, which are also Go paths with no self-emitter).

### Automated

- `make test-integration` - updated the existing happy-path assertion to check the returned copied-count, and added `TestCopyPreviousScoresEmptyPreviousIntegration` covering the benign empty-previous-cycle path (zero rows, no error, no false anomaly). Green on the merged tree (includes #437's per-answer status changes).
- `terraform fmt -check` and `tflint` pass on the new monitoring file.

### Manual end-to-end against a production-shape database

Exercised through the real `POST /api/v1/datacalls` endpoint against a local database loaded with production-shape data (192,782 scores across 6 existing data calls), with an OWNER-role token. Created a new data call whose deadline falls after the latest existing cycle, so the previous cycle (38,400 scores) was the rollover source.

| Check | Result |
| --- | --- |
| Rollover copied from the previous cycle | 38,400 copied, exactly matching the 38,400 expected - a full clean copy, no rows dropped by the FK join |
| Per-answer `status` reset on carried rows | all 38,400 rows landed as `not_started`, set atomically within the copy |
| Answer values preserved | non-null notes carried forward on all 38,400 rows |
| `ROLLOVER_ANOMALY` emitted on a clean copy | none - the expected-vs-copied check does not false-alarm |
| Insert-only gate | editing the new data call via `PUT` did not re-fire the copy (count stayed 38,400; the previous unconditional behavior would have doubled it to 76,800) |

The test data call was removed afterward and the database returned to its exact baseline. The error and partial-copy branches cannot be reproduced against a real schema (the `NOT NULL` FKs forbid the orphaned-row state), so they remain covered by the integration tests and the reasoning above rather than a live run.

## Notes

- The dropped-dead-FK path in the resilient copy is defense-in-depth against a state the schema currently prevents (`scores.fismasystemid` / `functionoptionid` are `NOT NULL` FKs with no `ON DELETE` cascade, so the DB refuses to orphan a score). It cannot be reproduced in a test without tearing down the constraints; the happy-path test proves the JOIN does not drop valid rows, and the empty-previous test proves the expected-vs-copied accounting does not false-alarm.
- Wrapping the data-call insert and the rollover copy in a single transaction (so they are truly atomic) is deliberately out of scope - it would require reworking the shared `queryRow`/`recordEvent` connection pattern, and a data call is semantically valid without a rollover.

### DIFF
--- a/backend/internal/model/datacalls.go
+++ b/backend/internal/model/datacalls.go
@@ -79,10 +79,15 @@ func FindDataCallByID(ctx context.Context, dataCallID int32) (*DataCall, error) 
 	return queryRow(ctx, sqlb, pgx.RowToStructByName[DataCall])
 }
 
-func findPreviousDataCall(dataCallID int32) (*DataCall, error) {
+func findPreviousDataCall(ctx context.Context, dataCallID int32) (*DataCall, error) {
 	// find the *previous* datacall by deadline (not datacallid): once historical
 	// calls are loaded, a backfilled year can out-id the real prior call, so
 	// ordering by datacallid would pick the wrong "previous" for score rollover.
+	//
+	// Known limitation (ztmf#448): this picks the globally-latest OTHER call, not
+	// the latest call with a deadline earlier than this one. Creating a backfill
+	// data call with a deadline before existing cycles resolves "previous" to a
+	// future cycle. Out of scope here; tracked separately.
 	prevDcSqlb := stmntBuilder.
 		Select(dataCallColumns...).
 		From("datacalls").
@@ -90,7 +95,7 @@ func findPreviousDataCall(dataCallID int32) (*DataCall, error) {
 		OrderBy("deadline DESC", "datacallid DESC").
 		Limit(1)
 
-	return queryRow(context.TODO(), prevDcSqlb, pgx.RowToStructByName[DataCall])
+	return queryRow(ctx, prevDcSqlb, pgx.RowToStructByName[DataCall])
 }
 
 func FindLatestDataCall(ctx context.Context) (*DataCall, error) {
--- a/backend/internal/model/datacalls.go
+++ b/backend/internal/model/datacalls.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"context"
+	"log"
 	"strings"
 	"time"
 
@@ -45,7 +46,18 @@ func (d *DataCall) Save(ctx context.Context) (*DataCall, error) {
 		return nil, err
 	}
 
-	go copyPreviousScores(dataCall.DataCallID)
+	// Roll the previous cycle's answers into a *newly created* data call only.
+	// Never on update: re-running the copy on an edit would duplicate every
+	// carried-over score (ztmf#411). Run it synchronously so the outcome is
+	// observable, but do not fail the create on a copy error - the datacall row
+	// is already committed and is valid without a rollover (the first-ever cycle
+	// legitimately copies zero rows). copyPreviousScores emits the loud
+	// ROLLOVER_ANOMALY signal on any zero/partial/errored copy.
+	if d.DataCallID == 0 {
+		if _, err := copyPreviousScores(ctx, dataCall.DataCallID); err != nil {
+			log.Println(err)
+		}
+	}
 
 	return dataCall, nil
 }
--- a/backend/internal/model/scoreprogress_integration_test.go
+++ b/backend/internal/model/scoreprogress_integration_test.go
@@ -99,7 +99,9 @@ func TestFindScoreProgressIntegration(t *testing.T) {
 	`, fismaSystemID, functionOptionID, prevDC, notes)
 	require.NoError(t, err)
 
-	copyPreviousScores(newDC)
+	if _, err := copyPreviousScores(ctx, newDC); err != nil {
+		t.Fatalf("copyPreviousScores: %v", err)
+	}
 
 	findForSystem := func() *ScoreProgress {
 		t.Helper()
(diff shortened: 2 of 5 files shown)


MackOverflow (dismissed):
The fix is correct, the reasoning is thorough, and the risk is well-contained (create never fails on copy error). None of the items above are blocking; two nits, not blockers, but worth thinking about: 

1. (Minor) The "runs on caller's context" claim is only half-true.
copyPreviousScores now takes ctx, but its first DB query — findPreviousDataCall (datacalls.go:93) — still hardcodes context.TODO(). A cancelled create request would not cancel that lookup. It's a cheap single-row query, so low impact, but if the goal is honoring cancellation, findPreviousDataCall should accept and use ctx too.

2. (Minor / pre-existing, now more consequential) findPreviousDataCall picks by deadline, excluding the new call.
It orders deadline DESC excluding dataCallID and takes the top row. If someone creates a backfill data call with an earlier deadline than existing cycles, "previous" resolves to a cycle with a later deadline (a future cycle). Not introduced by this PR, but now that the copy is synchronous and alarmed, a mis-resolved "previous" is more visible/impactful. Worth a tracking note even if out of scope.

danielbowne:
Thanks for the review. Addressed both in c5af6a3:

**Nit 1 (ctx claim only half-true).** Right - `findPreviousDataCall` was still on `context.TODO()`, so the "runs on caller's context" claim didn't hold for the first query. It has a single caller (`copyPreviousScores`), so I threaded `ctx` through it. A cancelled create request now also cancels the previous-cycle lookup, and the claim is fully accurate.

**Nit 2 (deadline resolution picks a future cycle for a backfill).** Agreed it's out of scope for this PR but more consequential now that the copy is synchronous and alarmed. Filed #448 to track it and added a `Known limitation` note pointing at that issue in the `findPreviousDataCall` comment, so the next person to touch it sees the constraint.

Re-requesting review since the push stales the prior pass.

MackOverflow (approved):
lgtm

PR #450, opened 2026-07-21
Resolve previous data call by deadline-before, not globally-latest

### LINKED ISSUE #448: findPreviousDataCall can resolve a future cycle as 'previous' for a backfill data call
### Description

`findPreviousDataCall` (`backend/internal/model/datacalls.go`) resolves the "previous" data call for score rollover by ordering all other data calls `deadline DESC` and taking the top row - i.e. the globally latest data call other than the one being created. It does not restrict to calls whose deadline is *earlier* than the new call's.

For the normal path (a new cycle with the furthest-future deadline) this is correct. But if someone creates a **backfill** data call with a deadline earlier than existing cycles, "previous" resolves to a cycle with a *later* deadline - a future cycle - and its answers get rolled forward into the backfill.

This is pre-existing behavior, not introduced by #447. It is called out here because #447 makes the rollover synchronous and adds a `ROLLOVER_ANOMALY` alarm, so a mis-resolved "previous" is now more visible and more impactful than when the copy was a silent goroutine.

### Expected Behavior

"Previous" should be the most recent cycle with a deadline strictly earlier than the new call's deadline (or a deliberately chosen, documented ordering), so a backfill does not pull a future cycle forward.

### Actual Behavior

`ORDER BY deadline DESC, datacallid DESC LIMIT 1` excluding only the new call's own id, so the globally-latest other cycle is selected regardless of whether it precedes or follows the new call.

### Acceptance Criteria

- Creating a data call whose deadline is earlier than existing cycles either resolves "previous" to the correct earlier cycle, or rolls nothing forward - never a future cycle.
- Decide and document the intended semantics of "previous" when deadlines are non-monotonic with creation order.

### Severity

Low - requires an out-of-order backfill to trigger; normal forward cycle creation is unaffected. Latent correctness issue worth fixing before more historical backfills are loaded.

danielbowne:
## Summary

`findPreviousDataCall` resolves the "previous" data call whose answers get rolled forward into a newly created cycle (via `copyPreviousScores`). It ordered all other data calls by deadline and took the top row - i.e. the globally-latest cycle other than the target.

That is correct for a normal new cycle (the one with the furthest-future deadline), but wrong for a **backfill**: creating a data call with a deadline earlier than existing cycles resolved a *future* cycle as its "previous" one and rolled that future cycle's answers backward into the backfill.

Fixes #448.

## Change

Restrict the lookup to cycles whose deadline is strictly earlier than the target call's own deadline:

```
WHERE deadline < (SELECT deadline FROM datacalls WHERE datacallid = ?)
ORDER BY deadline DESC, datacallid DESC
LIMIT 1
```

- A normal forward cycle still resolves the true prior cycle (same result as before for the non-backfill case).
- A backfill resolves the correct earlier cycle, or none - and "none" is already benign in `copyPreviousScores` (`ErrNoData` returns `(0, nil)`, nothing to roll forward).
- The strictly-less comparison also excludes the call itself, so no separate self-exclusion is needed.

## Testing

- `make test-integration` - added `TestFindPreviousDataCallByDeadlineIntegration`, which pins the fix: a backfill whose deadline sits between two cycles resolves to the earlier one (and explicitly not the later/future one - this assertion fails on `main`), the normal forward case still resolves the real prior cycle, and a call earlier than every cycle resolves to `ErrNoData`. All existing rollover and latest-cycle tests stay green.

## Notes

Surfaced during review of #447 (which made the rollover synchronous and alarmed, so a mis-resolved "previous" became more visible). This is the tracked follow-up.

### DIFF
--- a/backend/internal/model/datacalls.go
+++ b/backend/internal/model/datacalls.go
@@ -80,18 +80,26 @@ func FindDataCallByID(ctx context.Context, dataCallID int32) (*DataCall, error) 
 }
 
 func findPreviousDataCall(ctx context.Context, dataCallID int32) (*DataCall, error) {
-	// find the *previous* datacall by deadline (not datacallid): once historical
-	// calls are loaded, a backfilled year can out-id the real prior call, so
-	// ordering by datacallid would pick the wrong "previous" for score rollover.
+	// find the *previous* datacall: the most recent cycle whose deadline is
+	// strictly earlier than this call's deadline. Ordering is deadline-driven
+	// (not datacallid) because historical loads can carry a higher datacallid
+	// than the real prior call.
 	//
-	// Known limitation (ztmf#448): this picks the globally-latest OTHER call, not
-	// the latest call with a deadline earlier than this one. Creating a backfill
-	// data call with a deadline before existing cycles resolves "previous" to a
-	// future cycle. Out of scope here; tracked separately.
+	// The "strictly earlier than this call's deadline" restriction (not merely
+	// "the globally latest other call") is what fixes ztmf#448: a backfill data
+	// call can be created with a deadline BEFORE existing cycles, and picking the
+	// globally-latest other call would then resolve a future cycle as its
+	// "previous" and roll that future cycle's answers backward. Anchoring to
+	// deadlines before this call's makes a normal new cycle pick the real prior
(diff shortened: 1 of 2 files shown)


MackOverflow (approved):
clean, lgtm!

Issue #239, opened 2026-01-13
Track wrong Data Center Environments


jsos3-cms:
We should capture a list of FISMA systems with the wrong data center environment so that we can get around to fixing them in bulk once we've found more.

jsos3-cms:
This is supposed to be supported by the CFACTS Sync. Complete getting that data set into SDL, then have this be one of the usecases.

voidspooks:
Working this. Need to sync with Scott.

scott-aquia:
Met with Cameron to discuss; currently using Claude to develop/tweak SQL statement.
I would like to clone the DB table in Workbench for testing purposes

eschwein-usds:
Known Systems with incorrect Data Centers:
- HEDIS Patient Data system; listed as ‘CMS-Cloud-AWS’ but it should be ‘CMS-Cloud-MAG’
- Most CCSQ systems moved from generic AWS to ‘CMS-Cloud-AWS’ since the last data call
- AGX is now on Sharepoint, so SaaS instead of CMSDC
- Maybe Open Payments 2 OPS 2.0?
- VMS should be CMSDC (need to check), not AWS
- FISS and MCS need to change from OPDC to CMSDC

eschwein-usds:
And APPS is now on AWS, not CMSDC

voidspooks:
Status update, split per Jono's list above.

**Done in code (draft PRs):**
- Backend: CMS-Enterprise/ztmf#452 adds `GET /api/v1/datacentermismatches`, an admin-tier report of active systems whose CFACTS-reported data center environment disagrees with the value recorded in ZTMF. It compares the enrichment payload key `data_center_environment` against `fismasystems.datacenterenvironment` (case-insensitive), gated on the OpDiv `insights_enabled` flag like the rest of ZTMF Insights, decommissioned systems excluded. Each row also flags whether the CFACTS value exists in our `datacenterenvironments` vocabulary, since an unknown value needs a mapping row before the system can be corrected. Covers item 1 and the detect-and-notify half of item 3.
- Frontend: CMS-Enterprise/ztmf-ui#590 displays the CFACTS data center environment on the System Detail page (ZTMF Insights card) with a warning flag when it differs from the system's own value. Covers item 2.

Both are safe to merge ahead of the data: the payload key is provisional and the report/display stay empty until the pipeline ships the field, and renaming the key later is a one-line change in each repo.

[...]

PR #452, opened 2026-07-21
feat(#239): add /datacentermismatches wrong-data-center report

voidspooks:
## Summary

Adds `GET /api/v1/datacentermismatches`, an admin-tier report of active FISMA systems whose CFACTS-reported data center environment disagrees with the value recorded in ZTMF (#239, items 1 and 3 in detect-and-notify form).

- Compares the `system_enrichment` payload key `data_center_environment` (provisional until the ztmf-insights pipeline ships the field; a single constant in the model, so a rename is a one-line change) against `fismasystems.datacenterenvironment`, trimmed and case-insensitive. Uses `IS DISTINCT FROM` so a system with no recorded value still reports when CFACTS has one.
- Gated on the owning OpDiv's `insights_enabled` flag, same as the enrichment read. Decommissioned systems and enrichment rows without the payload key are excluded, so the report stays empty rather than wrong until the upstream data lands.
- RBAC mirrors `/insights`: unscoped admin tiers see everything, OpDiv tiers are scoped fail-closed, ISSO/ISSM get 403 (the ISSO-facing reporting flow is CMS-Enterprise/ztmf-ui#604, deliberately out of scope here).
- Each row carries `cfacts_value_known`: whether the CFACTS value exists in the `datacenterenvironments` vocabulary (#392). An unknown value needs a mapping row before the system value can be corrected, so those rows need different handling than a simple wrong value.

Companion UI change: CMS-Enterprise/ztmf-ui (draft PR linked from CMS-Enterprise/ztmf#239).

## Testing

- `make test-unit`, `make test-integration`, and `make test-e2e` all green locally (emberfall 158/158).
- New integration coverage: seeded mismatch visibility, match exclusion, decommissioned exclusion, OpDiv insights gate, unknown-vocabulary flagging, case/whitespace insensitivity, NULL ZTMF value, fail-closed OpDiv scoping.
- Emberfall: unscoped admin and HHS_READONLY_ADMIN see the staged rows; ISSO gets 403. The admin assertion expects two rows because an earlier suite test PUTs system 1002's environment to "Other", which turns the seeded match into a live mismatch; a comment in the yml explains the dependency.
- Endpoint also driven manually against the dev stack: admin JWT returns the seeded Shield Gen mismatch, ISSO JWT gets 403.
- `openapi.yaml` regenerated via `make generate-openapi`; redocly lint clean.

Note: `make test-unit` on a clean checkout currently fails without DB env vars sourced because the migrations package connects to the DB at package init and CMS-Enterprise/ztmf#415 added a test file that triggers it. Pre-existing and unrelated to this change; will file separately.

Snyk and other secret-dependent checks will show red as usual for fork PRs (no repo secrets).

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -562,19 +562,48 @@ ON CONFLICT (idm_name) DO NOTHING;
 -- Shield Gen (1003), which the Emberfall ISSO is assigned to. Gives the
 -- /systemenrichment endpoint E2E coverage via the users_fismasystems join.
 -- The payload is opaque to ztmf core (owned by the enrichment pipeline).
+-- data_center_environment ('CMS-Cloud-AWS', a canonical vocabulary value from
+-- migration 0045) deliberately disagrees with the system's own value
+-- ('Forest-Moon'), making this THE visible /datacentermismatches row (issue
+-- #239) with cfacts_value_known = TRUE.
 INSERT INTO public.system_enrichment (fisma_uuid, payload, synced_at) VALUES (
     'E1D00198-36D4-4EAB-8C00-501E1D000999',
-    '{"fisma_acronym":"SLD-GEN","cfacts":{"lifecycle_phase":"Operational","fips_impact_level":"Moderate"},"scoring":{"suggested_score":2,"suggested_label":"Initial","evidence_sources":["Kion","Hardenize"]}}',
+    '{"fisma_acronym":"SLD-GEN","data_center_environment":"CMS-Cloud-AWS","cfacts":{"lifecycle_phase":"Operational","fips_impact_level":"Moderate"},"scoring":{"suggested_score":2,"suggested_label":"Initial","evidence_sources":["Kion","Hardenize"]}}',
+    '2026-05-20 00:00:00+00'
+) ON CONFLICT (fisma_uuid) DO NOTHING;
+
+-- Enrichment for the Executor (1002, EMPIRE, active): CFACTS agrees with the
+-- system's own value ('Imperial-Fleet'), so /datacentermismatches must exclude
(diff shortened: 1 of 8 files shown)


jsos3-cms:
@voidspooks — one blocking issue before this lands; everything else looks solid (RBAC scoping, mismatch semantics, the FE contract match, nil-safety, and the rest of the tests all check out).

**Cross-OpDiv data leak via the non-unique `fismauid` join.** `datacentermismatches.go` does `INNER JOIN fismasystems fs ON LOWER(fs.fismauid) = LOWER(se.fisma_uuid)`, but `fismasystems.fismauid` has no unique constraint — this repo documents that exact hazard in `systemenrichment.go` (uses `EXISTS`), `FindFismaSystemByUUID` (uses `LIMIT 1`), and migration 0038. If two systems share a `fismauid`, the single enrichment row fans out and emits a phantom mismatch for the wrong system, which can surface a finding built from another OpDiv's data to a caller scoped elsewhere.

Suggest fixing on the in-repo rehome (#453, where CI can run): swap the join for an `EXISTS` / lateral-`LIMIT 1` dedup mirroring `systemenrichment.go`, and add a regression test with two `fismasystems` rows sharing a `fismauid` (no current fixture covers it). Once that's in, this reads good to go.

jsos3-cms:
The fan-out fix landed on the in-repo rehome CMS-Enterprise/ztmf#453 (`INNER JOIN LATERAL … LIMIT 1` + a duplicate-`fismauid` regression test), and it's marked ready there. Thanks @voidspooks!

jsos3-cms:
Merged in-repo via CMS-Enterprise/ztmf#453 (squash `cc2fd415`, with the cross-OpDiv join fix). Closing this fork PR. Thanks @voidspooks!

PR #453, opened 2026-07-21
feat(#239): add /datacentermismatches wrong-data-center report

jsos3-cms:
> **Note:** In-repo copy of #452 by @voidspooks, moved from the fork so Snyk and the deploy CI gates can run (those require org secrets unavailable to fork PRs). Commits and authorship are unchanged.
>
> Refs #452

---

## Summary

Adds `GET /api/v1/datacentermismatches` — an admin report of systems whose CFACTS-reported data center environment (the `data_center_environment` enrichment key) disagrees with the system's own recorded `datacenterenvironment`. The comparison is trim + case-insensitive; it excludes decommissioned systems and OpDivs without `insights_enabled`, and RBAC mirrors `/insights` (admin-tier only, fail-closed OpDiv scoping, ISSO/ISSM → 403). Companion to the frontend chip in ztmf-ui#590. Part of #239.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -562,19 +562,48 @@ ON CONFLICT (idm_name) DO NOTHING;
 -- Shield Gen (1003), which the Emberfall ISSO is assigned to. Gives the
 -- /systemenrichment endpoint E2E coverage via the users_fismasystems join.
 -- The payload is opaque to ztmf core (owned by the enrichment pipeline).
+-- data_center_environment ('CMS-Cloud-AWS', a canonical vocabulary value from
+-- migration 0045) deliberately disagrees with the system's own value
+-- ('Forest-Moon'), making this THE visible /datacentermismatches row (issue
+-- #239) with cfacts_value_known = TRUE.
 INSERT INTO public.system_enrichment (fisma_uuid, payload, synced_at) VALUES (
     'E1D00198-36D4-4EAB-8C00-501E1D000999',
-    '{"fisma_acronym":"SLD-GEN","cfacts":{"lifecycle_phase":"Operational","fips_impact_level":"Moderate"},"scoring":{"suggested_score":2,"suggested_label":"Initial","evidence_sources":["Kion","Hardenize"]}}',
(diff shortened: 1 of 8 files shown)


jsos3-cms:
Pushed the fan-out fix (`0e5c4a0`): the `fismasystems` join is now `INNER JOIN LATERAL (… ORDER BY fismasystemid LIMIT 1)`, so a `fismauid` shared across systems can no longer fan one enrichment row out across OpDivs. Added a `DuplicateFismauidDoesNotFanOut` integration regression test; the full datacentermismatches integration suite + build/vet/staticcheck are green. Marking ready — @danielbowne @MackOverflow ready for another look.

MackOverflow (approved):
A few small nits: 

Case-sensitivity mismatch between cfacts_value_known and write-time validation. Here cfacts_value_known matches the reference table with LOWER(d.datacenterenvironment) = LOWER(cfactsDC) (case-insensitive), but dataCenterEnvironmentExists (used when saving a system) does an exact case-sensitive col = $1. A CFACTS value differing only in case from a vocabulary entry would report cfacts_value_known = true here, yet fail write validation. Likely intentional (different semantics), but worth a one-line note so a future reader doesn't assume the two checks agree.
The JSON-extract expression is re-evaluated 4×. TRIM(se.payload->>'data_center_environment') appears in the SELECT, the cfacts_value_known subquery, and both WHERE clauses. Functionally fine and negligible at report scale (hundreds of systems), but a wrapping subquery/CTE that computes it once would read more cleanly. Optional.
opdiv_id typed as *int32. In practice the INNER JOIN opdivs guarantees non-NULL, so the pointer is defensive-only. Harmless; just noting it can never actually serialize null given the join.
No 500 failure-mode assertion, but that mirrors the rest of the codebase and isn't practical to unit-test here.

MackOverflow (dismissed):
lgtm

jsos3-cms:
Filed #456 to track the two non-blocking follow-ups from review: the split-OpDiv duplicate-`fismauid` under-report edge case (fails safe today) and adding a `LOWER(fismauid)` index.

MackOverflow (approved):
lgtm

PR #458, opened 2026-07-21
Add System Delegate role for contractor/support-staff data-call access

### LINKED ISSUE #455: Add a "System Delegate" role to support contractor/support-staff access for data calls
## Background

The #1 piece of feedback on ZTMF access (primarily from FDA and NIH) is difficulty getting enough people access to the tool to complete data calls. Today we offer an ISSO role and an ISSM role; ISSM only exposes system details for the systems it's granted on.

Without a low-friction way to grant access to contractors and supporting teammates, OpDivs tend to farm data collection out to spreadsheets instead of working in the tool.

## Proposal

Add a new **System Delegate** role (rather than renaming ISSM):

- Read/**write** access scoped to specific systems — delegates must be able to edit data call answers directly, otherwise the ISSO still has to re-enter everything
- Intended for contractors and support staff; Admins should be encouraged to treat it as a secure role to grant for data call completion
- Leave the existing ISSM role unchanged

## Why a new role instead of renaming ISSM

- ISSM is planned to evolve with future ABAC work: the goal is to grant it to Component-level ISSOs so they automatically get all systems belonging to their component (currently assigned by hand). Renaming it would force creating a new ISSM-like role later.
- A distinct role name makes tracking/logging more consistent — we can tell whether an action was performed by an ISSO or a delegate.

## Acceptance criteria

- [ ] New "System Delegate" role exists, assignable per system
- [ ] Role grants read/write on assigned systems' data call answers (otherwise locked down)
- [ ] Existing ISSM role and assignments are unchanged
- [ ] Role actions are distinguishable from ISSO actions in logs/tracking
- [ ] Training documents are updated to reflect the new role
- [ ] A "how-to guide" for Admins is created covering how to add all the roles, with a breakdown of each role's permissions

## Context

From a Slack discussion between Mackenzie Wartenberger and Elizabeth Schweinsberg (agreed outcome: "if adding a role isn't too hard, then let's add a system delegate role").

danielbowne:
> **Paired deployment.** This backend PR and the companion ztmf-ui PR (role picker + UI gating) are meant to ship together. Merging either to `main` auto-deploys it. Backend-alone is non-breaking - the API accepts the new role, it just is not selectable in the UI until the ztmf-ui PR lands - so do not mark this ready for merge until the ztmf-ui PR is ready to go out alongside it.

## Summary

Adds a new system-scoped user role, `SYSTEM_DELEGATE`, so contractors and support staff can be granted read/write access to specific FISMA systems' data-call answers without inheriting any of the ISSO/ISSM system-configuration surface.

Closes #455 (backend portion).

## Background

FDA and NIH reported they could not grant enough people access to complete Zero Trust data calls under the existing ISSO/ISSM roles, so teams fell back to spreadsheets. This role fills that gap. ISSM is intentionally left untouched (it is reserved for later ABAC auto-grant work), which is why this is a distinct role rather than a change to ISSM.

## What it does

The delegate's RBAC is identical to ISSO/ISSM - system-scoped via `users_fismasystems`, gated by `IsAssignedFismaSystem` for score reads/writes and data-call completion - with one deliberate carve-out: a delegate is forbidden from writing a system's target maturity, because that is the ISSO's risk assertion (#398), not a data-call answer.

- Register `SYSTEM_DELEGATE` in the role validation map
- Add an `IsSystemDelegate` helper and allow `OPDIV_ADMIN` and up to assign the role (non-escalating - same tier ceiling as ISSO/ISSM)
- Forbid delegates from `SaveFismaSystemTargetMaturity`
- Unit + controller coverage for the helper, assignability, and the carve-out
- Empire seed delegate user + Emberfall access-matrix block (answers 200, target maturity 403, admin surface 403)

## Notes

- **No schema migration.** `users.role` is free-text `VARCHAR(30)` gated by the app validation map; the only constraint is the legacy denylist, which the new value does not violate.
- Delegate actions are distinguishable from ISSO in the audit trail via the stored role value (AC #4).
- The AC set is split: this PR is backend only. The frontend role picker + UI gating land in a separate ztmf-ui PR, and the two documentation ACs (training material, admin how-to guide) are handled outside code. Both PRs are intended to ship together.

## Testing

- `make test-unit` - pass
- `make test-e2e` (Emberfall, ephemeral) - Failed: 0, including the new delegate access-matrix block
- OpenAPI regenerated - no drift

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -94,6 +94,14 @@ INSERT INTO public.users (userid, email, fullname, role, identity_provider)
     VALUES ('66666666-6666-6666-6666-666666666666', 'Isso.User@nowhere.xyz', 'ISSO Test User', 'ISSO', 'okta')
     ON CONFLICT DO NOTHING;
 
+-- System Delegate fixture (#455): contractor/support-staff tier, system-scoped
+-- like ISSO but answers-only (barred from target maturity). Assigned to Death Star
+-- (1001), which is enrolled in the open Audit cycle so E2E can exercise score writes.
+-- v4-conforming UUID so it satisfies isValidUUID when used as a path param.
+INSERT INTO public.users (userid, email, fullname, role, identity_provider)
+    VALUES ('55555555-5555-4555-8555-555555555555', 'Delegate.User@nowhere.xyz', 'System Delegate Test User', 'SYSTEM_DELEGATE', 'okta')
+    ON CONFLICT DO NOTHING;
+
 -- Pre-deleted user fixture for RestoreUser tests.
 -- UUID is v4-conforming (4 at position 14, 8 at position 19) so it satisfies
 -- isValidUUID's strict regex when used as a path param.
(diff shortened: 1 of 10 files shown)


MackOverflow (dismissed):
cleanish, one nit, not a blocker: 

1. Latent maintainability risk — the carve-out lives in exactly one place (low, worth noting).
delegate ⊆ answers is enforced by a single IsSystemDelegate() check in SaveFismaSystemTargetMaturity. There's no central guard or negative-space test asserting "a delegate can reach nothing an ISSO can that isn't a data-call answer." If a future PR adds another ISSO-writable non-answer surface, delegates silently inherit it. Consider a one-line pointer in the IsSystemDelegate doc comment listing the carve-out site(s), or a table-driven test enumerating delegate-forbidden endpoints, so the invariant fails loudly when it's next touched.

danielbowne:
Thanks @MackOverflow. Addressed the nit in cf3f1dc: added `TestSystemDelegate_ForbiddenNonAnswerSurfaces`, a table-driven tripwire enumerating delegate-forbidden non-answer endpoints (target maturity today), and documented the invariant plus its enforcement site on the `IsSystemDelegate` helper so the next surface that needs guarding fails loudly. Re-requesting your approval.

MackOverflow (approved):
clean, the UUID in the unit testing is different than the actual seed (4444 vs 5555) which is a little confusing, but its fine. otherwise this is great, thanks for the quick turnaround!

MackOverflow (dismissed):
lgtm

PR #460, opened 2026-07-21
adds guard on what systems a user can be assigned based on their opdivs

### LINKED ISSUE #449: Assignment write lets an unscoped admin grant a user a system outside their OpDiv
### Description

The user-to-system assignment write (`POST /api/v1/users/{userid}/assignedfismasystems`, `CreateUserFismaSystem`) enforces two OpDiv checks on the **acting admin**: `guardManageFismaSystem` (the admin manages the system's OpDiv) and `CanManageUser` (the admin shares an OpDiv with the target user). It never checks that the system's OpDiv is one the **target user** actually belongs to.

For an OpDiv-scoped admin (`OPDIV_ADMIN`) this gap is masked - they can only manage systems and users within their own OpDiv, so both are always the same OpDiv. But an unscoped admin (`OWNER` / `HHS_ADMIN`) satisfies both checks for every OpDiv, so they can assign a system from one OpDiv to a user who belongs to a different one - e.g. assign an FDA system to a CMS user. The assignment succeeds and the user ends up with access to a system outside their OpDiv.

### Steps to Reproduce

1. Authenticate as an `OWNER` (or `HHS_ADMIN`).
2. Pick a target user whose OpDiv is CMS.
3. `POST /api/v1/users/{cms_userid}/assignedfismasystems` with the `fismasystemid` of an FDA system.
4. The assignment is created - no error.

### Expected Behavior

The assignment is rejected (403/422) unless the system's `opdiv_id` is one of the target user's assigned OpDivs. A user can only be granted systems within their own OpDiv, regardless of the acting admin's tier.

### Actual Behavior

The write succeeds for an unscoped admin even when the system's OpDiv is not one the target user belongs to, creating a cross-OpDiv assignment.

### Acceptance Criteria

- The assignment write rejects any system whose `opdiv_id` is not in the target user's assigned OpDiv set.
- The existing `OPDIV_ADMIN` path is unchanged (already safe).
- Consider a scoped read endpoint - "systems assignable to user Y" (systems in the target user's OpDiv) - to back the frontend picker (paired UI issue), so the client is not filtering a global list. `FindFismaSystemsInput` already supports `OpDivIDs` / `RestrictToOpDivIDs`; this would resolve the target user's OpDivs into that filter rather than the acting caller's.
- Server-side enforcement stands on its own so the UI picker is a convenience, not the only guard.

### Severity

Medium - authorization / data-integrity gap. Requires an unscoped admin (OWNER/HHS_ADMIN) to trigger; scoped admins are already constrained.

MackOverflow:
> **Note:** In-repo copy of #457 by @costacalvin. Moved from the fork so Snyk and the deploy CI gates can run — those jobs require org secrets unavailable to fork PRs. Commits and authorship are unchanged. This copy also regenerates `backend/openapi.yaml` for the new endpoint (the fork PR omitted it, which fails the "openapi spec up to date" check).
>
> Refs #457

---

## Summary

Closes #449 — an unscoped admin (`OWNER`/`HHS_ADMIN`) could assign a FISMA system to a user in a **different OpDiv** (e.g. an FDA system to a CMS user). The assignment write validated the acting admin's authority over the system and the user, but never checked that the system's OpDiv is one the **target user** actually belongs to.

## Root cause

`CreateUserFismaSystem` enforced two checks on the acting admin — `guardManageFismaSystem` (admin manages the system's OpDiv) and `CanManageUser` (admin shares an OpDiv with the target). For an `OPDIV_ADMIN` these are always the same OpDiv, so the gap was masked. But an unscoped admin passes both for *every* OpDiv, so a cross-OpDiv assignment slipped through and the user gained access to a system outside their OpDiv.

## Changes

**Write enforcement (the fix)**
- New `User.CanBeAssignedFismaSystem(opdivID)` — reports whether a system in the given OpDiv may be assigned to this (target) user. Fail-closed: a nil OpDiv or a user with no OpDiv grants can hold no system.
- `CreateUserFismaSystem` now rejects (403) unless the system's `opdiv_id` is in the target user's assigned OpDiv set, **regardless of the acting admin's tier**. `guardManageFismaSystem` already fetched the system, so it's reused (no extra query).
- **Hardened against body override:** the path `userid` is re-pinned after JSON decode, so a request body carrying its own `"userid"` cannot retarget the write to a user other than the one the guards validated (this would otherwise reopen the gap).
- `DeleteUserFismaSystem` is intentionally left unchanged so an admin can still remove a stray/legacy cross-OpDiv assignment.

**Scoped read endpoint for the picker**
- New `GET /users/{userid}/assignablefismasystems` — returns systems in the target user's OpDivs (resolved into `FindFismaSystemsInput.OpDivIDs` + `RestrictToOpDivIDs`), intersected with the caller's own OpDivs when the caller is OpDiv-scoped, so the picker never offers a system the caller couldn't actually assign. Gated with the same `sharesOpDiv` read check the other per-user endpoints use. Backs the paired UI issue (ztmf-ui#586); server-side enforcement stands on its own.

## Behavior notes

- Enforcement is **strict fail-closed**: a user with **no OpDiv grant** (e.g. a not-yet-provisioned ISSO/ISSM) can be assigned **no** systems. Provisioning must grant an OpDiv before assigning systems. This only affects the *create* path — existing assignments are untouched, so no one loses access.
- ⚠️ **Frontend impact:** the provisioning flow must grant the OpDiv *before* assigning systems, or it will get a `403`. Tracked in ztmf-ui#586.

## Testing

- **Unit** — `TestUser_CanBeAssignedFismaSystem` (in-set / out-of-set / nil OpDiv / zero-grant), plus controller gate tests for the new read endpoint.
- **E2E (Emberfall)** — added cases: cross-OpDiv assign → `403`, in-OpDiv assign → `201`, body-`userid` override ignored (asserts `data.userid` == path user, using a valid-v4 UUID of a different real user so it exercises the real retarget path), and the read endpoint → `200`.
- **Seed** — granted the assignment-test ISSO fixture (Thrawn) an EMPIRE OpDiv so it represents a properly-provisioned user under the new invariant. One additive line; the fixture is used only by the assignment tests.

Verification: `go build ./...`, `go vet ./...`, `go test -short ./...`, and `make test-e2e` (**159 passed, 0 failed**).

## Acceptance criteria

- [x] The assignment write rejects any system whose `opdiv_id` is not in the target user's assigned OpDiv set.
- [x] The existing `OPDIV_ADMIN` path is unchanged (already safe).
- [x] Scoped read endpoint for "systems assignable to user Y".
- [x] Server-side enforcement stands on its own; the UI picker is a convenience, not the only guard.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -133,7 +133,12 @@ SELECT u.userid, (SELECT opdiv_id FROM public.opdivs WHERE code = 'EMPIRE')
         'Emperor.Palpatine@coruscant.empire',
         'Captain.Needa@executor.empire',
         'Opdiv.Admin@empire.test',
-        'Opdiv.Readonly@empire.test'
+        'Opdiv.Readonly@empire.test',
+        -- Thrawn is the assignment-test fixture. Since #449 an assignment write
+        -- rejects any system whose OpDiv the target does not hold (fail closed:
+        -- no OpDiv grant -> no assignable systems), so a properly-provisioned
+        -- ISSO must hold an OpDiv before systems can be assigned to them.
+        'Grand.Admiral.Thrawn@chiss.empire'
        )
 ON CONFLICT DO NOTHING;
 
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -450,6 +450,26 @@ func TestListUserFismaSystems_ISSOForbidden(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
 
+// --- ListAssignableFismaSystems ---
+
+func TestListAssignableFismaSystems_ReadonlyAdminAllowed(t *testing.T) {
+	r := httptest.NewRequest("GET", "/api/v1/users/11111111-1111-1111-1111-111111111111/assignablefismasystems", nil)
+	r = withUser(r, readonlyAdmin)
+	w := httptest.NewRecorder()
+
+	ListAssignableFismaSystems(w, r)
+	assert.NotEqual(t, http.StatusForbidden, w.Code)
+}
+
+func TestListAssignableFismaSystems_ISSOForbidden(t *testing.T) {
+	r := httptest.NewRequest("GET", "/api/v1/users/11111111-1111-1111-1111-111111111111/assignablefismasystems", nil)
+	r = withUser(r, issoUser)
+	w := httptest.NewRecorder()
+
+	ListAssignableFismaSystems(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
 // --- CreateUserFismaSystem ---
 
 func TestCreateUserFismaSystem_ReadonlyAdminForbidden(t *testing.T) {
--- a/backend/cmd/api/internal/controller/usersfismasystems.go
+++ b/backend/cmd/api/internal/controller/usersfismasystems.go
@@ -38,6 +38,73 @@ func ListUserFismaSystems(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, fismasystemids, err)
 }
 
+//	@Summary	List the FISMA systems assignable to a user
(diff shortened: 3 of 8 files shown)


a-dipietro (approved):
lgtm!

PR #463, opened 2026-07-22
Track time spent in ZTMF by system

costacalvin:
## Summary
Adds the logging and queries to measure how long people spend working a Data Call. The questionnaire records a `viewed` event when a user opens a question, and the metrics are delivered as **documented SQL** (`backend/docs/timespent_queries.sql`) — run against Postgres now, portable to Snowflake later.

**Closes #368** · Companion frontend PR: CMS-Enterprise/ztmf-ui#599

---

## Requirements

**Written acceptance criteria (issue #368):**
- [x] Sufficient logs to track how much time is spent on questions.
- [x] Queries for **total time per system**, **time per person per system**, and **average per question (within one system)**.

**Clarified in the issue comments (owner @eschwein-usds + collaborators):**
- [x] **Delivery = SQL queries/doc**, not an in-app report/download (Snowflake dashboard is a later step).
- [x] **This year:** a **save-gap proxy** over existing saves so results can be compared to last year; accurate view-based tracking is the "and" for going forward.
- [x] **Editors are the primary metric** (viewer data "useful down the line").
- [x] **30-minute idle cap** (the app session times out at 30 min).
- [x] Rough is fine — delivered as **raw seconds** (finer than the "nearest 5 min" suggestion, so it can be rounded downstream as needed).

---

## What this PR does
- **Logs question views** — `POST /api/v1/events/view` writes a `viewed` event (`resource='questionnaire'`, payload `{fismasystemid, datacallid, questionid}`). Editor-vs-viewer is **derived server-side** from the caller's role + the data-call deadline (never client-supplied); the data call must exist.
- **Ships the queries** — `backend/docs/timespent_queries.sql`, two modes × three rollups (per system / per person / per question):
  - **MEASURED** (view-based, this cycle forward): dwell = a question's view → the next view, capped at 30 min, editor-primary with viewer shown separately.
  - **PROXY** (save-gap, older data calls, for year-over-year): the gap before each save is attributed to that save's question (`LAG`), capped at 30 min, editor-only.
  - Plus an all-systems variant, Snowflake porting notes, and an optional performance-index DDL.
- **Removes** the in-app analytics/Excel/export layer from previous commit/impl (delivery is SQL, per the owner).
- **Integration test** runs the shipped SQL against a seeded DB so the queries stay verified as the schema evolves.

---

## Scope decisions

**Pulled INTO this PR** (were floated as possible follow-ups, included now):
[...]

PR #464, opened 2026-07-22
Track time spent in ZTMF by system

### LINKED ISSUE #368: Add ability to track time spent in ZTMF by system
### Feature Type

General Feature

### Workflow Description

For FY2026, it would be great if we had sufficient logs of active use such that we could have rough ideas of how long people spend on a system.

It would be fine to use existing system/event logs/database table to calculate this.
Do we have events for both the start of a person looking at a question and when they move on or save it?
Cumulative total for time spent in the Data Call period for a system is useful, and time spent per person, and average time spent per question.

### Acceptance Criteria

- [ ] We have sufficient logs to track how much time is spent on questions
- [ ] We have the queries to calculate total time spent per system, time spent per person for each system they edit, and average per question (within 1 system).

### Additional Context

_No response_

jsos3-cms:
This is the in-repo version of #463 by @costacalvin, moved onto an in-repo branch so the org-secret CI gates (Snyk, Infrastructure/Deploy) run and the change can deploy to dev on green. The commits are carried over unchanged and authorship is preserved; credit for the work is Calvin Costa's.

Refs #463 · Closes #368 · Frontend companion: CMS-Enterprise/ztmf-ui#601

## What this adds

Time-spent tracking for the ZTMF questionnaire (#368). A new `POST /api/v1/events/view` appends a `viewed` event when a user opens a questionnaire question, so analytics can bound how long each question was worked on before the next one is opened. Editor-vs-viewer is derived server-side from the caller's role and the data-call deadline — never sent by the client. The metrics ship as documented SQL in `backend/docs/timespent_queries.sql`: a MEASURED mode (dwell = a question's view to the next view, capped at 30 minutes, editor/viewer split) for this cycle forward, and a save-gap PROXY mode (editor-only) for year-over-year comparison against older data calls that predate view tracking. Each mode has three rollups (per system, per person, per question).

## Access control

A caller may only record a view for a system they can already see: unscoped-read admins and assigned ISSO/ISSM/SYSTEM_DELEGATE users short-circuit, and OpDiv-scoped admins are checked against the system's OpDiv. An unknown system or data call returns 404; a malformed body returns 400.

## Scope

Pulled in now (was floated as a possible separate ticket): the view-event instrumentation itself, so accurate data starts accruing immediately, and viewer-time capture alongside editor time (server-classified). Left as follow-ups: an in-app download/report UI, the Snowflake pipeline for the eventual dashboard, question-to-system applicability validation, rate-limiting/dedup of view events, separating high-volume view events from the audit log, and the optional performance index documented at the bottom of the SQL file.

## Verification

`go build ./...`, `go test -short ./...`, `staticcheck ./...`, and the openapi regeneration all pass locally, and both delivered-SQL integration tests (`TestTimeSpentQueriesMeasuredIntegration`, `TestTimeSpentQueriesProxyIntegration`) pass against a seeded database. The endpoint was also exercised end-to-end against the running frontend branch. Full detail in the review comment below.


jsos3-cms:
## Review summary — go

Rehomed and verified before opening. Results on this branch:

- Build + unit: `go build ./...` and `go test -short ./...` pass (510 tests). The lone `migrations` failure is the pre-existing needs-a-live-DB case, unrelated to this change and passing once a DB is present.
- Lint: `staticcheck ./...` clean.
- OpenAPI: regenerating the spec produces no diff, so the spec-check gate will pass.
- Delivered SQL: both integration tests pass against a seeded database, covering the MEASURED dwell math (view to next view, 30-minute clamp, editor/viewer split, and the "a save is not a boundary" property) and the PROXY save-gap attribution (the gap preceding a save attributed to that save's question, first save dropped).
- Endpoint end-to-end: against a running instance with a seeded DB, the endpoint returns 204 for an editor and for a viewer and derives the readonly flag correctly across all four cases — admin to editor, read-only admin to viewer, non-admin before deadline to editor, non-admin after deadline to viewer. It returns 403 for a system the caller has no relationship to, 400 for a malformed body, 404 for an unknown data call, and 401 without a token.

Access control, the server-side readonly derivation, and the events refactor were reviewed and are correct — the derivation matches the questionnaire's existing write-eligibility rule, and the refactor keeps the fire-and-forget audit behavior with no recursion.

Non-blocking follow-ups (coverage, not defects — the behavior is verified above): the two controller tests for the view endpoint pass without reaching the asserted path (a DB-free seam would make them meaningful), the readonly derivation has no direct unit test, and nothing asserts the 204 happy path or the 404 path through the Go write path. Worth a fast-follow, not a reason to hold this.

jsos3-cms:
## Non-blocking coverage follow-ups

Regression-protection only — the behavior was verified end-to-end (running frontend → backend, all four `readonly` branches, and the 403/400/404/401 matrix), so none block this PR. If any warrant a follow-on ticket, let us know.

- No direct unit test for the server-side editor/viewer derivation `IsReadOnlyAdmin() || (!IsAdmin() && After(dc.Deadline))`. Verified live across all four cases; extracting a pure `deriveReadOnly(user, deadline)` + a table test would lock it in.
- The two controller view-tests (`TestRecordQuestionView_ReadonlyAdminAllowed`, `_UnrelatedSystemNotAllowed`) pass without reaching the asserted path — they hit a DB-connection error first, and the unrelated-system one would still pass even if the guard were removed. A DB-free seam (inject `FindFismaSystem`/`FindDataCallByID`, like the existing `findUserByID` pattern) would make them meaningful. Also note neither runs in CI (CI runs only `-run Integration`).
- No test asserts the 204 happy path or the 404 unknown-data-call path through the Go write path.

Companion frontend follow-up and the full list are tracked on #368.

MackOverflow (approved):
looks good, no nits

PR #468, opened 2026-07-23
test(emberfall): assert array bodies now that emberfall#36 is fixed

### LINKED ISSUE #271: Add array body assertions to Emberfall ISSO CFACTS tests
Once aquia-inc/emberfall#36 is fixed and we upgrade, we can add proper body assertions to the ISSO CFACTS list test.

Currently skipped with a comment in `backend/emberfall_tests.yml` because Emberfall panics on `[]interface{}` comparison.

**What to add back:**

```yaml
  - url: http://localhost:8080/api/v1/cfactssystems
    method: GET
    headers:
      <<: *issoHeaders
    expect:
      status: 200
      headers:
        content-type: "application/json"
      body:
        json:
          data:
            - fisma_uuid: "E1D00198-36D4-4EAB-8C00-501E1D000999"
              fisma_acronym: "SLD-GEN"
```

This validates that the ISSO join filter actually returns the correct assigned system, not just a 200 with an empty array.

**Also update the Makefile** when upgrading Emberfall — the `--config` flag was renamed to `--tests` on main.

**Blocked by:** aquia-inc/emberfall#36 (PR #48 submitted)

voidspooks:
Closes #271.

## Summary

The blocker behind #271 (aquia-inc/emberfall#36, panic on array body.json comparison) was fixed in emberfall v0.4.1, which is the version the Makefile pins, and the `--config` to `--tests` rename is already in place. The literal ask (re-add the `/api/v1/cfactssystems` assertion) is moot since #326 removed that endpoint entirely. This PR adds array body assertions to the surviving equivalents that were weakened to status-only while the bug was live:

- ISSO scoped `GET /scores/aggregate?fismasystemid=1003`: pins the exact per-datacall row set, retiring the in-file TODO. Any other system's rows appearing fails on array length (emberfall array matching is length-exact, element fields subset-matched).
- ISSO aggregate for an unassigned system: exact empty-body match (`{"data":[]}`), so a scope leak fails on content, not just status.
- Admin aggregate smoke for 1001: pins one row per seeded datacall, so the cartesian-join regression described in the neighboring comment fails fast on row count.
- `GET /scores?fismasystemid=1001&datacallid=5` read-back: pins the row including `last_edited_by` audit fields (ztmf-ui#310) on the list path, previously covered only by integration tests.
- Corrects the stale comments still citing the closed emberfall#36 and the outdated DS-1 seed description (the seed also carries scores in datacalls 6 and 101-103, not just 3-4).

Note: the emberfall ISSO fixture is assigned to 1003 only, so the narrow historical bug the TODO referenced (fismasystemid filter ignored when FismaSystemIDs is pre-populated) is indistinguishable at the HTTP layer with this fixture; that path stays pinned by the model integration tests. The assertions here guard the broader contract: only the requested system's rows, exact row count, audit fields resolved on the list path.

## Validation

- `make test-unit`, `make test-integration`, `make test-e2e` all green locally (emberfall: Ran 172, Failed 0)
- Test-only change, no Go or API modifications

Expected-red: Snyk / secret-dependent CI (fork PR, no repo secrets).

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -1245,19 +1245,21 @@ tests:
   # Math correctness, response field presence, and Tier(score)==tier
   # consistency are all pinned in internal/model/scores_integration_test.go,
   # which runs against the real schema as part of `make test-full`.
-  # Emberfall's body assertion is exact-match (body.text is full-body
-  # equality, body.json equality fails on array-typed responses due to
-  # aquia-inc/emberfall#36), so these HTTP-layer smokes cover what we
-  # can express cleanly: cardinality contract, auth gates, and content
-  # type. Anything that survives unit + integration but breaks at the
-  # HTTP boundary still shows up here.
+  # These HTTP-layer smokes pin the cardinality contract, auth gates, and
+  # content type. Array body.json assertions work since aquia-inc/emberfall#36
+  # was fixed (v0.4.1), so row keys are pinned exactly below; score values
+  # stay in the integration suite.
   #
-  # The DS-1 seed (fismasystemid=1001) has 12 score rows across datacalls
-  # 3 and 4 in _test_data_empire.sql, guaranteeing a non-empty response.
+  # The DS-1 seed (fismasystemid=1001) has score rows in datacalls 3, 4, 6,
+  # and the HHS historical calls 101-103 in _test_data_empire.sql,
+  # guaranteeing a non-empty response.
 
   # 1. Scored system returns a populated aggregate via the full HTTP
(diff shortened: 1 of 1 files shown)


jsos3-cms:
Rehomed to #470 so the Snyk and emberfall E2E / Deploy CI gates run — they can't on a fork PR because it doesn't receive the org secrets. The commit was pushed unchanged and your authorship is preserved. Leaving this open until the rehome merges. Thanks @voidspooks.

jsos3-cms:
The rehome #470 has merged (squashed to main) with your authorship preserved, so this change is landed. Closing this fork as complete. Thanks @voidspooks.

PR #469, opened 2026-07-23
feat(hooks): auto-regenerate openapi.yaml in pre-commit (#391)

### LINKED ISSUE #391: Add a git hook to auto-regenerate openapi.yaml on schema/API changes (prevent drift CI failures)
## Problem
`backend/openapi.yaml` is generated from the Go handler annotations via `make generate-openapi`. CI's `openapi spec up to date` job regenerates the spec and fails on any drift. Contributors sometimes forget to regenerate (or hand-edit the file), so the drift only surfaces in CI after the fact and costs a review round-trip. This just happened on #390 (a field-ordering diff from a hand-edit).

## Goal
Regenerate the spec automatically on the developer's machine so drift never reaches CI, while keeping CI's check as the backstop and keeping CI read-only (no write token, which also means it works for fork PRs).

## Proposed approach (recommended)
Add a pre-commit hook (the repo already sets `core.hooksPath .hooks` in `make dev-setup`) that:
1. Detects staged changes to API-relevant Go files (annotated handlers with `// @...`, and request/response model structs).
2. Runs `make generate-openapi`.
3. If `backend/openapi.yaml` changed, stages it and prints a short note, so the commit always carries an up-to-date spec.

Keep it fast: no-op when no relevant Go files are staged. Requires `go tool swag`, which is already part of dev setup.

## Alternatives to weigh
- pre-push drift check: regenerate + `git diff --exit-code openapi.yaml`, fail the push on drift. Lighter and mirrors CI exactly, but blocks rather than fixes. The repo's pre-push hook already runs the test suite, so this could slot in beside it.
- Fail-closed pre-commit: regenerate, and if it changed, abort with "spec regenerated, please re-commit." Avoids auto-staging surprises at the cost of a second commit.

Recommendation: pre-commit auto-stage for the smoothest experience; fall back to the pre-push drift check if modifying files mid-commit is undesirable.

## Acceptance criteria
- Editing an annotated handler or a request/response struct and committing yields a commit whose `openapi.yaml` matches `make generate-openapi`, with no manual step.
- No relevant Go change means no regeneration overhead.
- CI `openapi spec up to date` stays as-is (backstop) and stays read-only.
- The behavior is documented (hook comment or CONTRIBUTING) so devs know it runs.

## Context
Prompted by the openapi drift on #390. Reinforces the standing rule that `openapi.yaml` is never hand-edited.

voidspooks:
Closes #391.

## Summary

Contributors sometimes forget to run `make generate-openapi` after touching annotated handlers or response models, so spec drift only surfaces in CI's `OpenAPI-Spec-Check` and costs a review round-trip (#390 most recently). This is the pre-commit auto-stage flavor the ticket recommends: when any backend Go file is staged, the hook runs `make generate-openapi` and, if `backend/openapi.yaml` changed, stages the regenerated file so the commit always carries an up-to-date spec. CI stays read-only and unchanged as the backstop.

Design notes:

- **Triggers on any staged backend `.go` file** rather than trying to detect API-relevant files. swag parses the whole backend tree (`--dir ./ --parseInternal`), so a struct change with no annotation in sight can still alter the spec; a false negative costs the CI round-trip this hook exists to prevent, while a false positive costs about a second (measured ~1.2s).
- **Generation failure blocks the commit** with swag's output shown, so broken annotations never reach CI.
- **A staged hand-edit to `openapi.yaml` is replaced by the canonical regen**, enforcing the never-hand-edit rule mechanically (a hand-edit is what caused the #390 drift).
- **Known limitation, documented in the hook:** swag reads the working tree, not the index, so partially staged Go files can leak unstaged edits into the spec. CI's drift check remains the backstop for that case.

## Validation

Exercised through real `git commit` runs:

| Scenario | Result |
| --- | --- |
| Annotation change staged | Spec regenerated + auto-staged; committed spec byte-identical to a fresh regen |
| Non-API Go change staged | Regen runs, byte-identical, "already up to date", spec untouched |
| No Go files staged | Regen (and the whole Go branch) skipped entirely |
| Broken annotation (`nosuchpkg.NoSuchType`) | Commit blocked, swag's ParseComment error shown |
| Hand-edited + staged `openapi.yaml` | Replaced by canonical regen, no trace of the hand-edit in the commit |

`make test-unit`, `make test-integration`, `make test-e2e` all green locally (emberfall: Ran 172, Failed 0).

Expected-red: Snyk / secret-dependent CI (fork PR, no repo secrets).

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -71,6 +71,27 @@ if echo "$STAGED_FILES" | grep -q '\.go$'; then
     fi
     echo "✓ Go compilation successful"
 
+    # --- OpenAPI spec sync (#391) -----------------------------------------
+    # swag parses the whole backend tree (--dir ./ --parseInternal), so any
+    # staged backend Go change can affect the generated spec. Regen costs ~1s,
+    # so it always runs rather than guessing which files are API-relevant (a
+    # wrong guess costs the CI round-trip this hook exists to prevent).
+    # Limitation: swag reads the working tree, not the index — with partially
+    # staged Go files the spec may include unstaged edits. CI's
+    # OpenAPI-Spec-Check remains the backstop for that case.
+    echo "→ Regenerating OpenAPI spec..."
+    if ! SWAG_OUT=$(make generate-openapi 2>&1); then
+        echo "$SWAG_OUT"
+        echo "❌ OpenAPI generation failed. Fix annotations before committing."
+        exit 1
+    fi
+    if ! git diff --quiet -- backend/openapi.yaml; then
+        git add backend/openapi.yaml
+        echo "✓ openapi.yaml regenerated and staged"
+    else
+        echo "✓ openapi.yaml already up to date"
+    fi
+
     echo "→ Running unit tests..."
     # Source dev.compose.env so tests that read DB_* env (e.g. controller tests
     # that touch pgx) point at the local compose-dev Postgres on :54321 instead


jsos3-cms:
Rehomed to #472 so the Snyk and backend CI gates run — they can't on a fork PR because it doesn't receive the org secrets. The commit was pushed unchanged and your authorship is preserved. Leaving this open until the rehome merges. Thanks @voidspooks.

jsos3-cms:
The rehome #472 has merged (squashed to main) with your authorship preserved, so this change is landed. Closing this fork as complete. Thanks @voidspooks.

Issue #36, opened 2024-05-29
workflow to deploy backend and infrastructure changes


talentedmrjones:
This workflow should be reusable between environments and called from other workflows when changes are detected

- build docker image when backend changes detected
- terraform apply when backend or infrastructure changes detected but only _after_ backend image is built

PR #470, opened 2026-07-23
test(emberfall): assert array bodies now that emberfall#36 is fixed

### LINKED ISSUE #271: Add array body assertions to Emberfall ISSO CFACTS tests

jsos3-cms:
This is the in-repo version of #468 by @voidspooks, moved onto an in-repo branch so the org-secret CI gates (Snyk + the emberfall E2E / Deploy jobs) run — they can't on a fork PR because it doesn't receive the org secrets. The commit is carried over unchanged and authorship is preserved; credit for the work is Cameron Testerman's.

Refs #468 · Closes #271

## What this adds

Test-only. Re-adds array-body assertions to the emberfall E2E suite (`backend/emberfall_tests.yml`) now that the emberfall#36 panic (comparing array `body.json`) is fixed in emberfall v0.4.1 — the version the Makefile and CI pin — and the `--config`→`--tests` rename is in place. It pins:

- the ISSO-scoped `GET /scores/aggregate?fismasystemid=1003` per-datacall row set (retiring the in-file TODO);
- an exact empty-body match for an unassigned-system aggregate (`{"data":[]}`), so a scope leak fails on content, not just status;
- the admin aggregate for 1001 (one row per seeded datacall, guarding the cartesian-join regression);
- the `GET /scores?fismasystemid=1001&datacallid=5` read-back including `last_edited_by` audit fields on the list path.

It also corrects stale comments citing the now-closed emberfall#36 and an outdated DS-1 seed description. No Go or API changes.

## Verification

Reviewed and rehomed on a go. Detail in the review comment below.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -1245,19 +1245,21 @@ tests:
   # Math correctness, response field presence, and Tier(score)==tier
   # consistency are all pinned in internal/model/scores_integration_test.go,
   # which runs against the real schema as part of `make test-full`.
-  # Emberfall's body assertion is exact-match (body.text is full-body
-  # equality, body.json equality fails on array-typed responses due to
-  # aquia-inc/emberfall#36), so these HTTP-layer smokes cover what we
-  # can express cleanly: cardinality contract, auth gates, and content
-  # type. Anything that survives unit + integration but breaks at the
-  # HTTP boundary still shows up here.
+  # These HTTP-layer smokes pin the cardinality contract, auth gates, and
+  # content type. Array body.json assertions work since aquia-inc/emberfall#36
+  # was fixed (v0.4.1), so row keys are pinned exactly below; score values
+  # stay in the integration suite.
   #
-  # The DS-1 seed (fismasystemid=1001) has 12 score rows across datacalls
-  # 3 and 4 in _test_data_empire.sql, guaranteeing a non-empty response.
+  # The DS-1 seed (fismasystemid=1001) has score rows in datacalls 3, 4, 6,
+  # and the HHS historical calls 101-103 in _test_data_empire.sql,
+  # guaranteeing a non-empty response.
 
   # 1. Scored system returns a populated aggregate via the full HTTP
   #    path. Regression target: 5xx anywhere in the new aggregation SQL,
-  #    JSON serialization of the new tier fields, content negotiation.
+  #    JSON serialization of the new tier fields, content negotiation. The
+  #    exact row assertion (one per seeded datacall with 1001 scores,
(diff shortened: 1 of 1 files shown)


jsos3-cms:
## Review summary — go

Rehomed on a go after an independent review that cross-checked every new assertion against the actual seed and implementation:

- The pinned row sets/counts match `_test_data_empire.sql` exactly: system 1001 aggregate → its six seeded datacalls (3, 4, 6, 101–103); ISSO-scoped 1003 aggregate → (4, 5, 101–103) where 5 is the runtime POST; unassigned-system → literal `{"data":[]}`; `GET /scores?…datacallid=5` → the single audit row written immediately before it (`last_edited_by`).
- Emberfall v0.4.1 matching semantics confirmed from source: arrays are length-exact with per-element subset matching, and the #36 panic fix is in v0.4.1 (the version the Makefile and CI pin, invoked with `--tests`).
- Not flaky: emberfall runs tests strictly in file order, and the aggregate SQL is deterministically ordered (`ORDER BY datacallid, fismasystemid, pillarid`), so the exact-array pins are stable.
- Confirmed test-only (one file), no Go/API change; the body's factual claims hold (`/cfactssystems` is gone → moot; DS-1 seed datacalls verified).

No blocking findings. Two informational notes: the exact-match assertions rely on emberfall's sequential execution (an existing pattern in this suite, not new), and they intentionally don't pin score math (owned by the integration tests).

danielbowne (approved):
Test-only, single file. Array body.json assertions restored on the aggregate + score-read paths now that emberfall v0.4.1 carries the #36 fix. Row pins line up with the DS-1/DS-3 seed sets the comments describe, empty-body match hardens the ISSO scope-leak guard, and the stale emberfall#36 comments are cleaned up. Good to merge.

PR #472, opened 2026-07-23
feat(hooks): auto-regenerate openapi.yaml in pre-commit (#391)

### LINKED ISSUE #391: Add a git hook to auto-regenerate openapi.yaml on schema/API changes (prevent drift CI failures)

jsos3-cms:
This is the in-repo version of #469 by @voidspooks, moved onto an in-repo branch so the org-secret CI gates (Snyk + the backend / Deploy jobs) run — they can't on a fork PR because it doesn't receive the org secrets. The commit is carried over unchanged and authorship is preserved; credit for the work is Cameron Testerman's.

Refs #469 · Closes #391

## What this adds

Adds an OpenAPI-regen step to the existing `.hooks/pre-commit` (wired via `git config core.hooksPath .hooks`). When any staged file is a backend `.go`, after the existing `go build` check the hook runs `make generate-openapi` and, if `backend/openapi.yaml` changed, stages the regenerated spec so every commit carries an up-to-date spec. A generation failure blocks the commit with swag's output shown, and a staged hand-edit to `openapi.yaml` is replaced by the canonical regen (the never-hand-edit rule, enforced mechanically). CI's `OpenAPI-Spec-Check` stays read-only as the backstop. Closes #391 — the spec drift that today only surfaces in CI and costs a review round-trip (e.g. #390).

## Verification

Reviewed and rehomed on a go. Detail in the review comment below.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -71,6 +71,27 @@ if echo "$STAGED_FILES" | grep -q '\.go$'; then
     fi
     echo "✓ Go compilation successful"
 
+    # --- OpenAPI spec sync (#391) -----------------------------------------
+    # swag parses the whole backend tree (--dir ./ --parseInternal), so any
+    # staged backend Go change can affect the generated spec. Regen costs ~1s,
+    # so it always runs rather than guessing which files are API-relevant (a
+    # wrong guess costs the CI round-trip this hook exists to prevent).
+    # Limitation: swag reads the working tree, not the index — with partially
+    # staged Go files the spec may include unstaged edits. CI's
+    # OpenAPI-Spec-Check remains the backstop for that case.
+    echo "→ Regenerating OpenAPI spec..."
+    if ! SWAG_OUT=$(make generate-openapi 2>&1); then
+        echo "$SWAG_OUT"
+        echo "❌ OpenAPI generation failed. Fix annotations before committing."
+        exit 1
+    fi
+    if ! git diff --quiet -- backend/openapi.yaml; then
+        git add backend/openapi.yaml
+        echo "✓ openapi.yaml regenerated and staged"
+    else
+        echo "✓ openapi.yaml already up to date"
+    fi
+
     echo "→ Running unit tests..."
     # Source dev.compose.env so tests that read DB_* env (e.g. controller tests
     # that touch pgx) point at the local compose-dev Postgres on :54321 instead


jsos3-cms:
## Review summary — go

Rehomed on a go after I reviewed this and confirmed the swag pattern it relies on:

- swag with `--dir ./ --parseInternal` parses the whole backend tree including internal structs, so the hook's choice to trigger on *any* staged backend `.go` file is correct — a struct change with no annotation in sight can still shift the generated spec, so filtering to "API-relevant" files would risk false negatives. (Checked against swag's documented CLI behavior.)
- Determinism holds: the hook and CI's `OpenAPI-Spec-Check` run the byte-identical `swag init … --dir ./ --parseInternal --v3.1 --outputTypes yaml -o ./`, with swag pinned via the `tool` directive in `backend/go.mod` + `go.sum` — so there's no local-vs-CI version skew that would stage a spec CI then rejects.
- Safe hook mechanics: `set -e` isn't tripped by the `if !`-guarded generate/diff steps (same idiom as the existing `go build` check); `git add backend/openapi.yaml` targets one named path; `make generate-openapi` needs no Docker.

One non-blocking note for future hardening: the documented working-tree-vs-index limitation is actually a bit broader than "partially staged files" — because swag parses the entire `backend/` tree, any dirty or untracked `.go` file can influence the regen (or block the commit on a broken annotation elsewhere). That mirrors the existing `go build` / `go test` steps in the same hook, and CI's drift check is the backstop, so it's fine to land; a `git stash --keep-index -u` wrapper is the option if it ever bites. Tooling-only change (a git hook), no Go or API modifications.

danielbowne (approved):
Pre-commit hook only, no runtime or CI-behavior change. The regen block is correctly gated inside the staged-Go check and sits after go build / before unit tests, so broken annotations surface with swag's output and block the commit. Staged-diff check on backend/openapi.yaml means clean commits stay no-op, and a hand-edited spec gets replaced by the canonical regen — enforcing the never-hand-edit rule mechanically. The working-tree-vs-index limitation is documented in the hook and CI's OpenAPI-Spec-Check remains the backstop. Good to merge.

MackOverflow (approved):
no nits, looks good

PR #473, opened 2026-07-23
Add isso self service endpoints

costacalvin:
Part of epic CMS-Enterprise/ztmf#467 · backend sub-issue Closes CMS-Enterprise/ztmf#462 · paired frontend CMS-Enterprise/ztmf-ui#598.

## Summary

Adds **ISSO self-service** for the `SYSTEM_DELEGATE` role (contractor/support-staff access to a system's data-call answers, introduced in #458). Today the only way to create or assign any user is the admin-gated Users surface — every user write path begins with `IsAdmin()`, and an ISSO is not an admin — so standing up a delegate requires an administrator round-trip.

This adds a **separate, narrow, system-anchored** API surface so an ISSO can add / remove / renew delegates on a FISMA system they are assigned to, and only that system, without unlocking the admin Users table or relaxing any existing `IsAdmin()` gate. Delegates carry a **mandatory expiration** and are denied at authentication once lapsed (via the same rejection path as a soft-deleted user); regular users never expire and are unaffected. The capability is enabled **per-OpDiv** via an admin-managed toggle.

## Endpoints

| Method | Path | Who | Purpose |
|---|---|---|---|
| `GET` | `/fismasystems/{id}/delegates` | system viewers | Current delegates on the system (with `access_expires_at` for the expired badge) |
| `GET` | `/fismasystems/{id}/delegate-candidates?q=` | ISSO/admin managing the system | Existing delegates eligible to attach (picker) |
| `POST` | `/fismasystems/{id}/delegates` | ISSO/admin managing the system | Add a new person (Name+Email) or attach an existing eligible delegate |
| `PATCH` | `/fismasystems/{id}/delegates/{userid}` | ISSO/admin managing the system | Renew/adjust a delegate's expiration |
| `DELETE` | `/fismasystems/{id}/delegates/{userid}` | ISSO/admin managing the system | Remove a delegate from this system |
| `PUT` | `/opdivs/{opdiv_id}/system-delegate-enabled` | **Owner + HHS admin only** | Toggle the "Add System Delegate Role" capability for an OpDiv |

`GET /opdivs` now returns `system_delegate_enabled` on each OpDiv (read side of the toggle).

## Behavior & decisions (per epic #467)

- **Actor scope:** ISSO only among non-admins (ISSM and delegates excluded); admins who manage the system also pass. Unauthorized/missing systems return **404** (no-leak), never 403.
[...]

Issue #467, opened 2026-07-23
ISSO self-service: add expiring System Delegates to owned systems


MackOverflow:
> Decisions below resolved by @eschwein-usds on 2026-07-22. Pending final acknowledgement from @MackOverflow before implementation starts.

> Moved from CMS-Enterprise/ztmf-misc#254.

## Epic Summary

The `SYSTEM_DELEGATE` role (backend #458, with its paired frontend change) lets contractors and support staff receive system-scoped access to a FISMA system's data-call answers, without inheriting any ISSO/ISSM configuration surface. Today, however, the only way to create or assign any user is the admin-gated Users surface: every user create/assign write path begins with an `IsAdmin()` check, and an ISSO is not an admin. Standing up a delegate therefore requires an administrator round-trip.

This epic lets an **ISSO self-serve**: from the FISMA system detail page they already use, an ISSO can add a delegate to a system they are responsible for, and only that system, and later revoke that delegate. Delegate accounts carry a mandatory expiration date; when a delegate's access lapses, they are treated as disabled at authentication time. Regular users are unaffected and never expire.

The work splits into a backend change (new scoped endpoint + authorization + expiry) and a frontend change (a delegates section on the system detail page). Both are tracked as sub-issues below.

## Success Criteria

- An ISSO can add a `SYSTEM_DELEGATE` to a FISMA system they are assigned to, by email:
  - **New person** (not yet a ZTMF user): the ISSO supplies Name and Email. The delegate's OpDiv is not asked; it is inherited from the acting ISSO. The role is set to `SYSTEM_DELEGATE`.
  - **Existing user**: the ISSO may attach an existing user only when that user is already a `SYSTEM_DELEGATE` and already has an OpDiv set that matches the ISSO's OpDiv. Every other existing-user case requires an administrator (see decision 3).
- An ISSO can remove a delegate from a system they are assigned to.
- An ISSO can only ever create or affect `SYSTEM_DELEGATE` records, and only on systems in their own assignment set. No path lets an ISSO escalate a role, alter a non-delegate (ISSO/ISSM/admin) user, add a delegate outside their OpDiv, or reach a system they do not own.
- Delegate expiration is mandatory in the ISSO add flow, defaulting to three months. A delegate whose expiration has passed is denied access at authentication, using the same rejection path as a soft-deleted user, while the delegate row and its system assignments are retained for renewal and audit.
- Regular (non-delegate) users have no expiration and are unaffected.
[...]

PR #474, opened 2026-07-23
Add ISSO self-service endpoints (rehome of #473)

### LINKED ISSUE #462: System Delegate self-service: scoped ISSO add/remove endpoint + optional expiry
Part of CMS-Enterprise/ztmf#467

> Parent-epic decisions resolved 2026-07-22 (see #467). Pending @MackOverflow acknowledgement before implementation starts.

## Feature Type

New API capability / authorization change.

## Workflow Description

Let an ISSO create and remove a `SYSTEM_DELEGATE` on a FISMA system they are assigned to, without opening the general admin user-management surface. The scope is anchored to the system in the request path, so an ISSO can never act outside the systems they own.

**Proposed endpoints (system-anchored):**

- `POST /api/v1/fismasystems/{fismasystemid}/delegates` - add or invite a delegate to this system.
- `DELETE /api/v1/fismasystems/{fismasystemid}/delegates/{userid}` - revoke a delegate from this system.
- Optional `PATCH /api/v1/fismasystems/{fismasystemid}/delegates/{userid}` - renew/adjust expiry.

**Authorization:** gate on `authdUser.IsAssignedFismaSystem(fismasystemid)` (admins also pass). A system the actor does not own returns `ErrNotFound` so existence is not leaked, mirroring the existing `guardManageFismaSystem` pattern.

**Add behavior, resolved (see #467 decision 3):**

- **Email resolves to no existing user (new person):** the request carries Name and Email. Create a `SYSTEM_DELEGATE` (identity provider derived as it is elsewhere), grant the request system's OpDiv so the target satisfies `CanBeAssignedFismaSystem`, and insert the `users_fismasystems` assignment, all in a single transaction. The delegate's OpDiv is always the acting ISSO's OpDiv (the system's OpDiv); it is never supplied by the client.
- **Email resolves to an existing user:** permitted **only** when that user is already a `SYSTEM_DELEGATE` **and** already has an OpDiv set that matches the request system's OpDiv. In that case, insert only the `users_fismasystems` assignment; do not touch the user's role or OpDiv.
- **Every other existing-user case is rejected and requires an administrator:** an existing account with no OpDiv set, an existing delegate whose OpDiv differs, or any non-delegate existing user (ISSO/ISSM/admin). Return a clear error indicating an administrator is required. This endpoint never sets or changes a user's OpDiv and never re-roles an existing user.

**Mandatory expiry (see #467 decisions 1 and 2):**

- Add a nullable `access_expires_at TIMESTAMPTZ` column to `users` (null = never expires). Regular users remain null.
- Expiry is **required** when this endpoint creates or invites a delegate. If the client omits it, default to three months from creation. The column stays nullable only so non-delegate users can be null.
- Enforce in the **same** authentication middleware that already rejects soft-deleted users: when `access_expires_at IS NOT NULL AND access_expires_at < now()`, return the same 403 `ACCOUNT_NOT_PROVISIONED` response. This is lazy, authoritative, and needs no scheduled job.
- At expiry we deny access only. The delegate row and its `users_fismasystems` assignments are retained so an ISSO can renew via `PATCH` without recreating the user (see #467 decision 4).

**OpDiv-level enablement toggle (see #467 decisions 6 and 7):**

- The System Delegate capability is enabled per OpDiv via an "Add System Delegate Role" toggle on the Manage OpDivs panel. The backend must persist a per-OpDiv enablement flag and expose read/write endpoints for it.
- Only the **HHS admin** and **Owner** roles may set this toggle. Authorize the write path to those two roles and reject all others (including ISSO and OpDiv Admin unless they also hold Owner/HHS admin).
- The ISSO self-service add flow above is gated on the delegate capability being enabled for the acting ISSO's OpDiv; when the toggle is off, the add is refused.

**Reuse:** `guardManageFismaSystem`, `CanBeAssignedFismaSystem`, `SetUserOpDivs` (carries `granted_by` for audit; new-person path only), `UserFismaSystem.Save` (`ON CONFLICT DO NOTHING`), and the existing identity-provider derivation.

## Acceptance Criteria

- [ ] New system-anchored delegate add/remove endpoint(s), gated by `IsAssignedFismaSystem` (admins also pass); non-owned or missing system returns not-found.
- [ ] New-person path creates a `SYSTEM_DELEGATE` from Name + Email, grants the system's OpDiv, and assigns the system in one transaction; OpDiv is always the acting ISSO's OpDiv and never client-supplied.
- [ ] Existing-user path attaches only when the user is already a `SYSTEM_DELEGATE` and already has an OpDiv matching the request system's OpDiv; it inserts only the assignment and modifies neither role nor OpDiv.
- [ ] Every other existing-user case (no OpDiv set, different OpDiv, or non-delegate) is rejected with an administrator-required error and no modification.
- [ ] An ISSO can never assign any role other than `SYSTEM_DELEGATE`, and never on a system outside their assignment set.
- [ ] Nullable `access_expires_at` migration (up adds column, down drops it); regular users unaffected (null).
- [ ] Delegate creation requires an expiration; when omitted it defaults to three months.
- [ ] Authentication denies an expired delegate via the existing soft-delete rejection path; the row and assignments are retained.
- [ ] Per-OpDiv "System Delegate enabled" flag is persisted, with read/write endpoints on the Manage OpDivs surface.
- [ ] Only HHS admin and Owner roles can set the "Add System Delegate Role" toggle; all other roles are rejected.
- [ ] ISSO delegate add is refused when the toggle is off for the acting ISSO's OpDiv.
- [ ] Every grant and revoke is recorded in the audit trail (`granted_by`).
- [ ] Each new ISSO-reachable write is added to the delegate access-invariant test set.
- [ ] Unit + `make test-integration` cover: ISSO on own system succeeds, ISSO on another system is denied, role-escalation attempt is denied, existing non-delegate is denied, existing delegate in a different OpDiv is denied, expired-delegate authentication is denied.
- [ ] Emberfall access-matrix rows added for the new endpoint(s).
- [ ] OpenAPI regenerated from swaggo annotations (`make generate-openapi`) with no drift.

## Additional Context

- All current user create/assign/delete gates begin with `!authdUser.IsAdmin()` and exclude ISSO; this issue deliberately adds a **separate, narrow** endpoint rather than relaxing those gates.
- "ISSO owns a system" is the `users_fismasystems` assignment (the same junction used for read/answer scope); `issoemail` on `fismasystems` is a contact/display field only and is not an authorization signal.

jsos3-cms:
## Summary

In-repo rehome of #473 by @costacalvin (Calvin Costa) so the repository CI gates (which fork PRs can't run) execute against this commit. Original authorship is preserved; the original PR #473 stays open as a draft until this merges.

Part of epic #467 · **Closes #462** · paired frontend CMS-Enterprise/ztmf-ui#598 (rehomed as ztmf-ui#607).

Adds **ISSO self-service** for the `SYSTEM_DELEGATE` role (contractor/support-staff access to a system's data-call answers, introduced in #458). Previously every user write path began with `IsAdmin()`, so standing up a delegate required an administrator round-trip; this lets an ISSO self-service delegates within their own scope.

## What changed

- New `delegates.go` controller + `auth/middleware.go` scope enforcement + `router.go` registration.
- Per-OpDiv "System Delegate Role" capability flag — migration `0049opdivsystemdelegateflag` (API + migration only for now; UI is FE work).
- User access-expiry — migration `0050usersaccessexpiry` + `users.go` management.
- `openapi.yaml` regenerated; emberfall e2e + `delegates_integration_test.go` coverage.

## Status

**Opened as a draft pending review.** It will be converted to a ready PR once the review verdict is Go; if issues are found it stays draft with the findings noted. This is the paired backend of the FE rehome ztmf-ui#607.

## Provenance / verification

Commit cherry-picked verbatim from #473 (`costacalvin/ztmf:enhancement/isso-self-service`), authorship preserved, atop current `main`. On this branch: `go build ./...` and `go vet` are green; migrations 0049/0050 self-register and don't collide (main is at 0048); `make generate-openapi` reproduces the committed `openapi.yaml` (no drift).

---

## Reviewer changes applied on top of the original (cc @costacalvin)

Two review nits resolved in-scope in a follow-up commit; the original commit is preserved beneath it:
- `sanitizeErr` now attaches a `DELEGATE_NOT_ENABLED` code to the capability-off 403, so the paired frontend can render an in-dialog guard instead of the global auth interceptor swallowing a bare 403 into a generic toast.
- Migration 0051 adds `CHECK (role = 'SYSTEM_DELEGATE' OR access_expires_at IS NULL)` as a DB backstop for the expiry invariant the write paths already uphold in application code.

## Deliberately out of scope for this ticket (follow-on)

Flagged so the boundary is explicit — not changed here:
- **Per-user vs per-assignment expiry** — renewing a multi-OpDiv delegate extends access in every OpDiv they hold. Accepted under the epic's current single-system model; would only need revisiting if delegates become multi-system. Not a defect in this ticket's scope.
- **`fmt.Sscan` path-param parse error unchecked** — matches the codebase-wide controller convention (fails closed to id 0 → 404). Left as-is to avoid diverging from the established pattern; not a regression.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -23,6 +23,10 @@ UPDATE public.opdivs SET is_parent = TRUE WHERE LOWER(code) = 'empire';
 -- migration only enables code='CMS', so set the test OpDiv explicitly here.
 -- REBELLION is intentionally left disabled to exercise the OpDiv-gated 404.
 UPDATE public.opdivs SET insights_enabled = TRUE WHERE LOWER(code) = 'empire';
+-- Enable the System Delegate self-service capability (#467) for EMPIRE so the
+-- ISSO add flow is exercisable; REBELLION is left disabled (migration default
+-- false) to exercise the toggle-off refusal (ErrDelegatesNotEnabled).
+UPDATE public.opdivs SET system_delegate_enabled = TRUE WHERE LOWER(code) = 'empire';
 
 -- 13 sister divisions of the Empire.
 INSERT INTO public.opdivs (code, name, is_parent, active) VALUES
@@ -337,6 +341,27 @@ INSERT INTO public.users_fismasystems VALUES ('33333333-3333-3333-3333-333333333
 INSERT INTO public.users_fismasystems VALUES ('44444444-4444-4444-4444-444444444444', 1003) ON CONFLICT DO NOTHING; -- Krennic -> Shield Gen
 INSERT INTO public.users_fismasystems VALUES ('66666666-6666-6666-6666-666666666666', 1003) ON CONFLICT DO NOTHING; -- Emberfall ISSO -> Shield Gen (for system_enrichment access E2E tests)
 INSERT INTO public.users_fismasystems VALUES ('55555555-5555-4555-8555-555555555555', 1001) ON CONFLICT DO NOTHING; -- System Delegate -> Death Star (answers-only E2E: score write allowed, target maturity 403)
+
+-- System Delegate self-service fixtures (#467). All are SYSTEM_DELEGATEs so the
+-- reuse-vs-reject branches of the add flow are exercisable against Death Star
+-- (system 1001, EMPIRE OpDiv, capability enabled above).
+--   - Reuse.Delegate: already a delegate holding EMPIRE and NOT yet on any system,
+--     so the existing-eligible path can attach it.
+--   - Wrong.Opdiv.Delegate: a delegate holding REBELLION (not EMPIRE), so the add
+--     flow must reject it as administrator-required.
+INSERT INTO public.users (userid, email, fullname, role, identity_provider)
+    VALUES ('5e1e6a7e-0000-4000-8000-000000000001', 'Reuse.Delegate@empire.test', 'Reuse Delegate', 'SYSTEM_DELEGATE', 'entra')
+    ON CONFLICT DO NOTHING;
+INSERT INTO public.users_opdivs (userid, opdiv_id)
+    SELECT '5e1e6a7e-0000-4000-8000-000000000001', opdiv_id FROM public.opdivs WHERE code = 'EMPIRE'
+    ON CONFLICT DO NOTHING;
+
+INSERT INTO public.users (userid, email, fullname, role, identity_provider)
+    VALUES ('5e1e6a7e-0000-4000-8000-000000000002', 'Wrong.Opdiv.Delegate@rebellion.test', 'Wrong OpDiv Delegate', 'SYSTEM_DELEGATE', 'entra')
+    ON CONFLICT DO NOTHING;
+INSERT INTO public.users_opdivs (userid, opdiv_id)
+    SELECT '5e1e6a7e-0000-4000-8000-000000000002', opdiv_id FROM public.opdivs WHERE code = 'REBELLION'
+    ON CONFLICT DO NOTHING;
 
 -- DataCall-System Assignments (Systems participating in audits)
 INSERT INTO public.datacalls_fismasystems VALUES (3, 1001) ON CONFLICT DO NOTHING; -- DS-1 in FY2024 review
--- a/backend/cmd/api/internal/auth/middleware.go
+++ b/backend/cmd/api/internal/auth/middleware.go
@@ -27,6 +27,12 @@ const (
 	// flow rejected an existing account it cannot self-serve, so the FE renders "an
 	// administrator must handle this user" instead of a generic validation error (#467).
 	CodeDelegateRequiresAdmin = "DELEGATE_REQUIRES_ADMIN"
+	// Set by the controller (sanitizeErr): the delegate add flow was refused because
+	// the OpDiv does not have the System Delegate capability enabled. Carries a code
+	// so the FE renders an in-dialog "not enabled for this OpDiv" guard rather than
+	// keying on the bare 403, which the global auth interceptor would otherwise
+	// swallow into a generic toast.
+	CodeDelegateNotEnabled = "DELEGATE_NOT_ENABLED"
 )
 
 // Package-level seams over the model lookups so tests can stub them without a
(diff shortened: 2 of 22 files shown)


jsos3-cms:
**Reviewed — Go.** Verified end-to-end against a live isolated Postgres (migrations 0049/0050 applied, empire seed), not just static read.

Authorization (the load-bearing part) checks out: all new `/delegates` + `/system-delegate-enabled` routes sit on the authenticated subrouter; a two-phase fail-closed gate scopes `OPDIV_ADMIN` to their own OpDiv and admits `ISSO` only for their assigned systems; `ISSM` is read-only by design; a `SYSTEM_DELEGATE` can never manage delegates (404, self-hidden). No authorization trusts a client-supplied ID — targets are re-loaded and re-checked server-side. The existing-user re-attach path (the usual privilege-leak spot) only auto-attaches a user who is *already* a `SYSTEM_DELEGATE` in the system's OpDiv; everything else returns `ErrDelegateRequiresAdmin`. The per-OpDiv capability flag and access-expiry are both genuinely enforced (expiry rejects at middleware on every request), verified with past-dated fixtures.

Also clean: no SQL injection (all squirrel/pgx parameterized), migrations safe on populated tables (0049 opt-in default false; 0050 nullable = never-expires, no lockout) with no numbering collision, OpenAPI byte-identical on regen, and test coverage explicitly hits the authz edges (out-of-scope 403/404, expiry, cross-OpDiv, capability-off).

Two non-blocking follow-ups worth a tracking issue (author already flagged the first):
- Add a DB `CHECK (role = 'SYSTEM_DELEGATE' OR access_expires_at IS NULL)` so the expiry invariant survives a future write path, not just `SaveUser`'s explicit nil-out.
- Revisit whether `SetDelegateExpiry` being per-user (not per-assignment) could extend a multi-OpDiv delegate's access in OpDivs the acting ISSO doesn't manage.

MackOverflow (approved):
lgtm

PR #479, opened 2026-07-27
fix(#471, #478): make the -short suite independent of database state

### LINKED ISSUE #471: make test-unit requires a live database: migration init() connects before -short can skip
## Problem
`make test-unit` (`go test -short ./...`) is the documented fast suite, but the `cmd/api/internal/migrations` package fails outright when no database is reachable and no DB_* env is set:

```
initializing config...
could not connect to db failed to connect to `user= database=sslmode=require`: ...
FAIL  github.com/CMS-Enterprise/ztmf/backend/cmd/api/internal/migrations
```

No test even runs (`-v` shows no `=== RUN`). Verified on current main (444dbd9) with a cold cache (`-count=1`).

## Root cause
Every migration file registers itself via `init()` calling `getMigrator().AppendMigration(...)`, and `getMigrator()` opens a live connection (`db.MigrationConn`) inside its `once.Do` (`migrations.go:56`), with `log.Fatal` on failure. Importing the package, which its test binary does, dials the database at package-init time, before `TestPopulateSkipsWhenSeededIntegration`'s `testing.Short()` guard can skip anything.

## Why it goes unnoticed
- The git hooks source `dev.compose.env` before `go test -short`, so with the dev stack up the init-time connect succeeds against :54321. Run `make test-unit` by hand without that env and it fails; with an unrelated Postgres on host :5432 it can instead hang the pgx handshake.
- The Go test cache can replay a pass from an env-sourced run even after the DB env is scrubbed (init-time env reads don't appear to invalidate the cache), so the failure presents as intermittent.

## Proposed fix
Decouple registration from connection: `init()`s append to a plain slice (pure, no I/O), and `Run()` constructs the tern migrator with a connection at the point migrations actually execute. The `-short` suite then runs DB-free as documented; migration behavior at boot is unchanged.

## Acceptance
- With no database reachable and no DB_* env: `go test -short -count=1 ./...` passes
- `make test-integration` and dev-boot migration/seed behavior unchanged

Noted in #452's description as "will be filed separately"; this is that filing.

### LINKED ISSUE #478: TestAddSystemDelegate_AllowedActorsPassGate passes only when the database is unreachable
## Problem

`TestAddSystemDelegate_AllowedActorsPassGate` (added in #474) calls the real `AddSystemDelegate` handler for an assigned ISSO and an OWNER and asserts the response is not 404/403. Those actors clear the in-memory pre-check, and then `guardManageDelegates` calls `model.FindFismaSystem(1)`. With no reachable database that query errors, the handler returns 500, and the assertion holds, which is the scenario the test's own comment describes. With a reachable seeded database the query succeeds instead, and the empire seed has no `fismasystemid = 1` (its IDs run 1001-2015), so the guard returns not-found and the test fails:

```
--- FAIL: TestAddSystemDelegate_AllowedActorsPassGate/ISSO_assigned (0.00s)
    delegates_test.go:126: Should not be: 404
```

The test only passes when the database connection fails, so it encodes a dependency on the database being unreachable. CI never notices because no DB env is set there. Locally it bites hard: the pre-commit hook from #472 sources `dev.compose.env` so `go test -short` points at the compose dev DB on :54321, which means every backend commit is blocked while the dev stack is up. Verified on current main (ecc4953) with `-count=1`.

## Fix

Extract the in-memory fail-closed pre-check at the top of `guardManageDelegates` (`delegates.go:49`) into a pure helper and assert the allowed actors against that helper directly. The gate result is then deterministic whether or not a database is reachable. The denied-actor tests are unaffected, since they fail closed before any DB work.

Surfaced while working #471, which is the same principle applied to the migrations package: the `-short` suite must not depend on database state. I'll carry the fix as a commit on the #471 PR.

voidspooks:
Closes #471. Closes #478.

Two commits, one theme: the `-short` unit suite must not depend on database state.

## Commit 1: fix(#478) delegate gate test

`TestAddSystemDelegate_AllowedActorsPassGate` drove the real `AddSystemDelegate` handler and asserted the response was not 404/403. Allowed actors clear the in-memory role check and proceed to `model.FindFismaSystem(1)`, so the assertion only held when that query failed. Against a live seeded database (which the #472 pre-commit hook guarantees whenever the dev stack is up) the empire seed has no `fismasystemid = 1` and the guard correctly returns not-found, failing the test and blocking every backend commit. Full mechanism on #478.

The fix extracts the fail-closed role pre-check from `guardManageDelegates` into `mayManageDelegates`, a pure helper, and asserts the allowed actors against it directly. Gate behavior is unchanged, and the denied-actor tests are untouched (they already fail closed before any DB work). The controller suite now passes identically with a live dev DB and with no database at all.

## Commit 2: fix(#471) pure migration registration

Every migration file's `init()` called `getMigrator().AppendMigration`, and `getMigrator` opened a live `db.MigrationConn` inside its `once.Do` with `log.Fatal` on failure, so importing the package dialed the database at package-init time, before any test's `testing.Short()` guard could run. `go test -short ./...` failed outright (or hung) without a reachable database, and the Go test cache replaying env-sourced passes masked it. Repro details on #471.

Registration now appends to a plain package-level slice via `appendMigration` (a mechanical one-line rename across all 51 migration files, verifiable with `grep -r getMigrator backend/`, which now returns nothing), and `Run()` opens the connection, builds the tern migrator, replays the registry in registration order, and migrates. Boot behavior is unchanged: same connection path, same `log.Fatal` on failure, same lexical execution order. `TestRegistryPopulated` pins the contract.

## Validation

The #471 acceptance criterion holds: with no DB_* env and a cold cache (`-count=1`), `go test -short ./...` passes with no database reachable. The controller suite passes identically with the live dev DB (env sourced, dev stack up) and with a scrubbed env. `make test-integration` and `make test-e2e` (emberfall) are both green, and the integration run exercises the new boot path for real: the test API container migrates and seeds a fresh ephemeral DB through the reworked `Run()`. `openapi.yaml` is byte-identical after regen, since no API surface changed.

Snyk / secret-dependent checks will show red as usual on fork PRs; maintainers re-run after pulling the branch.

### DIFF
--- a/backend/cmd/api/internal/controller/delegates.go
+++ b/backend/cmd/api/internal/controller/delegates.go
@@ -31,6 +31,15 @@ func pathFismaSystemID(r *http.Request) (int32, bool) {
 	return id, id != 0
 }
 
+// mayManageDelegates is the cheap fail-closed pre-check of the delegate
+// management gate: only an admin write tier, or an ISSO assigned to this
+// system, may proceed. An OPDIV_ADMIN passes here and is narrowed to its OpDiv
+// after the system loads. Pure (no I/O), so the role boundary stays assertable
+// with no database (#478).
+func mayManageDelegates(u *model.User, id int32) bool {
+	return u.IsAdmin() || (u.Role == "ISSO" && u.IsAssignedFismaSystem(id))
+}
+
 // guardManageDelegates verifies the acting user may manage delegates on the
 // system and returns the system for the caller to reuse. A missing system, an
 // OpDiv-less system, or an unauthorized actor all return ErrNotFound so the
@@ -43,10 +52,7 @@ func pathFismaSystemID(r *http.Request) (int32, bool) {
 // The OpDiv-scope tightening for an OPDIV_ADMIN needs the system's OpDiv, so it
 // runs after the load via the authoritative CanManageSystemDelegates check.
 func guardManageDelegates(r *http.Request, authdUser *model.User, id int32) (*model.FismaSystem, error) {
-	// Cheap fail-closed pre-check: only an admin write tier, or an ISSO assigned
-	// to this system, may proceed. An OPDIV_ADMIN passes here and is narrowed to
-	// its OpDiv after the system loads.
-	if !authdUser.IsAdmin() && !(authdUser.Role == "ISSO" && authdUser.IsAssignedFismaSystem(id)) {
+	if !mayManageDelegates(authdUser, id) {
 		return nil, ErrNotFound
 	}
 
--- a/backend/cmd/api/internal/controller/delegates_test.go
+++ b/backend/cmd/api/internal/controller/delegates_test.go
@@ -115,16 +115,16 @@ func TestListDelegateCandidates_DeniedActorsNotFound(t *testing.T) {
 	}
 }
 
-// An ISSO assigned to the system, and an unscoped admin, both pass the gate; with
-// no DB they fail downstream, but must not be gated with 403/404.
+// An ISSO assigned to the system, and an unscoped admin, both pass the gate.
+// Asserted against the pure pre-check rather than the full handler: past the
+// gate the handler does real DB work, so its status depends on the environment
+// (an absent DB errors, a live one 404s unless the seed carries this system),
+// which made the handler-level assertion pass only when the DB was unreachable (#478).
 func TestAddSystemDelegate_AllowedActorsPassGate(t *testing.T) {
 	issoAssigned := &model.User{UserID: "33333333-3333-4333-8333-333333333333", Role: "ISSO", AssignedFismaSystems: []*int32{int32Ptr(1)}}
 	for name, u := range map[string]*model.User{"ISSO assigned": issoAssigned, "OWNER": adminUser} {
 		t.Run(name, func(t *testing.T) {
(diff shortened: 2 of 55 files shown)


jsos3-cms:
Thanks @voidspooks. Fork PRs can't run our repository CI gates, so I've rehomed your two commits in-repo as #486 with your authorship preserved. Review verdict: **Go** — verified empirically that `-short` now runs DB-free (the original package-init DB dial reproduced on `main` and confirmed gone), a fresh Postgres migrates+seeds to version 51, and the delegate-gate extraction is behavior-identical with the denied-actor tests untouched. `Closes #471`/`Closes #478` now live on #486.

One tidy-up in the rehome: a trailing comment-only commit strips the issue numbers from a few code comments (we keep issue linkage on the PR `Closes` lines, not in comments). This PR stays a draft while #486 is in flight; I'll close it once #486 merges. Nothing needed from you.

jsos3-cms:
Merged in-repo via the rehome #486 (squash-merged, `6d51b84`), which closed #471 and #478. Your two commits were carried with authorship preserved; the migration-registration refactor also absorbed the metadata migrations (0052/0053) that landed on main in the meantime. Closing the source draft — thanks @voidspooks!

PR #486, opened 2026-07-27
Make the -short suite independent of database state (rehome of #479)

### LINKED ISSUE #471: make test-unit requires a live database: migration init() connects before -short can skip

### LINKED ISSUE #478: TestAddSystemDelegate_AllowedActorsPassGate passes only when the database is unreachable

jsos3-cms:
## Summary

In-repo rehome of #479 by @voidspooks (Cameron Testerman) so the repository CI gates (which fork PRs can't run) execute against these commits. Original authorship is preserved on both cherry-picked commits; the original PR #479 stays open as a draft until this merges.

**Closes #471. Closes #478.**

Makes the `-short` test suite independent of database state.

## What changed

- **#471 — pure migration registration:** migration `init()` no longer dials the database at package-import time. Each migration registers purely (`appendMigration`), and the DB connection is deferred to when migrations actually run, so `go test -short ./...` no longer requires (or hangs on) a reachable database.
- **#478 — delegate-gate test purity:** the fail-closed role pre-check is extracted from `guardManageDelegates` into a pure `mayManageDelegates` helper, and the allowed-actor test asserts against that helper directly instead of driving the live handler (which depended on a seeded `fismasystemid = 1`). Gate authorization behavior is unchanged; denied-actor tests are untouched.

## Provenance / verification

Two commits cherry-picked verbatim from #479 (`voidspooks/ztmf:fix/471-migrations-pure-registration`), authorship preserved, atop current `main`. On this branch: `go build ./...` + `go vet` green; `make generate-openapi` reproduces the committed `openapi.yaml` (no drift — no API surface changed). The DB-free `-short` behavior and a full migrate-against-real-DB run are being verified before this is marked ready.

## Review

Independent review verdict: **Go**, verified empirically — the DB-free `-short` behavior (the original package-init DB dial was reproduced on `main` and confirmed gone here), a fresh-Postgres migrate+seed reaching migration version 51 (same set/order/tracking), and the delegate-gate extraction proven behavior-identical (De Morgan) with all denied-actor fail-closed tests untouched. `go build`/`vet`/`-race`, the full suite, and the integration suite all pass; `openapi.yaml` byte-identical.

A trailing comment-only commit strips issue numbers from a few code comments per our comment convention (the `Closes` lines above are the linkage). Two non-blocking nits left as optional follow-up: a stale shared header comment above the delegate-gate tests, and the registry smoke test could pin the exact migration count.


danielbowne (dismissed):
Approving. Checked the two things that could bite here and both are clean.

Migration order is preserved: no renames or deletions, the file set is otherwise identical to main, and `Run()` replays the registry in the same lexical init order the old code appended in. The conversion is also compile-enforced, since `getMigrator` is gone entirely, so an unconverted file couldn't build rather than silently skip.

The gate extraction is behaviour-identical. `!IsAdmin && !(ISSO && assigned)` to `!(IsAdmin || (ISSO && assigned))` holds, and short-circuit order is preserved so `IsAssignedFismaSystem` still isn't called for an admin.

Two small follow-ups, neither blocking:

- The allowed-actor test now asserts the helper rather than the handler, so it no longer covers the wiring. If `guardManageDelegates` stopped calling `mayManageDelegates`, it would still pass. The denied-actor tests still drive the handler, so fail-closed stays covered end to end, which is the direction that matters. An integration test for the allowed path would close it.
- `TestRegistryPopulated` doesn't assert name uniqueness. A duplicated registration shifts every later version number, so it's worth three lines.

jsos3-cms:
Updated this branch onto current `main` and fixed two things — **retagging @danielbowne** since the new commits supersede the prior approval.

**1. Build break on merge (the important one).** #490 merged `0052` / `0053` to `main` after this branch was cut, and they still call the `getMigrator()` this PR removes — so merging as-approved would have broken the build (`undefined: getMigrator` on both files). Merged `main` in and converted both migrations to the pure `appendMigration` registration to match the refactor. Verified: `go build ./...` clean, the `-short` suite runs DB-free, and `make test-integration` migrates cleanly through `0053` via the new registry (plus the ztmf#490 CHECK↔vocab drift-guard test).

**2. Your registry-uniqueness follow-up.** Added a guard to `TestRegistryPopulated`. One wrinkle worth flagging: a strict *name*-uniqueness check fails on `main` today — two pre-existing distinct migrations share a label (`0002`/`0004` are both "ordering columns", and `0010` is labeled "users table role type", a copy-paste of `0009`'s name — `0010` is actually the soft-delete migration). tern versions by registration order, not name, so these are harmless to versioning but trip a strict check. So the guard is an **exact duplicate-registration** check (keyed on name+SQL): it catches the double-registration failure mode you flagged (which would shift every later version number) without false-positiving on the distinct-but-same-labeled pair.

If you'd prefer strict name uniqueness, I can rename those two labels as a follow-up — safe, since tern tracks by version number, not name — and the `0010` mislabel is worth fixing regardless. Everything else is unchanged from the version you approved.

danielbowne (approved):
Approving. Verified the merge fix: no `getMigrator` references remain, `go build ./...` is clean, and all 53 numbered migrations register through `appendMigration`. Good catch on the #490 collision - that would have broken the build on merge.

The duplicate-registration guard is narrower than strict name uniqueness, and your reasoning holds: tern derives versions from registration order, so the `0002`/`0004` shared label is harmless and a strict check would fail on `main` today. The exact-repeat key covers the failure mode I was worried about. Worth noting the name+SQL key will not catch a copy-paste duplicate with a modified comment, but that is a much less likely mistake than a straight double registration.

Please open the follow-up for the label cleanup - the `0010` mislabel is misleading when reading the migration list, and renaming is safe since tern tracks by version number.

Issue #395, opened 2026-07-07
Canonicalize HHS system attributes and serve them as validated select boxes


danielbowne:
### Feature Type
Database Component (plus UI + a data cleanup)

### Workflow Description
The HHS onboarding fields on a FISMA system (FIPS impact level, system type, HVA, cloud system, cloud service model, GOCO/COCO/GOGO, system operator, legacy) are free-text today, both in the edit form and the main systems table. They should be validated select boxes driven by the backend from a canonical set of allowed values - the same pattern #392 established for `datacenterenvironment` - so a user picks from a controlled list instead of typing, and the same field reads consistently across every OpDiv.

These are standard system attributes, not HHS-specific: FIPS-199 impact level, system type, etc. apply to every OpDiv including CMS. CMS values will be seeded separately.

The blocker is data quality. The values loaded from the HHS source are inconsistent, so "distinct values in the data" cannot be offered directly as options. Observed today:

- `fips`: High, Low, Moderate, **Undefined**
- `system_type`: 14 variants incl "Minor Standalone" / "Minor Stand-alone" / "Minor Application (Standlone)" (typo) / "Minor Application (Standalone)", "other" / "Other" / "Others -"
- `cloud_service_model`: "Saas" vs "SaaS", "IaaS,PaaS" vs "IaaS/PaaS" vs "IaaS,PaaS,SaaS"
- `cloud_system`: No / NO / Yes / YES
- `legacy`: no / No / yes / Yes
- `system_operator`: Agency / contractor / Contractor
- `goco_coco_gogo`: COCO / GOCO / GOGO / GOVO / No

So the work is: agree a canonical allowed set per field, clean the existing rows to it (upstream in the HHS source where possible so refreshed loads stay clean), then serve and validate against it.

### Acceptance Criteria
- [x] A canonical allowed-value set is agreed per enumerable field (FIPS, system type, HVA, cloud system, cloud service model, GOCO/COCO/GOGO, system operator, legacy)
- [x] Existing rows are canonicalized to that set (case/typo/near-duplicate collapse); source pipeline updated so refreshed loads arrive clean
- [ ] Backend serves the allowed values per field (extend the reference-data approach from #392) and validates writes against them
- [ ] Edit form renders these fields as select boxes for all OpDivs (not HHS-gated); main systems table shows them - tracked in CMS-Enterprise/ztmf-ui#460
- [x] `null` is handled where a value is not yet captured (e.g. CMS before its seed)

### Additional Context
[...]

PR #488, opened 2026-07-27
feat(#433): retype system-metadata columns to typed columns (booleans, CHECK-enums, text[])

### LINKED ISSUE #433: Backend: typed columns, constraints, and migration for system metadata
Sub-issue of #431 (metadata standardization epic).

## Scope

Retype the extended system metadata columns and migrate the existing data. The vocabulary endpoint and server-side validation moved to #395; this ticket is the schema change only.

Types below were verified against `stg_hhs_fismasystems` (1114 rows, the actual insert source) and re-verified against the live SDL load table `WIP_HHS_LOAD_FISMASYSTEMS` - zero drift between the two, same row count and same distributions across all eight columns. Zero off-vocab, case, or whitespace variants, so the canonicalization done under #395 held and every cast is clean.

- **Typed columns / constraints:**
  - Booleans (`*bool`): `hva`, `cloud_system`, `legacy`.
  - Enums: `fips` (Low/Moderate/High); `system_type` (Major Application / Minor Application / Minor Standalone / General Support System / Enterprise / Local / Other); `system_operator` (Agency/Contractor); `goco_coco_gogo` (GOCO/COCO/GOGO).
  - Multi-select stored decomposed (not one combined string): `cloud_service_model`, tokens IaaS / PaaS / SaaS / Other. 21 rows carry combined values today (`PaaS/SaaS` 8, `IaaS/PaaS/SaaS` 6, `IaaS/PaaS` 6, `IaaS/SaaS` 1) - split on `/`.

- **Every column is nullable and must stay nullable.** There is no NOT NULL candidate in the set. Null is a real value here meaning "not captured", and it is common: `cloud_service_model` 60% null, `legacy` 47%, `goco_coco_gogo` 43%, and even `hva` and `fips` have a null row each. Null must not be coerced to No or to a default. Constraints take the form `CHECK (col IS NULL OR col IN (...))`.

- **Migrate existing values in place + backfill.** Values are already canonicalized, so this is a type conversion rather than a cleanup.

## Never enforce the dependency rule as a DB constraint

#395 defines the rule `cloud_system = No` implies `cloud_service_model` and `cloud_vendor` are NULL. **13 existing rows violate it** (of 590 `No` rows, 13 carry a service model and 12 carry a vendor). This rule must stay application-level write validation in #395 and must never become a CHECK constraint on the table. As write validation the grandfathered rows sit untouched until someone edits them. As a CHECK it fails at migration time on those 13 rows and takes the entire retype down with it.

Those 13 rows are deliberately not being cleaned as part of this ticket. The fix isn't mechanical: either the flag is wrong or the service model is stray, and the flag is likelier wrong since nobody types "SaaS" for a non-cloud system by accident. The inverse also fails (73 `Yes` rows carry no service model, 191 carry no vendor), so neither field is authoritative and absence proves nothing. That's a semantic call for whoever owns the inventory, not a migration step.

The retype itself is unaffected: it's varchar to bool/enum on the flags and enums, and `cloud_service_model` decomposition doesn't read `cloud_system`.

## Out of scope

- `cloud_vendor` stays free text. 109 distinct values, no canonical set exists. Do not fold it into the enum work.
- `system_type = Other` is 231 rows (21%), the second-largest bucket after Major Application. It is a legitimate canonical value, not a cleanup backlog.

## SDL sync

These columns sync to Snowflake (`CORE.ZTMF_FISMASYSTEMS`) via the PG->SDL lambda, and the varchar to boolean/enum flip will change the synced values (Yes/No becomes true/false). We are explicitly accepting a broken refresh here: that data only reloads after a data call change, nothing reads or reports on it yet, and the export gets patched downstream after the retype lands. This ticket does not block on the lambda or the Snowflake column.

## Notes

- Optional bool fields must be `*bool` with conditional `.Set()` on update. A plain `bool` clobbers to false on a PUT that omits it.
- Depends on #395 for the canonical value set. Does not depend on the frontend (CMS-Enterprise/ztmf-ui#460) or the insights pipeline.

costacalvin:
## Summary

Sub-issue of the metadata-standardization epic (#431). Retypes the free-text HHS metadata columns on `fismasystems` into typed, self-validating columns. This is the **retype** half of the backend work; the vocabulary endpoint + friendly validation are #395 (stacked on this branch).

## What changed

**Migration `0052fismasystemsretype.go`** — canonicalizes existing values in place, then retypes:
- `hva`, `cloud_system`, `legacy`: `varchar` → `boolean`. **`null` stays "unknown"** and is never coerced to `false` (epic rule).
- `cloud_service_model`: `varchar` → `text[]` (decomposed, per the epic's "store decomposed, not one combined string"), with a CHECK on the elements (`IaaS`/`PaaS`/`SaaS`/`Other`).
- `fips`, `system_type`, `system_operator`, `goco_coco_gogo`: stay `varchar` with NULL-tolerant CHECK constraints (house style, matching `0046`/`0048`).
- Normalization collapses the documented off-canon spellings (`GOVO`→`GOGO`, `Minor Stand-alone`→`Minor Standalone`, etc., case/whitespace-folded); anything unmappable → NULL so the CHECKs can't fail the migration. `down` reverses cleanly.

**Model / `Save()`** — the three booleans become `*bool` and `cloud_service_model` becomes `[]string`; writes preserve the omit/clear/set three-state (`emptyToNil` for the array). Integration tests cover the typed round-trip, unknown-stays-NULL, the partial-update nil-guard, empty-slice clear, and CHECK rejection.

## Contract / coordination

- **API contract change:** these fields now serialize as `boolean` / array instead of strings. The frontend (ztmf-ui#626) consumes the new shapes.
- **SDL/Snowflake:** the varchar→typed change breaks the PG→Snowflake export (`CORE.ZTMF_FISMASYSTEMS`) in the private `ztmf-insights` repo. This is accepted (nothing reads it yet); a follow-up patches the export downstream.

## Verification

`go build`/`vet`/`test -short` green. Migration + typed round-trip need a seeded DB (CI).

## Notes for reviewers

- Data is mostly NULL for these columns today (HHS not yet loaded; CMS rows NULL), so the in-place normalization touches few rows.
- **#395 is stacked on this branch.** Merge this first, or merge #395 directly (it contains these commits).

Refs #431. Closes #433.

### DIFF
--- a/backend/cmd/api/internal/migrations/0052fismasystemsretype.go
+++ b/backend/cmd/api/internal/migrations/0052fismasystemsretype.go
@@ -1,0 +1,152 @@
+package migrations
+
+func init() {
+	getMigrator().AppendMigration(
+		"retype fismasystems HHS metadata columns: booleans, enum CHECKs, and a cloud_service_model text[]",
+		`
+-- ztmf#433: the HHS onboarding metadata columns were added as free-text
+-- varchar(255) in 0043. This migration makes them typed and self-validating at
+-- the database layer:
+--   - booleans (hva, cloud_system, legacy): varchar -> boolean, null = unknown
+--     (the epic rule is do NOT coerce a missing value to No);
+--   - enums (fips, system_type, system_operator, goco_coco_gogo): stay varchar
+--     with a CHECK constraint (house style, matching 0046/0048 - avoids the
+--     native ENUM ALTER TYPE dance);
+--   - cloud_service_model: varchar combo -> text[], with a CHECK on the elements.
+--
+-- Existing rows are canonicalized in place first so the type casts and CHECKs
+-- cannot fail the migration. Postgres data is mostly NULL for these fields today
+-- (HHS not yet loaded; CMS rows NULL), so this touches few rows, but the mapping
+-- covers every off-canon spelling catalogued in ztmf#395. Anything still
+-- unmappable is set to NULL ("not yet captured") rather than blocking startup.
+--
+-- The vocabulary endpoint, friendly 400 messages, and the cloud_system=No
+-- cross-field rule are ztmf#395, layered on top of these typed columns.
+--
+-- NOTE: this changes the JSON API contract (bool / array instead of strings) and
+-- breaks the PG->Snowflake SDL export (CORE.ZTMF_FISMASYSTEMS) in the private
+-- ztmf-insights repo. Both are expected and coordinated (danielbowne accepted a
+-- temporary SDL break; the export is patched downstream).
+
+-- 1. Canonicalize enum values in place, collapsing the ztmf#395 variants and
+--    nulling anything off-canon so the CHECKs below hold.
+-- Fold on lower(btrim(...)) like the other three enums below, so casing and
+-- stray whitespace ('major application', ' Major Application') canonicalize
+-- instead of silently dropping to NULL.
+UPDATE public.fismasystems SET system_type = CASE lower(btrim(system_type))
+    WHEN 'minor standalone'               THEN 'Minor Standalone'
+    WHEN 'minor stand-alone'              THEN 'Minor Standalone'
+    WHEN 'minor application (standalone)' THEN 'Minor Standalone'
+    WHEN 'minor application (standlone)'  THEN 'Minor Standalone'
+    WHEN 'minor application'              THEN 'Minor Application'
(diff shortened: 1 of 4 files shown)


jsos3-cms:
Superseded by the combined in-repo rehome #490 (squash-merged, `c689d8b`), which carried this retype work and closed #433. Original authorship preserved in the rehome. Closing the source draft — thanks @costacalvin!

PR #489, opened 2026-07-27
feat(#395): serve system-metadata vocab endpoint + validate writes

costacalvin:
## Summary

Sub-issue of the metadata-standardization epic (#431). Serves the canonical system-metadata vocabulary via the API and validates writes against it, on top of the #433 typed columns. Unblocks the frontend selects (ztmf-ui#626).

> **Stacked on #433.** This branch contains #433's commits plus #395's. The team can either **merge #433 first** (then this rebases to just its own delta) or **merge this directly** (which brings the retype too). Review the #395-specific commits: `serve system-metadata vocab endpoint + validate writes` and `support clearing tri-state booleans to unknown`.

## What #395 adds

- **`GET /api/v1/systemattributes`** — reference table (`0053systemattributes.go`) + endpoint serving the canonical option set per field (`field/value/description/selectable/ordr`), filterable by `?field=` and `?selectable_only=`. Mirrors the #392 `datacenterenvironments` pattern. The seeded set is kept in sync with #433's CHECK constraints (canonical only — no off-canon rows).
- **Friendly write validation** in `Save()` — enum fields and each `cloud_service_model` element are checked against the table for a field-level 400 before the DB CHECK would surface a raw error; plus the `cloud_system = No ⇒ cloud_service_model / cloud_vendor empty` cross-field rule (#431).
- **Tri-state boolean clearing** (per danielbowne's ztmf-ui#460 Yes/No/**Unknown** correction) — a boolean sent as explicit `null` clears to NULL (Unknown), while an omitted key is left unchanged (preserves the ztmf#442 partial-PUT protection). Presence is captured at the controller (`presentJSONKeys`) and passed to `Save` via `WithPresentBoolFields`; the model field stays a clean `*bool`.

## Frontend contract (ztmf-ui#626)

- Fields serialize typed: `hva`/`cloud_system`/`legacy` as `boolean`|`null`; `cloud_service_model` as a `string[]` array (not a slash-joined string).
- The endpoint serves rich rows; derive the selects from them.
- **Selecting "Unknown" must send the field as explicit JSON `null`** (not omit the key); omitting means "leave unchanged."

## Verification

`go build`/`vet`/`test -short` green (incl. a DB-free unit test for the null-vs-absent presence logic). Endpoint, validation, and bool-clear behavior have integration tests that run against a seeded DB (CI).

Refs #431, #433. Closes #395.

PR #490, opened 2026-07-27
Retype system-metadata columns + vocab endpoint & validation (rehome of #488/#489)

### LINKED ISSUE #395: Canonicalize HHS system attributes and serve them as validated select boxes

### LINKED ISSUE #433: Backend: typed columns, constraints, and migration for system metadata

jsos3-cms:
## Summary

In-repo rehome of the metadata-typing backend, combining #488 (#433 retype) and #489 (#395 vocab + validate) into a **single PR** per team decision — #489 already contains #488's retype commit, so this closes both. Authorship preserved on all three cherry-picked commits (@costacalvin); the original PRs #488 and #489 stay open as drafts until this merges.

**Closes #433. Closes #395.** Part of epic #431.

## What's here (three contributor commits, cherry-picked verbatim)

- **#433 retype** (`fe7898d3`): migration `0052` retypes the free-text HHS metadata columns to typed columns (booleans, CHECK-enums, `text[]`); `fismasystems.go` model updated. API contract change — these fields now serialize as `boolean`/array.
- **#395 vocab + validate** (`d9eb0ad2`, `a0a6122`): `GET /api/v1/systemattributes` (migration `0053` reference table), write-validation against the canonical vocab, and tri-state boolean clear-to-unknown.

## Reviewer fix applied on top

- **Seed data:** `_test_data_empire.sql` inserted `cloud_service_model` as bare scalars, invalid for the new `text[]` — it broke the seeded integration/e2e/dev DBs and would fail the CI `make test-integration` job. Fixed to array literals. (Authorship of the three feature commits preserved beneath this reviewer commit.)

## Status — draft pending full local end-to-end

⚠️ **This is an API-contract change; it must not deploy to prod ahead of the frontend that consumes the new shapes** (ztmf-ui#628, the rehome of #626). Being validated now with a **full local end-to-end test of this backend against ztmf-ui#628** before this comes out of draft.

Known follow-ups from review (non-blocking, tracked here): the `cloud_system=No ⇒ cloud_service_model/cloud_vendor empty` cross-field rule only checks fields in the current request (partial-update gap vs the #431 invariant); a `legacy` help-text wording TODO in migration 0053 to confirm with the stakeholder; and issue-number references to strip from code comments.

## Downstream: SDL → Snowflake sync (accepted break, tracked follow-up)

The varchar→boolean/enum/`text[]` retype flips the values that sync to `CORE.ZTMF_FISMASYSTEMS` through the PG→SDL lambda (`Yes/No` → `true/false`; `cloud_service_model` → array). Per the #433 discussion this is an **accepted, temporary break** — nothing reads that data yet and a failed refresh is tolerable — with the export patched downstream after this lands. Tracking it here so it isn't lost:

- [ ] Patch the PG→SDL export + Snowflake destination column types/consumers to match the retyped columns (post-merge follow-up; not a blocker for this PR).


jsos3-cms:
## Full local end-to-end result (this BE ↔ ztmf-ui#628)

Ran a full local e2e: stood up this branch on a fresh isolated Postgres (migrations 0052/0053 + the fixed empire seed — all clean), and drove the metadata flow with live `curl` against the running server.

**Backend contract: verified correct.**
- Correctly-typed PUT persists: `hva:true`, `cloud_system:true`, `legacy:false`, `cloud_service_model:["IaaS","PaaS"]` → `204`, and a re-GET returns exactly those typed shapes.
- Clear-to-Unknown via explicit `null` (`"hva": null`) → `204`, persists `null`.
- Off-canon values rejected: `"fips":"SuperHigh"` and `cloud_service_model:["…","NotARealModel"]` → `400 invalid input` with the offending field echoed.
- Enum fields (`fips`/`system_type`/`system_operator`/`goco_coco_gogo`) remain `varchar`+CHECK — unaffected by the retype, `""`-clears still work.
- `GET /api/v1/systemattributes?selectable_only=true` → the canonical 26-row vocabulary.

**Frontend dependency (ztmf-ui#628): needs alignment before the pair can merge.** #628 was built on the pre-retype string contract — it sends `hva:"Yes"` (string) and `cloud_service_model:"IaaS/PaaS"` (slash-joined string), which this branch's strict decode `400`s, and its load path calls `.split()` on what is now an array (runtime crash for any system with a populated `cloud_service_model`). The fix is FE-side (retype the four fields, send booleans + array, consume booleans + array); it's being applied on #628 now, then the e2e will be re-run to confirm the pair is green.

⚠️ **Merge order:** this is a breaking API-contract change. It must not merge to `main` (which auto-deploys prod) ahead of the aligned ztmf-ui#628 — otherwise the current prod frontend 400s on every metadata save.

jsos3-cms:
Cherry-picked @costacalvin's emberfall fixture fix (`6a628814`, from the `feat/395-vocab-validate` branch / #489) onto this branch as `d48df33` — it aligns the emberfall E2E fixtures to the typed metadata contract: `hva`/`cloud_system`/`legacy` → boolean, `cloud_service_model` → `text[]` array literals, and the clear-test to explicit `null`. Author preserved; committer signed.

Only that one commit was taken — the seed fix (`657a891` on Calvin's branch) is already here as `8b0be449`, so it was intentionally skipped. Verified locally: `go build ./...` and `go vet` clean, `emberfall_tests.yml` parses as valid YAML. The emberfall E2E smoke check passes green with it.

danielbowne:
Reviewed this against the retype ticket (#433) and ran a pre-flight on the dev dataset. Two notes, one clean bill of health and one thing I'd like added before merge.

**Pre-flight came back clean, the migration is safe on current data.**

Dev is at `dbversions` 48, so nothing here has landed on the shared database yet and there is no drift to unwind. Current distributions across the 255 rows in Postgres (CMS only, HHS inventory not loaded to PG yet, which matches the assumption in 0052's header comment):

| column | values |
| --- | --- |
| `hva` | No 242, NULL 11, Yes 2 |
| `fips` | Moderate 225, High 21, Low 9 |
| `cloud_system`, `legacy`, `cloud_service_model`, `system_type`, `system_operator`, `goco_coco_gogo` | all NULL |

Zero off-canon values. Every cast in 0052 is either a no-op or a clean fold, and no row hits the `ELSE NULL` fallback. Worth re-running the same check against prod before this merges, but on dev there is nothing to salvage and nothing that would abort the migration.

**The one gap: the canonical value sets now live in two places with nothing enforcing agreement.**

0052 hardcodes them as CHECK constraints, and 0053 seeds the same sets into `systemattributes`. The header comment on 0052 says the lists "must stay in sync with the ztmf#395 systemattributes selectable=TRUE seed", which is the right instinct, but a comment is the only thing holding it together. Adding or renaming a value means editing a migration and a seed in lockstep, and if they drift the failure is unpleasant: the vocabulary endpoint offers a value the database then rejects with a constraint violation on save. That reads to the user as a broken form, and it would not show up in any test we currently have.

Suggest an integration test that reads the CHECK constraint definitions out of `pg_constraint` and asserts, per field, that the permitted set equals the `selectable = TRUE` rows in `systemattributes`. That turns a silent runtime failure into a red build the moment the two drift. Same shape for the `cloud_service_model` element CHECK against its own seeded set.

Everything else here holds up well. The presence-tracking approach for the booleans is the right call, and it is the part I was most worried about going in: `*bool` alone cannot distinguish an omitted field from an explicit null, so without it "set to Unknown" would have been unreachable on a PUT and the tri-state would have been boolean in name only. Good call also on canonicalizing in place before the casts rather than trusting the source data to be clean, and on the DOWN migration documenting that it restores the canonical form rather than pretending the original free text is recoverable.

Separately, #433's body has fallen behind this branch and disagrees with it in a couple of places. I'm commenting there rather than here, since none of it changes the code.

jsos3-cms:
Thanks @danielbowne — genuinely useful review, and thanks for running the dev pre-flight.

**On the one gap (CHECK ↔ vocab drift): added a test that closes it.** `TestMetadataVocabMatchesCheckConstraintsIntegration` (in `backend/internal/model/systemattributes_integration_test.go`) reads the live CHECK definitions out of `pg_constraint` (`pg_get_constraintdef`) and asserts, per field, that the permitted set equals the `selectable = TRUE` rows in `systemattributes` — the four enum CHECKs (`fips`, `system_type`, `system_operator`, `goco_coco_gogo`) plus the `cloud_service_model` element CHECK. If 0052 and 0053 ever drift, it fails with a message naming the field and both sets, instead of the silent offer-then-reject on save. The booleans carry no CHECK (the boolean type is their constraint), so they have no permitted set to compare and are out of scope. Verified against a fresh, freshly-migrated isolated DB via `make test-integration` — green, and the two sets agree today, so it locks that in.

**On the prod pre-flight:** agreed it's worth doing. The migration is safe on any data (every cast is a no-op or a clean fold; off-canon folds to NULL rather than aborting), so this is a confirm-the-distribution check, not a migration gate — I'll leave the actual prod run to someone with prod DB access as a pre-merge step rather than reach into prod from here.

**Everything else stands as-is** — no changes to the presence-tracking, the in-place canonicalization, or the DOWN migration.

Two open threads, unchanged: the `legacy` help-text wording is still the `TODO` on the `value=''` row in 0053 — send me the final copy and I'll drop it in; and the coordinated merge with ztmf-ui#628 still holds (neither prod deploy is safe alone).

danielbowne:
Backend looks ready to me, and I'm good for it to go with ui#628. One small thing I'd like changed first, details at the bottom.

The gap I raised, the CHECK constraints and the vocabulary seed drifting apart with nothing catching it, is closed by the new integration test. CI is green and the branch is mergeable.

Pre-flight came back clean on both sides. Prod is 255 CMS rows with no off-canon values in any of the eight columns and nothing that would fold to null on the cast. Tonight's HHS load is 1,114 rows and every column conforms, including the 21 combined `cloud_service_model` values, which split cleanly.

That second check is the one that mattered. 0052 canonicalizes rows already in the table, but anything inserted after the constraints go live gets rejected rather than cleaned up, so the load data had to be canonical going in. It is.

**The change: drop the `description` column rather than fill in its TODO.**

I went looking to send you the final `legacy` wording and found the row is already populated with the older technical-criteria definition (end-of-life software, unsupported vendor, no active development team). That framing was corrected on #395 in July. `legacy` is a funding disposition: HHS recording that it does not intend to spend money remediating that system, deliberately not a statement about the system's age or technology.

Rather than swap the string, I think the column should come out. Nothing consumes it. The frontend fetches `description` but never renders it, it only uses `selectable = FALSE` to keep help rows out of the dropdown. So today it is an unused column whose sole content is a definition we no longer stand behind, and it keeps the vocabulary endpoint's payload shape diverging from the `datacenterenvironments` pattern it was meant to mirror.

Concretely: drop `description` from the table in 0053, drop the `('legacy', '', ...)` help row and the `value = ''` convention that exists to support it, and drop `description` from the `SystemAttribute` type in ui#628. The drift test is unaffected since it only compares `selectable = TRUE` rows. Help text then lives in `fieldConfig` alongside the labels, which is where the other field copy already sits.

Not a correctness issue and nothing user-facing today, so if you'd rather land this as-is and strip it in a follow-up, I won't hold the merge over it. But it is a smaller diff now than after it ships.

jsos3-cms:
Implemented @danielbowne's recommendation that the vocabulary endpoint be a plain allowed-values list (the `datacenterenvironments` shape) with no per-value help text — field help copy now lives in the frontend via the `FieldConfig` `helpText` prop that shipped in ztmf-ui#628. Changes:

- Dropped the `description` column and the `value=''` help row from the systemattributes seed (migration 0053).
- Removed the `Description` field from `model.SystemAttribute` and regenerated `openapi.yaml` — the endpoint response no longer carries `description`.
- Updated the affected integration test. The CHECK ↔ vocab drift-guard test is unaffected (it compares the `selectable=TRUE` values, not help rows).

This also retires the seeded `legacy`-definition placeholder (the straw-man that carried the old `TODO`) — that copy now lives in the UI per your confirmed wording.

Verified: `go build` / `vet` clean, OpenAPI spec regenerated with no drift, and `make test-integration` green against a fresh migrated DB.

MackOverflow (approved):
clean, lgtm.

PR #492, opened 2026-07-29
Harden Aurora backups: 35-day PITR, AWS Backup enrollment, and monthly logical dumps

danielbowne:
## Why

An unintended survey response edit was discovered five weeks after it happened, and there was no backup old enough to recover from. This is the infrastructure side of making that recoverable.

The cluster is about 105 MB and Aurora meters backup storage on change volume rather than volume size, so cost is not a decision input here - the whole thing lands under $0.15/month worst case. Simplicity and late-discovery coverage are what drove the choices.

## What

Three legs, because no single one covers the observed failure:

**1. Point-in-time recovery raised to 35 days in prod.** Dev and impl stay at 1 day; neither holds data worth recovering at that horizon. 35 is the Aurora ceiling and deliberately not enough on its own - the incident surfaced at exactly 35 days, so PITR alone would have covered it with zero margin and missed it entirely at six weeks. That is what justifies the other two legs.

**2. Enrollment in the existing CMS OIT `Daily15_Weekly90` plan.** This account already has seven backup plans executing daily against a locked vault. Every one selects purely on an `AWS_Backup` tag, and nothing in the account carries it, so no backup job has ever run. Enrollment turned out to be one tag rather than a new backup stack. That buys 15 daily plus weekly recovery points held 95 days.

**3. A monthly `pg_dump` to a new Object Lock bucket**, driven by EventBridge Scheduler running the existing ops image with a different entrypoint. This is the only copy with no expiry, which matters given the usage pattern - fourteen months elapsed with two writes, so an edit can plausibly sit unnoticed for a year. It is also the only artifact you can read without standing up a cluster, which is the point: recovery for a single bad row is never "restore the cluster", since every Aurora restore path produces a new cluster and rolling prod back weeks would discard everything else. The real procedure is to stand up a reference copy, diff the rows, and apply a targeted UPDATE.

Coverage by discovery lag:

| lag | PITR @ 35d | `d15_w90` chain | monthly dump |
| --- | --- | --- | --- |
| <= 35 days | yes | yes | yes |
| 36-95 days | no | yes | yes |
| 96 days - 1 year | no | no | yes |
| > 1 year | no | no | yes |

## Drift closed on the cluster

Separate from the backups, and the reason `rds.tf` has more diff than you might expect:

- Backup and maintenance windows are pinned. Both environments were running AWS-assigned random windows that differed from each other (prod `06:19-06:49` / `sun:07:05`, dev `08:37-09:07` / `tue:10:22`).
- The cluster parameter group is pinned to the value already in use, so a manual swap shows as drift instead of persisting silently.
- `engine_version` is now ignored via `lifecycle`. With `auto_minor_version_upgrade` on, AWS applies minor patches on its own schedule, and the pinned `16.11` literal meant the next patch would make every subsequent plan propose a downgrade. Refresh still records the real version, so the instance's reference to it continues to resolve.
- `deletion_protection` is on, which does mean `terraform destroy` on the cluster now requires clearing the flag by hand first.

## One design note worth flagging

The obvious staleness alarm - have the dump publish a success pulse, alarm on missing data over 40 days - is not expressible in CloudWatch. An alarm's total evaluation range (`period` x `evaluation_periods`) is capped at 604800 seconds, so the longest possible lookback is 7 days, and a monthly pulse leaves about 29 consecutive days with no datapoint that any legal alarm reads as breaching. `terraform validate` does not catch this because it never calls the API; it fails at apply.

So freshness is measured rather than inferred. A small daily probe task lists the bucket and publishes the age of the newest dump, and the alarm reads that over a single 1-day period. Side benefit: it trusts the bucket rather than the previous job's own report, so an upload that was reported successful but never landed still trips it.

## Verification

- `terraform fmt`, `terraform validate`, `tflint` clean.
- `terraform plan` against **dev**: 20 to add, 2 to change, 0 to destroy. Both changes in-place on the cluster and its instance, no replacement. `backup_retention_period` correctly shows no change in dev, and `engine_version` shows no change, confirming the `ignore_changes` behaves.
- The ops image was built and run locally to confirm both scripts are present and executable, the RDS trust bundle is in place, and `date -u -d` parses the timestamp format S3 returns.

**The dev image is preloaded.** `ztmf/ops:c61f01a` is pushed to dev ECR and SSM `ztmf_ops_tag` points at it, because `ops-image.yml` only publishes on push to `main` while the infra applies on PR - without preloading, the task definitions would have registered against an image with no scripts in it. Re-planning after the preload shows 21 to add, 2 to change, 1 to destroy, where the destroy is `aws_ecs_task_definition.ztmf_ops` rolling to a new revision. The next `ops-image.yml` run on `main` will overwrite that SSM value with its own build, which is fine.

Prod has not been planned yet.

## On the `needs-refinement` label

@jsos3-cms - you are the Terraform expert here and I would rather you shape this than rubber-stamp it. The label is not "this is half-finished" - the functionality is complete and verified against dev. It is that this is bootstrap: I picked defaults where the strategy left room, and those are the parts worth your eye rather than the plumbing.

Specifically, the judgment calls I would like you to push back on:

- **Object Lock is GOVERNANCE, not COMPLIANCE**, at 365 days. COMPLIANCE cannot be overridden by anyone including account root for the full retention period, which felt like a large commitment to make on an account we do not solely control. If you disagree, that is a one-line change.
- **Dev stays enrolled in `d15_w90`** (about two cents a month) on purpose, because the tag-selection path has never executed in this account and dev is where we find out whether the CMS-managed selection role can actually back up Aurora. Drop it if you would rather not pay for that signal.
- **Pinning the backup and maintenance windows** changes them in both environments on first apply. Non-disruptive, but it is a real change and the specific times are arbitrary.
- **`ignore_changes = [engine_version]`** versus disabling `auto_minor_version_upgrade` and managing the version explicitly. I chose to keep patching automatic and stop Terraform fighting it, but the other direction is more auditable for a FISMA system and I would defer to you.
- **A second task definition for the daily probe.** EventBridge Scheduler's ECS target has no container override, so a different command means a different task definition. If there is a cleaner idiom, I would rather use it.

Worth knowing before it applies: an unmerged PR's Terraform in shared dev gets reverted by the next PR's apply, since those run from `main`.

Do not hold back on restructuring. Refine further if you want to.

### DIFF
--- a/backend/ops/Dockerfile
+++ b/backend/ops/Dockerfile
@@ -4,10 +4,21 @@ RUN apk add --no-cache \
         aws-cli-v2 \
         bash \
         ca-certificates-bundle \
+        coreutils \
         curl \
         jq \
         postgresql-16-client
 
+# RDS global trust bundle, so psql/pg_dump can run with sslmode=verify-full
+# instead of merely encrypting. Not present in the distro CA bundle.
+ADD --chmod=0444 https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem \
+    /etc/ssl/certs/rds-global-bundle.pem
+
+COPY --chmod=0555 dump-db.sh /usr/local/bin/dump-db.sh
+COPY --chmod=0555 check-dump-age.sh /usr/local/bin/check-dump-age.sh
+
 USER nonroot
 
+# Default is the interactive shell target used by scripts/db-tunnel.sh. The
+# scheduled monthly backup task overrides this with dump-db.sh.
 CMD ["sleep", "infinity"]
--- a/backend/ops/check-dump-age.sh
+++ b/backend/ops/check-dump-age.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+#
+# Daily staleness probe for the monthly database dumps.
+#
+# Publishes ZTMF/Backup DaysSinceLastDump so the freshness of the backups can
+# be alarmed on. This exists as a separate daily job rather than as a heartbeat
+# emitted by the dump itself because CloudWatch caps an alarm's total
+# evaluation range (period x evaluation_periods) at seven days. A monthly
+# success pulse leaves ~29 consecutive days with no datapoint, so no legal
+# alarm can distinguish "between scheduled dumps" from "dumps stopped
+# happening". A daily gauge of the age of the newest object can.
+#
+# It reads the bucket rather than trusting a metric the dump job wrote, so it
+# also catches an upload that was reported as successful but did not land.
+#
+# Required environment (supplied by the task definition):
+#   DUMP_BUCKET   bucket holding the dumps
+#   ENVIRONMENT   dev | prod (key prefix and metric dimension)
+#   AWS_REGION    region for the AWS CLI calls
+
+set -euo pipefail
+
+: "${DUMP_BUCKET:?DUMP_BUCKET is required}"
+: "${ENVIRONMENT:?ENVIRONMENT is required}"
+
+readonly METRIC_NAMESPACE="ZTMF/Backup"
+readonly PREFIX="pgdump/${ENVIRONMENT}/"
+
+log() { echo "[$(date -u +%Y-%m-%dT%H:%M:%SZ)] $*"; }
+
+log "listing s3://${DUMP_BUCKET}/${PREFIX}"
+
+# Only the .dump archives count. A stray globals file without its archive is
+# not a usable backup and must not reset the clock.
+newest="$(aws s3api list-objects-v2 \
+  --bucket "${DUMP_BUCKET}" \
+  --prefix "${PREFIX}" \
+  --query 'sort_by(Contents[?ends_with(Key, `.dump`)], &LastModified)[-1].LastModified' \
+  --output text)"
+
+if [[ -z "${newest}" || "${newest}" == "None" ]]; then
+  # An empty bucket is indistinguishable from a bucket whose contents were
+  # deleted. Both are reported as maximally stale rather than as missing data,
+  # so a brand new environment surfaces immediately instead of looking healthy.
+  log "no dumps found under ${PREFIX}; reporting maximum age"
+  age_days=9999
+else
+  # GNU date, from the coreutils package added to the image for this. The
+  # busybox date that wolfi-base ships by default does not parse the
+  # "+00:00" offset S3 returns in LastModified.
+  newest_epoch="$(date -u -d "${newest}" +%s)"
+  now_epoch="$(date -u +%s)"
+  age_days=$(((now_epoch - newest_epoch) / 86400))
+  log "newest dump ${newest}, ${age_days} days old"
+fi
+
+aws cloudwatch put-metric-data \
+  --namespace "${METRIC_NAMESPACE}" \
+  --metric-name DaysSinceLastDump \
+  --value "${age_days}" \
+  --unit Count \
+  --dimensions "Environment=${ENVIRONMENT}"
+
+log "done"
--- a/backend/ops/dump-db.sh
+++ b/backend/ops/dump-db.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+#
+# Monthly logical backup of the ZTMF Aurora cluster to S3.
+#
+# Run as a scheduled Fargate task off the same ops image used for interactive
+# database access (see infrastructure/backup-dumps.tf). This is deliberately a
+# logical dump rather than another snapshot: recovery for a single bad row is
+# never "restore the cluster", because every Aurora restore path produces a new
+# cluster and rolling production back weeks to fix one row would discard
+# everything else. The real procedure is to stand up a reference copy, diff the
+# affected rows, and apply a targeted UPDATE, so the artifact's readability
+# matters more than its restorability. A pg_dump restores into local docker in
+# minutes; a snapshot needs a new cluster.
+#
+# It is also the only copy with no expiry, which matters given the usage
+# pattern: an unnoticed edit can plausibly sit for a year.
+#
+# Required environment (supplied by the task definition):
+#   DB_SECRET_ID  Secrets Manager ARN of the RDS-managed master credential
+#   DB_ENDPOINT   cluster writer endpoint
+#   DB_PORT       postgres port
+#   DB_NAME       database to dump
+#   DUMP_BUCKET   destination S3 bucket
+#   ENVIRONMENT   dev | prod (key prefix and metric dimension)
+#   AWS_REGION    region for the AWS CLI calls
+
+set -euo pipefail
+
+: "${DB_SECRET_ID:?DB_SECRET_ID is required}"
+: "${DB_ENDPOINT:?DB_ENDPOINT is required}"
+: "${DB_PORT:?DB_PORT is required}"
+: "${DB_NAME:?DB_NAME is required}"
+: "${DUMP_BUCKET:?DUMP_BUCKET is required}"
+: "${ENVIRONMENT:?ENVIRONMENT is required}"
+
+readonly METRIC_NAMESPACE="ZTMF/Backup"
+
+# Assigned then marked readonly separately: `readonly X="$(cmd)"` makes the
+# builtin's exit status the one `set -e` sees, so a failing mktemp would leave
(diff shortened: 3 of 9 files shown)


jsos3-cms:
Thanks for this @danielbowne . Let me start digging in!

jsos3-cms:
Thanks for the writeup, @danielbowne — this is a lot more considered than "needs-refinement" implies, and several things I went in skeptical of held up under checking (Object Lock vs. lifecycle transitions, `arn_without_revision`, the write-only S3 IAM shape, the `mktemp`/`readonly`/`trap` ordering, and the "ALARM then clears" claim). I cross-checked the load-bearing claims against the AWS provider docs and the RDS/S3/CloudWatch references, plus an independent review pass. Verdicts on your five, then the changes I'd ask for.

## Your five judgment calls

1. **Object Lock GOVERNANCE 365d — keep it.** The threat model here is accidental/unnoticed edits, not a malicious admin destroying backups; GOVERNANCE covers that and leaves the documented break-glass path. COMPLIANCE would be over-committing on an account we don't solely control. One thing to keep true: no routine or CI role should hold `s3:BypassGovernanceRetention` — the dump task role is write-only with no delete, which is exactly right.
2. **Dev stays enrolled in `d15_w90` — keep it, strongly.** Two cents a month to prove the CMS-managed selection role can actually back up Aurora via the tag path before prod leans on it is a good trade.
3. **Pinned backup/maintenance windows — keep.** Non-disruptive. Only nit: the monthly dump fires at `06:30`, exactly the end of the `06:00-06:30` backup window — nudge it to `07:00` for a clean gap so a slightly-long backup can't contend.
4. **`engine_version` — keep `auto_minor_version_upgrade = true` + `ignore_changes`, but add drift alerting** (C2 below). I'd rather keep CVE-patch currency than manage the version by hand, but not silently — since `ignore_changes` deliberately blinds `terraform plan`, the signal has to come from outside Terraform.
5. **Second task definition — keep two.** I confirmed `ecs_parameters` has no command/env override, so the templated ECS target genuinely can't vary either per-invocation. The universal-target form (`arn:aws:scheduler:::aws-sdk:ecs:runTask` + `input` JSON with `Overrides`) would collapse it to one task def, but it trades the typed, provider-validated `ecs_parameters`/`network_configuration` blocks for a hand-authored JSON blob — worse for an audited repo. Two small task defs is the readable choice.

## Changes I'd ask for

### C1 — blocker: `check-dump-age.sh` crashes on an empty bucket

`check-dump-age.sh`'s S3-listing JMESPath raises a `JMESPathTypeError` when the prefix is empty, because `list-objects-v2` omits `Contents` entirely (not an empty array). Under `set -euo pipefail` that aborts the script **before** the `if [[ -z "${newest}" || "${newest}" == "None" ]]` fallback that's supposed to publish `age_days=9999` — so on a fresh/empty bucket the daily probe crash-loops and never writes its sentinel or log line (the alarm still ends up correct via `treat_missing_data=breaching`, but for the wrong reason). Null-coalesce before sorting:

```
sort_by(Contents[?ends_with(Key, `.dump`)] || `[]`, &LastModified)[-1].LastModified
```

### C2 — engine_version drift alerting

Since `plan` can't see an AWS-applied patch by design, surface it with a native RDS event subscription to the existing alarms topic. The DB-cluster `maintenance` category carries `RDS-EVENT-0173` (engine version upgraded / patch complete) and `RDS-EVENT-0290` (patched X⇒Y):

```hcl
resource "aws_db_event_subscription" "ztmf_db_engine" {
  name        = "ztmf-db-engine-events-${var.environment}"
  sns_topic   = aws_sns_topic.ztmf_alarms.arn
  source_type = "db-cluster"
  source_ids  = [aws_rds_cluster.ztmf.cluster_identifier]
  # maintenance carries the applied-patch events (RDS-EVENT-0173 / -0290) and
  # the "upgrade available" heads-up (RDS-EVENT-0156).
  event_categories = ["maintenance"]
}
```

Verify RDS can publish to `ztmf_alarms` (same-account default is usually enough; add an `sns:Publish` statement for the RDS service principal if delivery fails). Cluster-level only — don't also subscribe the instance, or you'll get duplicate noise.

### C3 — alarm sliding-window jitter

`ztmf_db_dump_stale` uses `period=86400`, `evaluation_periods=1` on a once-daily metric. The default sliding (non-calendar-aligned) window means routine Fargate start-time jitter can momentarily leave a 24h window with neither yesterday's nor today's datapoint → a ~1-minute false ALARM→OK blip straight to the email SNS. The provider pinned here (`~> 5.82`) has no `evaluation_window`/`WallClockWindow` argument to fix it properly, so widen `period` to `93600` (26h) to absorb the jitter — still far under the 604800s ceiling and far below the 40-day intent.

### C4 — IAM `PassRole` self-containment

`aws_iam_role_policy.ztmf_db_dump_check_scheduler` grants `iam:PassRole` only for `module.db_dump_check.role_arn`, but the check task's `execution_role_arn` is `module.ops_task_execution.role_arn` — authorized only by the sibling `dbDumpSchedulerPermissions` (both attach to the same scheduler role). It works today, but if the dump leg is ever removed the check schedule breaks with an opaque `AccessDenied` on `PassRole`. Add `module.ops_task_execution.role_arn` to the check-scheduler policy's `PassRole` `Resource` so each policy stands on its own.

### C5 — fresh-enable alarm

The alarm sits in ALARM from first enable until the first monthly dump lands (and with C1 unfixed, the probe crash-loops meanwhile) — not "clears once the probe has run," since the probe reports maximum age on an empty bucket. Fix that line in the description, and add a runbook step: after enabling, run one dump immediately (`aws ecs run-task` against the dump task def — the image is preloaded) to seed the bucket and clear the alarm. Do this **only post-merge** — a seeded object in the shared-dev bucket pre-merge can wedge the dev pipeline, since the next PR's apply-from-`main` would try to revert the bucket and Object Lock turns that into a failed destroy.

### C6 — optional / discretion

- Mirror the cluster's "remove this block for a deliberate version change" comment onto the instance's `ignore_changes` block. The instance block is redundant (Aurora won't let the instance version diverge from the cluster) but harmless; the comment prevents a future asymmetric edit.
- The `0.0.0.0/0:443` egress matches the existing `ztmf_sync_lambda` SG, so it's not new posture. If you want to close it repo-wide, adding `"monitoring"` to the `aws_vpc_endpoint.ztmf` `for_each` would drop the NAT path for CloudWatch here and for kion's `PutMetricData` too — marginal cost given nine endpoints already exist. Separate hygiene item, not a blocker.

## Follow-up

The scheduling-convergence work is tracked in CMS-Enterprise/ztmf-misc#169 — migrate the repo's scheduled Lambdas (incl. `ztmf-kion-key-rotate`) onto `aws_scheduler_schedule`, for which this PR is the first reference implementation. Related to this but explicitly not a dependency.

jsos3-cms:
Separately from the design feedback above — the deploy-safety gates to clear before this applies (these are pre-apply checks, not change requests):

- [ ] **Prod plan.** Prod hasn't been planned yet. Run `terraform plan` against prod and confirm the blast radius: `backup_retention_period` `1→35` is the only retention change, the `db_cluster_parameter_group_name` pin shows no change, the backup/maintenance window pins are in-place (not replacements), and `deletion_protection` is in-place.
- [ ] **Parameter group.** Confirm the current `db_cluster_parameter_group` in **both** dev and prod is already `default.aurora-postgresql16` (`aws rds describe-db-clusters`). A cluster on a custom group would be force-reverted to default on apply — disruptive.
- [ ] **Prod ops image.** Confirm the ops image is preloaded to **prod** ECR with SSM `ztmf_ops_tag` pointing at it (dev was preloaded; prod wasn't) before the prod task definitions register against an image with no scripts in it.

jsos3-cms:
@danielbowne checking in on this one. Nothing has moved since the review notes on the 29th, and the PR is still a draft, so I want to make sure it isn't waiting on me.

Where it stands: C1 is the only blocker (the `check-dump-age.sh` JMESPath crash on an empty bucket), and the three pre-apply gates above are still unchecked, notably the prod plan and the prod ops image. Everything else in that comment was design discussion or optional.

Are you picking the C1 fix up, or would you rather I take it? Happy either way, I just want to know whether to plan around it. If this is parked behind other work, say so and I'll leave it alone.

danielbowne:
Picking this back up next sprint. I drafted it a while ago out of concern about our data-loss exposure, and the substance still stands - cleaning it up against current main for a re-release.

PR #377, opened 2026-06-30
Login/OIDC observability for dev: ALB access logs, ELBAuth alarms, and backend 401 instrumentation

jsos3-cms:
## Summary
Entra login on dev was failing, and the failure left no signal: the CloudFront -> internal ALB -> ECS path runs native authenticate-oidc at the ALB, and several backend rejection paths returned 401 without logging. This change closes the visibility gaps and instruments the backend login path so a failure is diagnosable in real time. A follow-up commit will add the targeted Entra fix once this instrumentation confirms the exact rejection cause on dev.

## Changes
- Enable ALB access logs on the internal ztmf-api ALB (per-request actions_executed / error_reason) to the existing ztmf-logs-<account>-use1 bucket and rest-api-alb prefix; the bucket policy and SSE-S3 are already in place, so no bucket change is needed.
- Add a ztmf-alarms SNS topic with ELBAuthError / ELBAuthFailure alarms on the ALB -- the first wired alarm route in this repo -- with a configurable email subscription.
- Instrument SessionHandler's 401 branches to log the failing branch, the error, and which claims were present (booleans only; no token values, no email/UPN/PII).

## Test plan
- [x] ALB access_logs.s3.enabled=true; per-request log objects present under rest-api-alb/AWSLogs/... (applied and verified on dev)
- [x] ELBAuthError / ELBAuthFailure alarms route to the ztmf-alarms SNS topic (verified with a synthetic alarm-state test)
- [x] Okta login succeeds end-to-end (verified on dev)
- [x] After this branch deploys to dev, reproduce the Entra login and read the `login: reject branch=...` line to confirm the exact rejection cause

Relates to the login canary follow-ups #376 and CMS-Enterprise/ztmf-ui#443.

PR #495, opened 2026-07-30
chore(infra): enable Entra dual-IdP in prod (entra_enabled = true)

danielbowne:
Flips `entra_enabled = true` for prod. Same one-line change as #350 did for dev, same file shape, different tfvars.

## Why now

Needed to kick off the **2026 data call**. HHS/OpDiv participants authenticate via Entra — `users.identity_provider` is `okta` for CMS and `entra` for HHS/OpDivs (migration 0030). With the flag false in prod, the `/login/entra*` listener rule doesn't exist and `/api/v1/auth/lookup` falls through to the OIDC-gated `/api/*` rule, so those users have no login path at all. Every non-CMS participant is blocked until this flips.

## Scope

One line plus its comment, `infrastructure/tfvars/prod.tfvars`. No code, no migrations. `dev.tfvars` is untouched (already `true` on main).

## What dev has proven

Dev has run dual-IdP for ~6 weeks. Nothing Entra-related has landed since #408 on 2026-07-09.

Prod also gets the flip in better shape than dev did — dev was flipped bare and shook out five follow-ups afterward, all of which are already on main here:

- #351 — CloudFront `/login/*` behavior to the ALB origin (Entra login was AccessDenied without it)
- #354 — Okta JWT `kid` validation + key-fetch hardening
- #352 — HHS-wide admins can set `identity_provider` on user create
- #377 — ALB access logs, ELBAuth alarms, backend 401 instrumentation
- #408 — logout clearing app session + both ALB OIDC cookies

## Pre-merge blockers

These are prod-account operational state, not code. Not verifiable from the repo, so they need confirming before merge. Status and references updated 2026-07-30.

- [x] **`ztmf_session_signing_key` seeded in prod** (`scripts/bootstrap-entra-secrets.sh`). An empty key fails session minting closed (`secrets.tf:37`). Note this breaks **both** IdPs, not just Entra — post-flip, Okta also routes through `SessionHandler` to mint the app session cookie. An unseeded key means nobody logs into prod.
  - **Still open.** The container exists but there is no evidence a value was written. The 2026-06-15 15:03:20Z timestamp on both secrets falls inside #341's merge-triggered PROD `terraform apply` (15:03:03Z to 15:05:50Z), so it records container creation, not seeding. No workflow invokes the bootstrap script, so CI cannot have written them. Settling this needs `LastChangedDate` against `CreatedDate`, or CloudTrail `PutSecretValue`, neither of which is reachable from a human credential. The script generates this key itself, so no external input is required to close it.
- [x] **`ztmf_entra_oidc` seeded in prod.** Terraform reads this one at plan time (`data.tf:74`), so an unseeded secret fails the apply rather than breaking at runtime.
  - **Still open**, same evidence as above. Closing it needs tenant id, client id and client secret from the Entra tenant owners.
- [ ] **Entra app registration has the prod redirect URI** and prod users are **assigned to the app**. Flagging specifically because of the AADSTS50105 (user-not-assigned) hit on dev on 2026-06-30 — and per the scope note in `monitoring-login-auth.tf:12-16`, the ELBAuth alarms *cannot* catch that class. Entra rejects it before any callback, so no datapoint is ever emitted. It's the most likely first failure on a fresh prod tenant and it will be silent.
  - **Still open, and the only item that is not ours to close.** The URI needed is `https://ztmf.cms.gov/oauth2/idpresponse`. Also worth settling in the same conversation: whether prod gets its own app registration rather than sharing dev's, since sharing means one client secret spans both environments and stops `AUTH_ENTRA_AUDIENCE` distinguishing them.
- [x] **Decide login alarm routing for prod.** `alarm_notification_email` is unset in `prod.tfvars`, and `monitoring-login-auth.tf:31` count-gates the SNS subscription on a non-empty string — so prod gets the topic and the alarms with **zero subscribers**. Dev's `jono@aquia.us` is flagged in its own comment as an interim individual address to be replaced before prod. Deliberately left out of this PR since it needs a real destination decision (shared inbox or Slack webhook); happy to fold it in here or take it as a follow-up.
  - **Done, #496.** Both dev and prod now route to `ISPGZeroTrust@cms.hhs.gov`, which also de-personalises dev. Applied, and both SNS subscriptions confirmed rather than left pending.

### Found during review and fixed ahead of this PR

- **#497** — the session cookie is now host-only. `AUTH_COOKIE_DOMAIN` resolved to the bare apex in prod, so this flip would have handed a live prod session cookie to the dev and impl hostnames. Merged and deployed ahead of this PR, so the condition is prevented rather than remediated. Detail in CMS-Enterprise/ztmf-misc#257 (closed).
- **#498** — `AUTH_COOKIE_DOMAIN` renamed to `AUTH_ORIGIN_HOST`, since after #497 it only feeds the same-origin check.

Tracked separately, neither gating this PR: CMS-Enterprise/ztmf-misc#258 (cross-environment cookie check, blocked *on* this PR) and CMS-Enterprise/ztmf-misc#259 (session tokens are unbound bearer credentials with no revocation path).

## Deploy behavior

Nothing applies while this is open, draft or not. `orchestration-prod.yml` triggers only on `pull_request: closed` + merged. `orchestration-dev.yml` does run on the open PR, but it applies with `dev.tfvars`, so it can't exercise this change — there's no pre-merge plan output for prod. **Merging is the deploy.**

The flag swap is destroy-and-create in a single apply: the two Okta rules (priorities 1-2) are destroyed and the four dual-IdP rules created, so priorities never collide. `moved` blocks in `alb-internal.tf:127-135` keep the un-indexed addresses from churning.

Worth having someone on a prod Okta login the moment it applies — priority 4 takes the ALB out of the `/api/*` auth path and hands the gate to the backend session cookie, which is the highest-consequence part of the change.

Refs #341, #350

### DIFF
--- a/infrastructure/tfvars/prod.tfvars
+++ b/infrastructure/tfvars/prod.tfvars
@@ -4,10 +4,13 @@ domain_name_prefix     = ""
 ecs_service_task_count = 1
 # job_code = "ZTMF_SCORING_USER"
 
-# Entra dual-IdP. Keep false until validated on dev and both secrets are
-# seeded in the prod account (scripts/bootstrap-entra-secrets.sh), then flip to
-# true to enable the second identity provider in production.
-entra_enabled = false
+# Entra dual-IdP. Validated on dev since #350 (2026-06-17), so flipping this to
+# true adds the per-IdP ALB rules, moves /api/* off ALB OIDC to backend session
+# validation, and injects the Entra + session env into the API task. Okta login
+# is unchanged. Required for the 2026 data call: HHS/OpDiv participants
+# authenticate via Entra (users.identity_provider, migration 0030) and have no
+# login path in prod while this is false.
+entra_enabled = true
 
 
 # TLS cert rotation Lambda


jsos3-cms:
Reviewed this against the code it activates rather than the diff alone, since the diff is one flag. The wiring matches what the description says it does, and the gates are clean: `terraform fmt -check` and `terraform validate` pass, `go build ./...` and `go vet ./...` are clean, `go test -short ./...` is 576 passing across 19 packages, and OpenAPI is in sync.

Review surfaced a potential security risk that gates applying this in prod. It is tracked in CMS-Enterprise/ztmf-misc#257, deliberately not detailed here. Please read that ticket before the flag is applied. It does not require reworking this PR; the fix lands separately and should merge first.

One operational note on the apply itself. Flipping the flag destroys `authz` and `ztmf_api` at priorities 1 and 2 while creating four rules at priorities 1 through 4 in the same apply, with no explicit ordering between them. The same graph shape applied cleanly in dev, so this is a readiness note rather than a defect, but be ready to re-apply if `CreateRule` returns a priority collision. Since merge is the deploy here and no CI gate produces a prod plan, running `terraform plan -var-file=tfvars/prod.tfvars` locally beforehand is a cheap look at the exact resource-level plan before it becomes live.

jsos3-cms:
Status update on the gate for this PR. Two of the four items are closed, one is closed by decision, and one remains.

## Resolved

**Alarm routing, #496, merged.** The login and OIDC alarms now route to the shared ISPG Zero Trust inbox in both dev and prod. Prod previously had no destination at all, so the SNS subscription count-gated to zero subscribers and every alarm fired into nothing. Both subscriptions have been confirmed, so delivery is live rather than pending.

**Session cookie scoping, #497, merged.** The application session cookie is now host-only in both setters, so it cannot be delivered to a sibling environment when this flag turns on. Background is in CMS-Enterprise/ztmf-misc#257. Both setters changed together, since a cookie is only removed by an expiry whose `Domain` matches how it was stored, and `sameOrigin` was deliberately left untouched so the Origin check behaves exactly as before. The config-field split described in that ticket is still outstanding and tracked there, not here.

One deploy note carried over from #497: dev sessions minted before that merge were domain-scoped, and a host-only expiry cannot delete them, so logout will not clear those particular cookies. They expire on their own within the 3 hour session TTL, measured from issuance rather than from last activity. Leaving dev quiet for three hours after the merge drains them completely. This is dev-only; prod never issued a domain-scoped cookie because the flag has been false there throughout.

**Prod secret values, closed by decision.** `ztmf_entra_oidc` and `ztmf_session_signing_key` exist as containers in prod and were seeded in mid-June. Verifying the contents is not reachable from a human credential, since the only principal with read access is the Actions OIDC role, whose trust policy is scoped to the GitHub provider. Rather than add a job to read them, we are trusting the values on three grounds: dev runs the same tenant and works today, the seeding is roughly six weeks old against a typical six to twenty four month client secret expiry, and a malformed payload fails the apply at `jsondecode` before creating anything. The residual risk is a well formed but stale client secret, which is a question for the tenant owners rather than something CI can answer.

## Outstanding

**The prod redirect URI must be registered in Entra.** ALB OIDC calls back to `https://<host>/oauth2/idpresponse`. The dev app registration has dev's hostname allowlisted; prod needs `https://ztmf.cms.gov/oauth2/idpresponse` added. Without it every prod login fails at the identity provider with a redirect URI mismatch, before any of this code is reached.

This sits with the team that owns the Entra tenant, so it is the only remaining item that is not ours to close.

Three things worth asking them in the same message, since a single round trip is cheaper than three:

1. Add the prod redirect URI.
2. Confirm the client secret currently seeded in prod is current, and give its expiry date.
3. Confirm the prod user set is assigned to the app registration. Worth being explicit about this one: the alarms added in #496 deliberately do not catch `AADSTS50105 user-not-assigned`, because Entra rejects it before any callback and it produces no `ELBAuth*` datapoint. So a missing assignment looks like healthy infrastructure with users unable to log in and nothing firing.

Also worth settling with them: whether prod reuses dev's app registration or gets its own. Reuse is less work, but it means one client secret spans both environments and it makes the `AUTH_ENTRA_AUDIENCE` pin stop distinguishing them, which is the opposite of what that pin was added for. A separate prod registration is the better posture and is barely more work than adding a redirect URI.

## Go / No-Go

**No-Go until the redirect URI is registered.** Everything on our side is done. Merging this PR performs the prod apply, so there is no window in which it merges without going live, which is why the redirect URI cannot be treated as a follow up.

Once that lands, the recommended sequence is to leave dev quiet for three hours first so the legacy dev cookies from #497 drain, then merge.

Nothing above asks for a change to this PR. The diff is correct as written and the review of it is clean: `terraform fmt` and `validate` pass, `go build` and `go vet` are clean, `go test -short ./...` is 576 passing across 19 packages, and CI is green on all applicable checks.

jsos3-cms:
Tidying this up, since there are a few comments above now and the picture has moved. Treat this one as the current state; the lists in the earlier comments are superseded.

**One correction first.** I said earlier that prod's two secrets were seeded in mid-June, and leaned on that to argue we could trust their contents. That was wrong. The mid-June timestamp is Terraform creating two empty containers during #341's apply, not a value being written. Seeding happens by hand via `scripts/bootstrap-entra-secrets.sh`, which no workflow invokes, and nothing in the repo records it being run against prod. Someone may have run it off the record, but the safe assumption is that it has not happened.

The good news is that being wrong about this is cheap. `ztmf_entra_oidc` is read during the apply, and an unseeded secret has no version, so the read fails and this PR's apply stops before changing anything. The signing key goes into ECS by ARN, so if it is empty the task will not start and the deploy fails. Either way it is a red pipeline rather than an outage.

**Done since this opened.** #496 pointed the login alarms at the shared ISPG inbox in both dev and prod, and both subscriptions are confirmed. #497 made the session cookie host-only and is deployed to prod. Because that one landed first, prod will hand out host-only cookies from the moment Entra turns on, so the cross-environment cookie problem is prevented rather than patched after the fact.

**#498** renames `AUTH_COOKIE_DOMAIN` to `AUTH_ORIGIN_HOST`, tidying up the config field that #497 left with a misleading name. It does not gate this PR and can land whenever.

**What is left is really one conversation with the Entra tenant owners.** Four asks, cheaper as a single message:

1. Add `https://ztmf.cms.gov/oauth2/idpresponse` to the app registration.
2. Hand over tenant id, client id and client secret so prod can be seeded.
3. Confirm the client secret is current, and share its expiry date.
4. Confirm the prod user set is assigned. Worth being explicit on this one: the alarms added in #496 do not catch `AADSTS50105 user-not-assigned`, because Entra rejects it before any callback, so a missing assignment looks like healthy infrastructure while nobody can actually log in.

Also worth settling with them whether prod gets its own app registration instead of sharing dev's. Sharing means one client secret spans both environments, and it stops `AUTH_ENTRA_AUDIENCE` telling them apart.

**So, not yet.** Everything on our side is done or in review. Once the redirect URI is in and the secrets are seeded, the only thing worth doing before merging is leaving dev quiet for a few hours so the older cookies from #497 age out on their own. After that this is good to go.

jsos3-cms (approved):
After resolving the nits and follow ups, we should be in a good position. Approving, just need to confirm once data load is complete

PR #496, opened 2026-07-30
fix(infra): route alarm notifications to the shared ISPG inbox

jsos3-cms:
This sets the login/OIDC alarm destination to the shared ISPG Zero Trust inbox in both dev and prod. Prod had no destination at all, so the SNS subscription was count-gated to zero subscribers and the login auth alarms had nowhere to go. Dev was pointed at a personal address that was only ever meant to stand up alerting during the Entra pilot.

An email subscription requires a one-time manual confirmation click after apply. Because dev's address is changing, Terraform will replace that subscription too, so both dev and prod will send a fresh confirmation email that someone monitoring the shared inbox needs to click.

## Test plan

- [ ] `terraform fmt -check` passes in `infrastructure/`
- [ ] After apply, `aws sns list-subscriptions-by-topic` for the dev topic shows the `ISPGZeroTrust@cms.hhs.gov` endpoint with a `SubscriptionArn` that is not `PendingConfirmation`
- [ ] After apply, `aws sns list-subscriptions-by-topic` for the prod topic shows the same endpoint with a confirmed `SubscriptionArn`

## Blockers

This is one of two prerequisites for merging #495 (enable Entra dual-IdP in prod), so that monitoring exists and is confirmed before the flag flips. Refs #495

### DIFF
--- a/infrastructure/tfvars/dev.tfvars
+++ b/infrastructure/tfvars/dev.tfvars
@@ -18,7 +18,7 @@ entra_enabled = true
 # the dev Entra pilot. Replace it with a shared team inbox or a Slack webhook
 # before this routing is relied on beyond dev / before prod, so paging never
 # depends on one person being reachable.
-alarm_notification_email = "jono@aquia.us"
+alarm_notification_email = "ISPGZeroTrust@cms.hhs.gov"
 
 # TLS cert rotation Lambda
 # ACM ARN sourced from SSM Parameter Store /ztmf/dev/cert-rotation/acm-arn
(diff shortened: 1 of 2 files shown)


jsos3-cms:
I have confirmed the subscription for dev and I am standing by for prod subscription information

MackOverflow (approved):
Small, correct, well-documented, low-risk, and it fixes a genuine prod gap (alarms with no destination). Only real caveat is operational, not code: confirm the subscription and treat that as the hard prerequisite for [#495](https://github.com/CMS-Enterprise/ztmf/pull/495). 

Should we consider a follow-up for a more robust paging channel than email? Slack alert (down the line?)

Issue #500, opened 2026-07-31
[P0] Data call rollover resolves one global predecessor; FY26 inherits the HHS import and drops CMS's own cycle (reproduced on prod data)


costacalvin:
> **P0 — blocks FY26 data call creation in prod.** `copyPreviousScores` runs once, at creation, and snapshots whatever exists at that moment. There is no re-run path. This has to be decided and in place *before* the FY2026 call is created in prod.

## Summary

Rollover picks **one predecessor cycle, globally, for every system**. That was correct while ZTMF ran a single annual cadence. It is no longer correct now that CMS carries two cycles in the same year — its own hand-entered ZTMF call *and* the HHS import.

This is now **reproduced end to end on prod data**, not theoretical.

## Reproduction

A current prod snapshot was restored into local docker, the API brought up off `main`, and a real **FY2026 Q3** data call created through `POST /api/v1/datacalls` as an `OWNER`. Actual code path, actual prod data.

| Data call | Scores | Role |
|---|---|---|
| FY2026 Q3 (new) | 38,160 | created |
| FY25 ZTM | 38,160 | **copied from this** |
| FY2025 Q3 | 9,589 | **ignored** |

The new cycle inherited the HHS import and ignored CMS's own hand-entered data call.

### Three measured consequences

1. **24 systems carried forward showing 0/40** — the identical 24 that are broken in FY25 ZTM. They score exactly **1.00**, which renders as a legitimate "lowest tier" result rather than an error.
2. **15 CMS systems that *did* answer FY2025 Q3 received zero rows in FY2026.** They open blank despite having answered.
3. **`status` correctly reset to `not_started` on all 38,160** — ztmf#435 working as intended. Do not regress this.

## The code

`DataCall.Save` calls `copyPreviousScores` on create — [datacalls.go:56-60](backend/internal/model/datacalls.go#L56-L60) (insert only, `d.DataCallID == 0`).
`copyPreviousScores` → `findPreviousDataCall` — [scores.go:750-825](backend/internal/model/scores.go#L750-L825), [datacalls.go:82-107](backend/internal/model/datacalls.go#L82-L107).

`findPreviousDataCall` picks ONE predecessor globally, for every system:

```sql
WHERE deadline < (SELECT deadline FROM datacalls WHERE datacallid = ?)
ORDER BY deadline DESC, datacallid DESC
LIMIT 1
```

[...]

PR #501, opened 2026-07-31
fix(rollover): key the FY2026 predecessor by OpDiv (ztmf#500)

costacalvin:
Closes the P0 in #500 for the FY2026 cycle. **Temporary by design — this block is deleted once the FY2026 data call exists in prod.**

## Problem

`copyPreviousScores` resolves **one** predecessor cycle globally, for every system. That was correct while ZTMF ran a single annual cadence. It is not correct now that two independent cycles cover FY2025:

| Cycle | Owner | Deadline |
|---|---|---|
| `FY2025 Q3` | CMS's own hand-entered call | 2025-05-07 |
| `FY25 ZTM` | HHS import | 2025-09-30 |

`findPreviousDataCall` orders by `deadline DESC` (the #448 fix — correct, and kept). So `FY25 ZTM` wins on its later September deadline and **every CMS system inherits HHS's imported view of itself** instead of the answers its ISSOs submitted. Reproduced on a prod snapshot in #500: the new cycle took all 38,160 rows from `FY25 ZTM` and ignored `FY2025 Q3` entirely.

## Approach

A system's OpDiv selects exactly **one** source cycle; the other is not read at all.

- CMS systems (`opdivs.code = 'CMS'`) → `FY2025 Q3`
- every other OpDiv → `FY25 ZTM`

Exclusion, not preference. The two cycles are separate lineages, not one record with gaps — CMS ran its own call, and the HHS ZTM cycles were loaded independently with their systems only arriving in ZTMF recently. Where they overlap (the import carries its own view of some CMS systems), the CMS entry is the record and the imported view is not authoritative. That overlap is exactly what made the global predecessor wrong. Confirmed with the data owner: there is no case where a CMS system should pull from the HHS cycle.

Routing is by `fismasystems.opdiv_id`, which is `NOT NULL` with an FK to `opdivs` (migrations 0027/0029), so every system resolves and none can be routed by accident.

## Gating

Two independent conditions, so the override cannot leak onto any other data call before it is removed:

1. **Name** — target starts with `FY2026` or `FY26`. Both year forms are accepted deliberately: FY2026 is a single unified cycle covering all of HHS including CMS, its final name is not fixed, and the prior conventions disagree (`FY2025 Q3` four-digit vs `FY25 ZTM` two-digit). Declining the real cycle over a naming guess is the costlier failure.
2. **Deadline** — target's deadline must be later than *both* sources'. This is the **#448 invariant restated**: a call named FY26 but deadlined earlier is a backfill, not the next cycle. The override replaces `findPreviousDataCall` entirely, so it carries #448's guarantee itself rather than inheriting it.
3. **First cycle only** — the target must be the earliest cycle matching those prefixes, ordered by `(deadline, datacallid)`. Without this a second same-year call (`FY2026 Q4`, a redo, a backfill) passes gates 1 and 2 and re-sources from the 2025 cycles, *discarding* everything the real FY2026 cycle accrued. This gate is what makes the broad prefix in gate 1 safe. It reads live state, not history — deleting a data call and recreating it re-arms the override, so dev stays re-testable.

Sources resolve by **name, not id** — `datacalls.datacall` is `UNIQUE` (migration 0013) and ids differ between dev and prod. Any decline logs a specific reason and falls through to the existing global path unchanged, so this is inert on any environment lacking both 2025 cycles.

## Safety

**The diff is purely additive — no existing line is removed.** `findPreviousDataCall` and the normal copy path are byte-identical.

- `status` still seeded as the `'not_started'` literal in the same `INSERT…SELECT` (#435 — the answer carries forward, its review state does not)
- `DISTINCT ON (fismasystemid, functionid)` guarantees one answer per question; `scores` has no uniqueness constraint, so a source cycle holding duplicates would otherwise copy both
- rollover stays best-effort: a copy error is logged, never fails the data call create

## Observability

One `ROLLOVER_HARDCODE` line per create. On the FY26 create:

```
ROLLOVER_HARDCODE datacall=<new> status=active mode=exclusion
  cms_source=<id>("FY2025 Q3") other_source=<id>("FY25 ZTM")
  cms_systems=… other_systems=…
  src_rows=… selected_rows=… excluded_rows=…
  copied=… deduped=… dropped_unresolvable=…
  from_cms_source=… from_other_source=…
  cms_stranded=… other_stranded=…
  misfiled_carried=… unscored_env=…
```

`from_cms_source` is the headline check — under the old behavior it would be **0**. `excluded_rows` is the fix working: HHS's view of CMS systems, discarded.

`status=inactive` with a reason means the override declined and the normal path ran (`source_lookup`, `target_name_not_fy2026`, `target_deadline_not_after_sources`, `target_is_a_source`).

Alarms reuse the existing `ROLLOVER_ANOMALY` token, so they page through the current CloudWatch metric with no new infra:
- `reason=hardcode_empty_copy` — rows selected, none written
- `reason=hardcode_stranded_systems` — a system answered somewhere but rolled forward empty
- `reason=hardcode_unresolvable_rows` — assigned rows the copy could not carry because `functionoptionid` resolves to nothing

Two diagnostics were corrected while building this and are worth noting for anyone reading the log:
- `misfiled_carried` is restricted to systems that **have** a scoring environment. Without that, every row for a system whose `datacenterenvironment` is unmapped or maps to a NULL `scoring_key` (the `DECOMMISSIONED` marker, migration 0045) reports as mis-filed — measured 36 reported / 36 false on the local fixture. Those are counted separately as `unscored_env`.
- Row accounting separates `excluded_rows` (dropped by the OpDiv filter) from `deduped` (collapsed duplicates). Conflating them would report deliberately excluded rows as deduplication and hide how much exclusion drops.

`dropped_unresolvable` is a genuine partial-copy signal. An earlier revision of this PR argued it was unnecessary because enforced FKs meant the copy's INNER JOINs could not drop a row. **That was wrong** — `scores_functionoptionid_fkey` reports `convalidated = t` only because it was created against an empty table, and bulk loads run under `session_replication_role = replica`, which bypasses FK triggers. The real data contains 33 rows across 11 systems in `FY2025 Q3` with `functionoptionid = -1`, which has no `functionoptions` row. The copy discards them silently, so the check reports 33 rather than never firing. See the review comment below.

## Testing

- `make test-unit` — passes
- `make test-integration` (isolated freshly-seeded DB) — passes
- `TestMatchesRolloverHardcodeTarget` — 12 cases covering both naming conventions, case-insensitivity, and rejection of FY2027 / backfills / the source cycles
- **End-to-end through `POST /api/v1/datacalls`** against a prod-like snapshot: 993 systems, 13 OpDivs, 262 CMS / 731 other. Both routing branches exercised.

Verified against the database after the run:

| Assertion | Result |
|---|---|
| CMS rows not sourced from `FY2025 Q3` | **0** |
| non-CMS rows not sourced from `FY25 ZTM` | **0** |
| rows with status ≠ `not_started` | **0** |
| duplicate answers per (system, function) | **0** |
| orphan rows carried forward | **0** |
| `FY2027 Q1` | declined, `reason=target_name_not_fy2026`, fell through to `findPreviousDataCall` |

Against unpatched `main` on the same data:

| | main | this PR |
|---|---|---|
| CMS systems fed the HHS imported view | 226 | **0** |
| Systems covered | 957 | 969 |
| Systems rolled forward empty | 21 | 9 |
| Systems properly scored (not exactly 1.00) | 196 | 272 |

The committed fixture cannot exercise this — `_test_data_empire.sql` has `FY25 ZTM` but no `FY2025 Q3`, so the override is inert there and no existing test path changes behavior.

**Note for anyone modifying the copy SQL:** the `::int` casts on the CASE branches are load-bearing and cannot be validated in psql. Literal ids and `PREPARE p(int,…)` both supply the type information pgx omits, so a psql harness will certify SQL that fails at runtime with `integer = text`. It needs a real POST.

## Before creating the prod cycle

1. Confirm the exact names — `SELECT datacallid, datacall, deadline FROM datacalls ORDER BY deadline DESC;`. A mismatch means a **silent no-op**: FY26 would roll from HHS exactly as today, with only a `status=inactive` log line to show for it.
2. Name the FY2026 call starting `FY2026` or `FY26`, deadline after 2025-09-30.
3. **Decide the seven CMS-tagged HHS arrivals** — see the review comment; identifiers are in the internal handoff doc. They carry forward today and roll forward empty under exclusion. Retagging them to their true OpDiv makes this a clean win.
4. Check the log: `from_cms_source > 0` and `other_stranded=0`. `cms_stranded` is expected to be **9** on current data, from three known causes — not a reason to abort, but confirm the breakdown matches.

The copy runs **once**, at creation, with no re-run path (#411). If the log is wrong, recovery is deleting the data call — `scores_datacallid_fkey` cascades — and recreating it. Verify before anyone begins answering.

## Removal

Delete the block between `TEMPORARY HARD-CODE - ztmf#500` and `END TEMPORARY HARD-CODE`, plus the 4-line call site in `copyPreviousScores`, and optionally `TestMatchesRolloverHardcodeTarget`. No schema change, no migration, nothing to unwind — the rows written are ordinary `scores` rows identical in shape to a normal rollover.

From FY2026 onward there is a single global predecessor again containing every system, so #500 goes dormant on its own. The durable per-OpDiv design (#500 decisions 2 and 3, and the missing cycle-ownership column on `datacalls`) is still open and only matters if genuinely divergent per-OpDiv cadences return.

### DIFF
--- a/backend/internal/model/scores.go
+++ b/backend/internal/model/scores.go
@@ -742,107 +742,32 @@ ORDER BY ps.datacallid, ps.fismasystemid, ps.pillarid
 // TEMPORARY HARD-CODE - ztmf#500. DELETE THIS ENTIRE BLOCK after the FY2026
 // data call has been created, along with the call site in copyPreviousScores.
 //
-// Rule: CMS systems roll forward from CMS's own hand-entered ZTMF cycle; every
-// other OpDiv rolls forward from the HHS import cycle. This replaces the single
-// global predecessor (findPreviousDataCall) for one cycle only.
+// Rule: CMS systems roll forward from CMS's own hand-entered cycle; every other
+// OpDiv rolls forward from the HHS import cycle. Replaces the single global
+// predecessor (findPreviousDataCall) for the FY2026 cycle only.
 //
-// OpDiv is an EXCLUSION, not a preference. A system's OpDiv selects exactly one
-// source cycle and the other is not read at all: a CMS system never inherits an
-// HHS-import answer, even for a question it has no CMS-cycle answer for. Danny
-// Bowne confirmed there is no case where a CMS system should pull from the HHS
-// cycle - any such gap was already resolved in the data cleanup - so a CMS
-// system missing from FY2025 Q3 is a data problem to fix at the source, not one
-// to paper over at rollover time.
+// EXCLUSION, not preference: a system's OpDiv selects exactly one source cycle
+// and the other is not read at all. The two cycles are separate lineages, and
+// where they overlap the CMS entry is the record. A system with no rows in its
+// assigned cycle therefore rolls forward EMPTY - see cms_stranded below.
 //
-// This is safe because the assigned cycle IS each side's source of truth: a CMS
-// system's answers live in the CMS cycle, so reading only that cycle reads
-// everything it has. Exclusion is not discarding a fallback, it is declining to
-// overlay a second, non-authoritative copy of the same systems.
-//
-// The two cycles are separate lineages, not one record with gaps. CMS ran its
-// own hand-entered call; the HHS ZTM cycles were loaded independently and their
-// systems only arrived in ZTMF recently. Where they overlap - the HHS import
-// carries its own view of some CMS systems - the CMS entry is the record and the
-// imported view is not. That overlap is precisely what made the global
-// predecessor wrong: it handed CMS systems the imported view because that cycle
-// happened to have the later deadline. So a CMS system appearing in FY25 ZTM is
-// expected and is not evidence that it needs to read from there.
-//
-// Consequence, intended: FY25 ZTM's rows for CMS systems do not carry into FY26
-// at all. They remain queryable as history under their own datacallid, but the
-// forward line for a CMS system is its CMS-cycle answer only.
-//
-// The mechanical risk that remains is a system with no rows at all in its
-// assigned cycle - it rolls forward EMPTY, and this copy runs once with no
-// re-run path (ztmf#411). Under the rule above that count should be zero on both
-// sides. It is measured rather than assumed ("stranded" in the diagnostics
-// below), because zero is the premise the design rests on and a one-shot,
-// unrecoverable copy is a bad place to find out otherwise. Confirming it against
-// the target database before creating the real cycle costs one query:
-//
-//	SELECT o.code, COUNT(*) FROM fismasystems fs
-//	  JOIN opdivs o ON o.opdiv_id = fs.opdiv_id
-//	 WHERE fs.fismasystemid IN (SELECT fismasystemid FROM scores
-//	                             WHERE datacallid IN (<cms>, <other>))
-//	   AND NOT EXISTS (
-//	         SELECT 1 FROM scores sc
-//	           JOIN functionoptions fo ON fo.functionoptionid = sc.functionoptionid
-//	          WHERE sc.fismasystemid = fs.fismasystemid
-//	            AND sc.datacallid = CASE WHEN o.code = 'CMS' THEN <cms> ELSE <other> END)
-//	 GROUP BY o.code;
-//
-// The JOIN to functionoptions is load-bearing: "has rows in the assigned cycle"
-// is not the same question as "has rows the copy can carry". A system whose only
-// answers are functionoptionid = -1 sentinels passes the first test and still
-// rolls forward empty. Omitting the join under-reports - it predicted 8 against
-// an actual 9 on the prod-like snapshot, and the one it missed was exactly that
-// case.
-//
-// A second accepted consequence: exclusion carries mis-filed rows
-// (ztmf-misc#256) forward with no escape hatch. A system whose assigned cycle
-// holds a functionoptionid pointing outside its own scoring environment keeps
-// that row and scores exactly 1.00; the other cycle's correct answer is not
-// consulted. misfiled_carried below counts these.
-//
-// fismasystems.opdiv_id is NOT NULL with an FK to opdivs (migrations 0029,
-// 0027), so the OpDiv always resolves and no system can be routed by accident.
-// Migration 0028 backfilled every pre-existing row to CMS, so the CMS branch is
-// the overwhelmingly common one.
-//
-// Sources are resolved by NAME, not by hard-coded IDs: datacalls.datacall is
-// UNIQUE (migration 0013) and the IDs differ between dev and prod, so names are
-// the only portable handle. If either name is missing the override does nothing
-// and the caller falls through to the existing global path unchanged - so this
-// is inert on any environment that lacks both 2025 cycles.
+// Sources resolve by NAME, not id: datacalls.datacall is UNIQUE (migration 0013)
+// and ids differ per environment. If either name is missing the override is inert
+// and the caller falls through to the global path unchanged.
 const (
 	rolloverHardcodeCMSSource   = "FY2025 Q3" // CMS's own hand-entered cycle
 	rolloverHardcodeOtherSource = "FY25 ZTM"  // HHS import
 	rolloverHardcodeCMSOpDiv    = "CMS"       // opdivs.code seeded by migration 0026
 )
 
-// The override is gated to the FY2026 cycle on TWO independent conditions, so it
-// cannot leak onto any other data call while this block is still deployed:
-//
-//  1. The target's name starts with one of rolloverHardcodeTargetPrefixes.
-//     Excludes FY2027 and anything else created before the removal ticket lands,
-//     which is what makes "delete this after FY26 kicks off" safe rather than a
-//     deadline. Both prefixes are accepted because the two source cycles
-//     themselves disagree on the convention ("FY2025 Q3" vs "FY25 ZTM") and the
-//     FY26 name is chosen by the operator at creation time.
-//  2. The target's deadline is later than BOTH sources' deadlines. This is the
-//     ztmf#448 invariant restated: a backfill call with an earlier deadline must
-//     not pull a 2025 cycle forward, no matter what it is named. The override
-//     replaces findPreviousDataCall entirely, so it has to carry #448's guarantee
-//     itself rather than inherit it.
-//
-// A declined target is logged with the specific reason and the actual name, so a
-// gate that rejects the real FY26 call is diagnosable from the log line alone
-// rather than looking like a normal rollover. That failure mode matters: the copy
-// runs once, at creation, with no re-run path (ztmf#411).
+// Gated to FY2026 on two independent conditions so it cannot leak onto another
+// data call before this block is deleted: the target's name matches one of these
+// prefixes, AND its deadline is later than both sources' (the ztmf#448 invariant,
+// which this path must carry itself since it bypasses findPreviousDataCall).
+// Both prefixes are accepted because the source cycles disagree on the convention
+// ("FY2025 Q3" vs "FY25 ZTM") and the operator types the FY26 name.
 var rolloverHardcodeTargetPrefixes = []string{"FY2026", "FY26"}
 
-// matchesRolloverHardcodeTarget reports whether name is the FY2026 cycle this
-// override is scoped to. Case-insensitive: the operator types the name.
 func matchesRolloverHardcodeTarget(name string) bool {
 	upper := strings.ToUpper(strings.TrimSpace(name))
 	for _, prefix := range rolloverHardcodeTargetPrefixes {
@@ -853,12 +778,10 @@ func matchesRolloverHardcodeTarget(name string) bool {
 	return false
 }
 
-// copyPreviousScoresOpDivHardcoded performs the OpDiv-keyed rollover described
-// above. handled is false when the override does not apply, in which case the
-// caller must fall back to the normal global path. It does not apply when either
-// named cycle is missing, when the new call is itself one of them, or when the
-// new call fails either target gate (name, deadline ordering). Every decline is
-// logged with a specific reason.
+// copyPreviousScoresOpDivHardcoded performs the OpDiv-keyed rollover. handled is
+// false when the override does not apply (missing source, target is a source, or
+// either gate rejects it), in which case the caller falls back to the global path.
+// Every decline logs a specific reason.
 func copyPreviousScoresOpDivHardcoded(ctx context.Context, dataCallID int32) (copied int64, handled bool, err error) {
 	conn, err := db.Conn(ctx)
 	if err != nil {
@@ -872,11 +795,9 @@ func copyPreviousScoresOpDivHardcoded(ctx context.Context, dataCallID int32) (co
 		targetName                 string
 		targetDeadline             time.Time
 	)
-	// The two id subqueries MUST stay ahead of the GREATEST: a missing source name
+	// The id subqueries MUST stay ahead of the GREATEST: a missing source name
 	// yields NULL, which fails the scan into int32 and takes the inactive path.
-	// GREATEST would not catch it on its own - unlike most Postgres functions it
-	// ignores NULL arguments, so GREATEST(NULL, deadline) returns the deadline and
-	// a half-present source pair would look valid.
+	// GREATEST alone would not catch it - it ignores NULL arguments.
 	err = conn.QueryRow(ctx, `
 		SELECT
 			(SELECT datacallid FROM datacalls WHERE datacall = $1),
@@ -889,8 +810,6 @@ func copyPreviousScoresOpDivHardcoded(ctx context.Context, dataCallID int32) (co
 		rolloverHardcodeCMSSource, rolloverHardcodeOtherSource, dataCallID,
 	).Scan(&cmsSourceID, &otherSourceID, &latestSourceDeadline, &targetName, &targetDeadline)
 	if err != nil {
-		// Either name is absent (NULL -> scan error) or the query failed. Not an
-		// anomaly: the override simply does not apply here.
 		log.Printf("ROLLOVER_HARDCODE datacall=%d status=inactive reason=source_lookup err=%v", dataCallID, err)
 		return 0, false, nil
 	}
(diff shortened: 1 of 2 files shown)


costacalvin:
## Review update — two defects found and fixed (b6961dd)

Ran the override end-to-end through `POST /api/v1/datacalls` against a prod-like snapshot (993 systems, 13 OpDivs) for the first time. Both routing branches were exercised, and two real defects surfaced. **The original commit 9cad00b copied zero rows.**

### 1. The copy failed outright — `SQLSTATE 42883, integer = text`

Both branches of `CASE WHEN o.code = $4 THEN $2 ELSE $3 END` are bind parameters. Postgres has nothing to infer from, resolves the CASE to `text`, and `integer = text` has no operator.

Reproduced directly:
```
PREPARE untyped AS SELECT ... WHERE s.datacallid = CASE WHEN 'CMS' = $1 THEN $2 ELSE $3 END;
ERROR:  operator does not exist: integer = text
```

Every harness used to validate the original commit substituted literal ids, and one declared explicit types via `PREPARE excl(int,int,int,text)` — both supply the type information pgx does not. The validation was the thing that hid the bug. Fixed with `::int` on both branches in the copy and the diagnostics.

### 2. The stated reason for having no partial-copy check was false

The comment claimed enforced FKs meant the copy's INNER JOINs could not drop a row. They can and do:

```
 datacallid | functionoptionid | orphan_rows | systems
          3 |               -1 |          33 |      11
```

`scores_functionoptionid_fkey` reports `convalidated = t` only because it was created against an empty table, and bulk loads run under `session_replication_role = replica`, bypassing the FK triggers. The snapshot contains rows the schema calls impossible.

Restored as `dropped_unresolvable`, counted with a LEFT JOIN, alarmed via `ROLLOVER_ANOMALY`. It reports **33** — a signal that was silent before.

Also corrected the pre-flight stranding query in the header: it tested "has rows in the assigned cycle" rather than "has rows the copy can carry", predicting 8 stranded against an actual 9. Joining `functionoptions` makes it agree at 9.

### End-to-end result

```
ROLLOVER_HARDCODE datacall=40 status=active mode=exclusion
  cms_source=3("FY2025 Q3") other_source=36("FY25 ZTM")
  cms_systems=262 other_systems=731
  src_rows=48110 selected_rows=38957 excluded_rows=9153
  copied=38836 deduped=121 dropped_unresolvable=33
  from_cms_source=9596 from_other_source=29240
  cms_stranded=9 other_stranded=0
  misfiled_carried=29364 unscored_env=0
```

Verified against the database afterward:

| Assertion | Result |
|---|---|
| CMS rows not sourced from FY2025 Q3 | **0** |
| non-CMS rows not sourced from FY25 ZTM | **0** |
| rows with status ≠ `not_started` | **0** |
| duplicate answers per (system, function) | **0** |
| orphan rows carried forward | **0** |
| `FY2027 Q1` gate | declined, `reason=target_name_not_fy2026`, fell through to `findPreviousDataCall` |

### Against unpatched `main`, same data

| | main | this PR |
|---|---|---|
| CMS systems fed the HHS imported view | 226 | **0** |
| Systems covered | 957 | 969 |
| Systems rolled forward empty | 21 | 9 |
| Systems properly scored (not exactly 1.00) | 196 | 272 |

`misfiled_carried=29364` is inherited, not caused — FY25 ZTM already scores ~80% of its systems at 1.00 because the staged HHS answers point at the CMS-Cloud-AWS function set regardless of environment. The override improves on its source (761 → 697 systems at 1.00) because CMS systems now draw from their well-matched cycle.

### Open item before the prod cycle

`cms_stranded=9`, from three distinct causes — all data quality, none of them code:

| Count | Cause | Impact |
|---|---|---|
| 1 | Duplicate `fismauid`; the sibling row holds the FY2025 Q3 answers | Covered either way |
| 1 | Its only FY2025 Q3 answers are `functionoptionid = -1` sentinels | Empty either way |
| **7** | **CMS-tagged HHS arrivals** — ZTM answers only, no CMS-cycle answers | **Carry forward today; roll forward empty under exclusion** |

Only the last group is a real trade: seven systems that carry forward under the current bug and roll forward empty under exclusion. Retagging them to their true OpDiv before creating the prod cycle makes this a clean win; leaving them means seven systems start FY26 blank. That is a data decision, not a code change.

Identifiers for all nine are in the internal handoff doc rather than here — this repo is public.

Related: `fismauid` has no unique constraint; a handful of UIDs span multiple system rows.

costacalvin:
Setting ready for review so we can test on dev

jsos3-cms (dismissed):
Approving. The P0 fix is correct, the diff is genuinely additive so nothing that works today can regress, and the verification behind it is more thorough than this repo usually sees. Two things below should be closed before the real FY2026 call is created in prod, but neither is a reason to hold the merge, and I explain why at the end.

Worth saying first: finding and fixing the untyped CASE yourself is the single most valuable thing in this PR. A defect that only appears at runtime, where every psql harness certifies the broken SQL because literals and `PREPARE` both supply the type pgx omits, is exactly the kind that ships. The note telling the next person not to remove the casts is the right artifact to leave behind. Same for retracting the FK claim rather than leaving a comment that sounded reassuring and was false.

## Verified

I checked these rather than taking the description's word for them.

Column parity with the normal path is exact: both INSERTs use `fismasystemid, datecalculated, notes, notes_is_ai_summary, functionoptionid, datacallid, status` in the same order, and both INNER JOIN `fismasystems` and `functionoptions`, so the rows `dropped_unresolvable` counts were already being dropped before this PR. It instruments existing behaviour rather than introducing it.

The `::int` fix is complete. There are three CASE sites, not two, and all three cast both branches: the copy at the `WHERE`, the `assigned` CTE, and the `dropped_unresolvable` subquery. Every other parameter in the new SQL sits in a position where Postgres infers from the column, which is the case the bug did not apply to.

`DISTINCT ON (s.fismasystemid, fo.functionid) ... ORDER BY ..., s.scoreid DESC` is deterministic on the primary key, and because the `WHERE` pins one source cycle per system before dedup evaluates, it cannot pick a winner from the excluded cycle.

The `status` literal is preserved, so ztmf#435 holds. The deadline gate uses a strict `After`, so an equal deadline declines. Sources resolve by name against a UNIQUE column. Everything is parameterized. `DataCall.Save` still logs and returns the created call, so rollover stays best-effort. And the new log lines emit only ids, counts, and the cycle name, which is appropriate for a public repo.

The comment explaining why the id subqueries must stay ahead of `GREATEST` is correct and load-bearing. `GREATEST` ignoring NULL really would swallow a missing source.

## Two things to close before the prod cycle

**A second FY2026-prefixed call re-triggers the override.** The name gate is a prefix match and the deadline gate only requires the target to be later than the two 2025 sources. So once the real FY2026 cycle exists and has answers in it, a later call named `FY2026 Q4`, or a same-year redo or backfill, passes both gates and sources from `FY2025 Q3` and `FY25 ZTM` again, discarding the lineage the real FY26 cycle accumulated. This is not a fallback missing an edge case: the override replaces `findPreviousDataCall`, so it actively overrides the correct answer. The test list reaches FY2027, which shows the removal-slips risk was considered, but a second same-year call is the nearer one given `FY2025 Q3` establishes quarter suffixes as the convention. `FY2024 Backfill` is rejected for the wrong year rather than for being a backfill, so it does not cover this. A third gate requiring the target be the earliest FY26-prefixed call would close it, or pin the exact name once pre-flight step 1 has confirmed it.

**`o.code = 'CMS'` is an exact-case comparison against an editable field, and its failure is silent.** `OpDiv.Save`'s update branch sets `code` unconditionally and `validate` only trims, so an OWNER can rename the CMS OpDiv's code through `PUT /opdivs/{opdiv_id}`. The important part is that migration 0026's uniqueness index is on `LOWER(code)`, which `validate`'s own comment references, so the schema already treats this column as case-insensitive and a rename from `CMS` to `cms` collides with nothing and saves cleanly. If that happens before the FY26 rollover runs, the comparison matches zero rows, every system falls to the `ELSE` branch, and the entire estate sources from `FY25 ZTM`. That is this PR's own P0, reproduced exactly.

What makes it worth fixing rather than noting is that none of the three anomaly triggers fire: `copied` is nonzero because everything copied from the other bucket, and nobody is stranded by the code's definition. The only signal is `cms_systems=0` in an info-level line that a human has to read. Given the copy runs once with no re-run path, a case-insensitive comparison would match the schema's own model, and promoting `cms_systems == 0` to `ROLLOVER_ANOMALY` would make the failure page rather than sit in a log.

## Follow-ups, not for this PR

`DISTINCT ON` is the only dedup guard anywhere in the copy path, and it lives inside the block marked for deletion. The permanent path has no dedup, `scores` has no uniqueness constraint on system and function, and ztmf#500 already calls this out as an implementation hazard under decision 4. So this PR correctly papers over a pre-existing gap for exactly the window it is alive, then hands the codebase back unchanged. That is the right scope for a P0 fix, but the guard disappearing on removal is worth capturing in the removal ticket so it is a decision rather than an accident.

Automated coverage is the name gate only, 36 of 283 lines. The copy SQL, the exclusion routing, both deadline branches, and all eleven diagnostic counters have none, and the fixture cannot reach them because it has `FY25 ZTM` but no `FY2025 Q3`, so the override declines at `source_lookup` on every test run. You disclose this, and the manual prod-snapshot run is real validation. The reason I would still push on it is the untyped CASE: that defect proved the copy SQL can be wrong in a way only a real execution catches, and right now nothing repeatable executes it. Adding a `FY2025 Q3` row to the fixture would make the override live in integration tests and turn a one-off into a gate.

One wording nit for whoever does the removal. Deleting `TestMatchesRolloverHardcodeTarget` is described as optional. It is not: the test calls `matchesRolloverHardcodeTarget` directly, so removing the production symbols without it breaks the build.

## Why this is an approval

Merging is not the consequential act here. `copyPreviousScores` only runs when a data call is created, so deploying this code changes nothing on its own, and both items above bite at cycle-creation time, which is already gated behind the human checklist in the description. The diff removes no existing line, so the override either fires or falls through to byte-identical behaviour.

So the two gate items belong on that checklist rather than in front of the merge. I would want them either fixed or written down as accepted residual risk before anyone creates the FY2026 call, not discovered afterwards, because there is no re-run path and recovery means deleting the data call.

Gates confirmed on my side: `go build ./...` and `go vet ./...` clean, `make test-unit` green across 19 packages, and the 12 new subtests pass.

costacalvin:
Both review items fixed, plus a third the testing surfaced — `5cda677`.

### 1. Second FY2026-prefixed call re-triggered the override

Correct, and it was the nearer hazard. Added a third gate: the target must be the **first** cycle matching the prefixes, ordered by `(deadline, datacallid)` as `findPreviousDataCall` orders. Candidates are filtered in Go with the same matcher the name gate uses, so the two cannot disagree.

Declines with `reason=not_first_fy2026_cycle` and logs the earlier name. The gate reads live state, not history — deleting a data call and recreating it re-arms the override, verified three times, which keeps dev re-testable given there is no `DELETE /datacalls` endpoint.

This also let the prefix list stay broad. FY2026 is a single unified cycle covering all of HHS including CMS — merging the two 2025 lineages is the point of the change — so there is no second cycle to exclude, the final name is not fixed, and the prior conventions disagree (`FY2025 Q3` four-digit vs `FY25 ZTM` two-digit). Declining the real cycle over a naming guess is the costlier failure. Gate 3 is what makes the broad prefix safe.

### 2. `o.code = 'CMS'` exact-case against an editable column

Verified and fixed. `OpDiv.Save` updates `code` unconditionally, `validate` only trims, and migration 0026's uniqueness index is on `LOWER(code)` — so the rename saves cleanly and the comparison then matches zero rows.

Now `UPPER()` at all seven comparison sites, matching the schema's own case-insensitive model. Your point about it being unalarmed was the important half: `copied` stays nonzero because the other bucket copies fine, and the CMS-cycle-only systems strand as `other_stranded`, indistinguishable from the expected baseline. `cms_systems == 0` now raises `ROLLOVER_ANOMALY reason=hardcode_no_cms_systems`.

### 3. Wording nit

You're right that deleting `TestMatchesRolloverHardcodeTarget` is not optional — it calls `matchesRolloverHardcodeTarget` directly and the build breaks without it. Corrected. Also fixed the FY2025 Q3 deadline in the description above: `2025-05-07`, not `2025-06-30`.

---

### Summary of Danny's Validation against a local restore of current prod 

Taken after the latest full load and re-point: 1,529 systems, 153,547 scores, dbversion 53.

```
status=active mode=exclusion  cms_systems=276 other_systems=1253
src_rows=47869 selected_rows=38796 excluded_rows=9073
copied=38676 deduped=120 dropped_unresolvable=33
from_cms_source=9476 from_other_source=29200
cms_stranded=1 other_stranded=0  misfiled_carried=40 unscored_env=0
```

- **`from_cms_source` 0 → 9,476.** The same create on pre-patch code took all rows from `FY25 ZTM` and ignored `FY2025 Q3` entirely. `excluded_rows=9073` is the imported view of CMS systems being discarded.
- **`cms_stranded=1`** — one CMS system present only in the HHS import, so it opens FY2026 blank. Exclusion behaving as specified; a data decision, not a code one.
- **`dropped_unresolvable=33`** — pre-existing rows across 11 systems whose `functionoptionid` has no `functionoptions` row. The copy correctly refuses to carry them. Separate ticket.
- **`misfiled_carried=40`** — worth noting these do score: the aggregate has no implicit decommissioned filter, and `unscored_env=0` confirms all 40 sit on systems with a live scoring key. Small, but real rather than noise, and it belongs with the 33.

jsos3-cms (approved):
Re-approving. Both items are addressed, and I verified `5cda677` against the code rather than the commit message.

**Gate 3 is correct.** The precedence predicate `(deadline < $2 OR (deadline = $2 AND datacallid < $1))` is a consistent total order on `(deadline, datacallid)`, matching how `findPreviousDataCall` orders, and adding the equal-deadline tiebreak makes it deterministic where the original relied on a strict inequality alone. Filtering candidates in Go with the same matcher the name gate uses is the right call: it makes the two structurally unable to disagree, which a second SQL predicate would not have. A scan error declines and falls through rather than proceeding, so the gate fails safe. And reading live state rather than history is the property that keeps dev re-testable, as you say.

Your argument for keeping the prefix list broad now holds, and it did not before. Gate 3 is what changes that: with only gates 1 and 2, a broad prefix meant any FY26-named call could be hijacked, so narrowing the name was the only protection available. With gate 3, the first matching cycle is the only one ever treated as the target, so breadth costs nothing and declining the real cycle over a naming guess stops being a risk. That is the better trade.

**The OpDiv fix is more thorough than the summary suggests.** `UPPER()` is at all seven sites, and line 969 is the one worth calling out: normalizing the projected `opdiv_code` inside the `missing` CTE, not just the direct comparisons, is what keeps the `cms_stranded` and `other_stranded` split correct downstream. Missing that would have left the diagnostics disagreeing with the routing while the copy itself looked right. Constant stays uppercase with a comment saying why, which is the guard that matters for whoever edits this next.

`cms_systems == 0` now raising `ROLLOVER_ANOMALY reason=hardcode_no_cms_systems` closes the half I cared most about. An unalarmed silent reproduction of the original P0 was the worst shape available here.

## One new pre-flight item, created by gate 3

Gate 3 cuts both ways, and the flip side belongs on the checklist before the prod create.

It protects a *second* FY26 call from being wrongly overridden. It also means a *pre-existing* FY26-prefixed call makes the real cycle decline. So if a test FY2026 call is ever created on prod and left in place, creating the real cycle afterwards falls through to `findPreviousDataCall` and sources from the HHS cycle, which is the P0 again. The decline is visible as `reason=not_first_fy2026_cycle` with the earlier name, so it is detectable, but only by reading the log.

Worth adding to the pre-flight: confirm no FY26-prefixed data call already exists before creating the real one. The existing step 1 query (`SELECT datacallid, datacall, deadline FROM datacalls ORDER BY deadline DESC`) already surfaces it, so this is a line in the checklist rather than more code.

## Coverage, unchanged in kind but larger in area

The test changes in this commit are comment-only, so the roughly 99 new lines including `earlierRolloverHardcodeTargets` and the `UPPER()` change carry no automated coverage. That is the same gap I raised before rather than a new one, and `earlierRolloverHardcodeTargets` needs a database, so a pure unit test is genuinely awkward. Not blocking, and your manual verification against the prod restore is real. Restating it only because the untested surface grew and this code runs exactly once.

The numbers from Danny's validation read correctly to me. `from_cms_source` at 9,476 against 0 on pre-patch code is the fix working, `excluded_rows=9073` is the discarded imported view, and `cms_stranded=1` is exclusion behaving as specified. Agreed that the 33 unresolvable rows and the 40 mis-filed carries belong together in their own ticket rather than here.

Gates on my side at `5cda677`: `go build ./...` and `go vet ./...` clean, 297 model tests pass, CI green on all 11 checks, and GitHub verifies the signature.

costacalvin:
Removal tracked in #502 — includes the two build-breaking gotchas (the `pgxpool` import and `TestMatchesRolloverHardcodeTarget` are both load-bearing) and the open decision on whether the `DISTINCT ON` dedup guard should survive the block or be accepted as reverting to the status quo.

PR #511, opened 2026-08-03
fix(fismasystems): let OpDiv-scoped admins write the ISSO name

### LINKED ISSUE #509: Allow OpDiv-scoped admins to write isso_name (gated to the HHS tier and silently discarded)
### Description

`isso_name` is grouped with the HHS onboarding fields in the system write path, so only the HHS tier can set it. That makes the field uncorrectable for OpDiv-scoped admins on the systems they administer.

`PUT /fismasystems/{id}` gates the 12 HHS onboarding fields on `HasUnscopedRead()`, which covers `OWNER`, `HHS_ADMIN`, and `HHS_READONLY_ADMIN` only. For any other caller the handler calls `copyHHSMetadata(existing, f)`, which copies the stored values over the request so a scoped admin cannot wipe data they cannot see. `ISSOName` is one of the fields copied, so an `OPDIV_ADMIN` write to `isso_name` is replaced with the stored value and the response reports success.

The model layer already supports the write. `isso_name` is present in both the INSERT column list and the UPDATE column map, and `blankToNil` maps an empty string to NULL. Only the controller gate stands in the way.

The grouping made sense when `isso_name` arrived with the rest of the onboarding import, but it does not hold up now: the other 11 fields describe the system, while this one is a contact name that whoever administers the OpDiv is the right person to keep current.

### Steps to Reproduce

1. Authenticate as an `OPDIV_ADMIN` holding a grant for the system's OpDiv
2. `PUT /api/v1/fismasystems/{id}` with a changed `isso_name` and the system's other required fields
3. Observe a 200 response
4. `GET /api/v1/fismasystems/{id}` and observe `isso_name` unchanged

### Expected Behavior

An `OPDIV_ADMIN` can write `isso_name` on systems in an OpDiv they hold a grant for. `OWNER` and `HHS_ADMIN` can write it on any system. Every other tier cannot, and the remaining 11 HHS onboarding fields stay restricted to the HHS tier.

An empty string continues to write NULL, so clearing the field restores the email-derived name that the list read falls back to.

### Actual Behavior

The write is accepted, silently discarded, and reported as successful.

### Environment

all

### Acceptance Criteria

- [ ] `isso_name` is carved out of the blanket `copyHHSMetadata` copy so a scoped admin's value survives
- [ ] The carve-out is authorized by the existing `CanManageFismaSystem(sys.OpDivID)` check, so an `OPDIV_ADMIN` is confined to their granted OpDivs
- [ ] The other 11 HHS onboarding fields are unchanged in behavior for every tier
- [ ] `ISSO`, `ISSM`, `SYSTEM_DELEGATE`, and the read-only admin tiers still cannot write the field, and are still rejected before reaching the carve-out
- [ ] An empty string still writes NULL rather than an empty string
- [ ] An omitted `isso_name` still leaves the stored value untouched, preserving partial-PUT and importer protection
- [ ] Tests cover: `OPDIV_ADMIN` in-scope write persists; `OPDIV_ADMIN` out-of-scope write is forbidden; a non-admin tier is rejected; the other HHS fields are still copied over for a scoped actor; clear-to-NULL works
- [ ] `openapi.yaml` reflects the field being writable if its current description says otherwise

### Severity

Medium - Impacts usage but has a workaround

### Affected Component

Backend API

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

A direct data-side correction per system, which does not scale and leaves the field uneditable for everyone else.

jsos3-cms:
Closes #509

## What was wrong

`isso_name` arrived alongside the HHS onboarding columns and was grouped with them in the system write path, so only the unscoped-write tiers could set it. An `OPDIV_ADMIN`'s value was replaced with the stored one and the request still returned 204, so the edit looked like it worked.

That mattered more than a normal permission gap, because `isso_name` is a stored column and the systems list read resolves the display name as:

```sql
COALESCE(fismasystems.isso_name,
         (SELECT fullname FROM users WHERE LOWER(email) = LOWER(fismasystems.issoemail) LIMIT 1))
```

The stored column wins. CMS systems have it NULL, so their name derives from `issoemail` and self-heals whenever the email is corrected. Systems whose column the onboarding import populated keep the imported name forever: updating the ISSO email changes the email and nothing else. With the field also locked in the UI, those names were uncorrectable by anyone.

## The change

Two deleted assignments. `isso_name` no longer participates in the field set that gets cleared on create and restored on update for scoped actors.

No new authorization, because the authorization already ran. `guardManageFismaSystem` evaluates `CanManageFismaSystem` against the **stored** system's `opdiv_id` before the restore, so an `OPDIV_ADMIN` reaching that point may already manage that specific system. On create, an OpDiv-tier actor must already pass `IsAssignedOpDiv` for the `opdiv_id` they send. Both paths are unchanged by this PR and both already scope the actor correctly. No model changes were needed either: `isso_name` was already in the INSERT column list and the UPDATE column map.

Resulting matrix:

| Actor | `isso_name` |
|---|---|
| `OWNER`, `HHS_ADMIN` | writable on any system |
| `OPDIV_ADMIN` | writable on systems in an OpDiv they hold a grant for |
| `OPDIV_ADMIN`, out of scope | 403, nothing written |
| `HHS_READONLY_ADMIN`, `OPDIV_READONLY_ADMIN`, `ISSO`, `ISSM`, `SYSTEM_DELEGATE` | rejected before reaching the field |

The other 11 fields keep their existing behavior for every tier.

Three request states are preserved: an omitted key leaves the stored value alone, `""` writes NULL, a value is written. The NULL case is load-bearing rather than an edge case, because it is what restores the email-derived name. Clearing the field is the correct fix for a system that should track its ISSO's user record, and the frontend change will say so at the point of edit.

Both helpers are renamed after the permission rule rather than the workbook the fields came from, since the old `*HHSMetadata` names are what made a contact name read as protected system metadata. One comment claiming the copy shielded "data they cannot see" is corrected: scoped admins can read every one of those columns, the list and GET reads return them all and only filter rows by OpDiv. The copy is partial-PUT protection, not confidentiality.

## Reviewer note: the E2E suite was asserting the bug

`emberfall_tests.yml` Test B needed two fixes, and they are the reason the diff touches that file:

- Part 1 expected `isso_name: null` in the response after an OpDiv admin create. That was the silent discard written down as expected behavior.
- Part 3 sent the same `isso_name` an unscoped admin had already stored, so the trailing GET passed whether the write landed or was restored. It now sends a distinct name with an explicit override on the merge key, so the assertion can actually distinguish the two.

## Test plan

- `go test -short ./...` (no database, what CI gates on): 597 pass
- integration suite against a freshly seeded database: 87 pass
- Emberfall end to end: 187 ran, 0 failed
- `make generate-openapi` then `make lint-openapi`: spec regenerated from the Go doc comment, Redocly exit 0, no errors. The spec is generated, so it is not hand-edited.
- Direct database check after the E2E run, which is the evidence that matters most:

| System | Path | `isso_name` | `system_owner` | `hva` |
|---|---|---|---|---|
| 1002 | UPDATE by OpDiv admin | their value landed | unscoped admin's stored value, not the one the OpDiv admin sent | unscoped admin's stored value, not the one sent |
| created | INSERT by OpDiv admin | persisted | null | null |

Note for anyone running these locally: `make test-e2e` and `make test-integration` both rely on a foreground `sleep`, which is blocked in some sandboxed shells. When that happens `make` exits 0 **without running the tests**. Drive the containers and test binaries directly with a readiness poll if a green exit looks too fast to be real. They are fine in CI.

## Not in scope

Two pre-existing behaviors that this PR does not introduce and should not be blamed on it:

- `Save` rejects `cloud_system == false` alongside a non-empty `cloud_service_model`. A row carrying that combination from an import cannot be edited at all by a scoped admin, because the restore repopulates both and the save 400s. Latent today, untouched here.
- `PUT /fismasystems/0` matches the `[0-9]+` route regex and the handler treats ID 0 as a create, so that URL is an alias for POST. Authorization is unaffected, since the create path still requires the OpDiv grant, but explicitly rejecting 0 is worth a small cleanup ticket.

Once anyone writes `isso_name` on a CMS system, the derived value is shadowed until the field is cleared. That is intended for OpDivs with no load, and the frontend change covers communicating it.

The frontend half is CMS-Enterprise/ztmf-ui#661 and must not ship until this deploys, or the field reports success and discards the change.

### DIFF
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -424,6 +424,35 @@ func TestSaveFismaSystem_ReadonlyAdminForbidden(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
 
+// isso_name is writable by OpDiv-scoped admins, so it is the one field in the
+// extended set whose gate could plausibly be loosened further. These two tiers
+// are the ones a "let the ISSO fix their own name" or "read-only should be able
+// to correct a typo" change would reach for, and both must stay behind
+// SaveFismaSystem's admin gate. The gate returns before any database call.
+func TestSaveFismaSystem_ISSONameDeniedTiersForbidden(t *testing.T) {
+	for name, u := range map[string]*model.User{
+		"ISSO":                 issoUser,
+		"OPDIV_READONLY_ADMIN": opdivReadonly,
+	} {
+		t.Run(name, func(t *testing.T) {
+			body := jsonBody(t, map[string]any{
+				"fismauid":     "12345678-1234-4abc-8def-123456789abc",
+				"fismaacronym": "TEST",
+				"fismaname":    "Test System",
+				"isso_name":    "Should Never Persist",
+			})
+			r := httptest.NewRequest("PUT", "/api/v1/fismasystems/1", body)
+			r.Header.Set("Content-Type", "application/json")
+			r = mux.SetURLVars(r, map[string]string{"fismasystemid": "1"})
+			r = withUser(r, u)
+			w := httptest.NewRecorder()
+
+			SaveFismaSystem(w, r)
+			assert.Equal(t, http.StatusForbidden, w.Code, "%s must not write isso_name", name)
+		})
+	}
+}
+
 // --- DeleteFismaSystem ---
 
 func TestDeleteFismaSystem_ReadonlyAdminForbidden(t *testing.T) {
--- a/backend/cmd/api/internal/controller/fismasystems.go
+++ b/backend/cmd/api/internal/controller/fismasystems.go
@@ -104,9 +104,10 @@ func GetFismaSystem(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, fismasystem, nil)
 }
 
-// clearHHSMetadata nils the 12 HHS onboarding fields on a FismaSystem.
-// Called on INSERT when the acting user lacks unscoped read access.
-func clearHHSMetadata(fs *model.FismaSystem) {
+// clearUnscopedOnlyFields nils the 11 system-attribute fields only an
+// unscoped-write admin may set. Called on INSERT when the acting user lacks
+// unscoped read access.
+func clearUnscopedOnlyFields(fs *model.FismaSystem) {
 	fs.HVA = nil
 	fs.FIPS = nil
 	fs.SystemType = nil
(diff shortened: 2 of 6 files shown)


danielbowne (approved):
Verified the authorization claim against the code rather than taking the body's word for it - guardManageFismaSystem is the same call that fetches the stored row, so every scoped-admin update passes CanManageFismaSystem against the stored OpDiv before the field passes through. Clean.

The omitted-vs-empty-vs-value handling and its test coverage are exactly right, and clearing to restore the email-derived name is the correct long-term state for systems with imported names.

Two non-blocking notes:

1. Confirm the ztmf-ui companion is queued - the field is still read-only in the edit modal, so this alone doesn't unblock OpDiv admins.
2. Verified rather than assumed: system_owner / system_owner_email have the same frozen-import shape and a worse one - no COALESCE fallback exists for the owner name, so imported values can never self-heal, and roughly 1,180 of 1,200 live non-CMS systems carry imported owner values today. Filed as #512 with the data, using this PR's pattern - deliberately not widening this PR.

jsos3-cms:
Thanks @danielbowne, and thanks for re-deriving the authorization claim from the code instead of taking the body's word for it. That was the part I most wanted a second pair of eyes on, since the whole justification for this being a two-line change rests on it.

**On the frontend companion, it is queued and written, and holding on a defect its review just found.** CMS-Enterprise/ztmf-ui#661, P1, in progress, committed on `fix/isso-name-editable-by-opdiv-admin` with 25 tests covering the per-role gate and the write signals. It is deliberately unpushed for the ordering reason you name: shipping it before this deploys would give an OpDiv admin an editable field, a success toast, and a discarded write, the same defect already tracked as ztmf-ui#651.

A review pass on that branch then found a second problem, worth mentioning here because it is about this field's semantics rather than the frontend in general. The detail page echoes its local draft into state after a save instead of refetching, which is harmless for every other extended field because the value sent is the value stored. `isso_name` is the first field where the post-save truth is computed server-side, so clearing it to restore the derived name currently displays an em dash until something refetches. That one bites `OWNER` and `HHS_ADMIN` today, independent of this PR, since they were never behind the gate. Being fixed on that branch before it goes up.

**One precision that affects #512's UI scope.** The read-only field you saw in the edit modal is real, but worth knowing that `EditSystemModal` has exactly one production render, in `Title.tsx`, with `mode` hardcoded to `create`. There is no production edit-mode caller, so edits to an existing system go through the System Details page edit view instead. Both surfaces read the same `fieldConfig` entry, so a single flag change covers them, but the surface to actually verify for a UI companion is the detail page. The modal only governs what can be set at system creation.

**On #512**, the shape is right and the no-COALESCE point makes it strictly worse than this one. `isso_name` at least had a derivation to fall back to, which is why clearing the field is a real remedy here; for the owner fields the stored column is the only source there is, so an imported value is permanently wrong with no self-healing path. The volume you measured makes that worth doing properly rather than folding into this PR, and I agree with keeping it separate. I am not picking it up right now.

One thing I would suggest for #512 when someone does take it: the two owner fields are currently in the same preserved set as the nine genuine system attributes, so moving them out means the set stops being "the fields a scoped admin may not write" and becomes something narrower. It is worth deciding at that point whether the remaining nine still share a coherent rule, or whether the grouping should be rethought rather than shrunk twice.

danielbowne:

Issue #512, opened 2026-08-03
Let OpDiv-scoped admins maintain system owner contact fields


danielbowne:
## Problem

#511 made `isso_name` writable by OpDiv-scoped admins. `system_owner` and `system_owner_email` have the same frozen-import shape but a worse fallback story: unlike `isso_name` there is no COALESCE derivation from a user record, so an imported owner name can never self-heal - the stored column is the only source there is.

No onboarding load refreshes these fields for non-CMS OpDivs, and scoped admins cannot write them. So when a system's owner changes, there is currently no path to correctness at all: not via the UI, not via a load, not via derivation.

On current prod-shaped data, roughly 1,180 of 1,200 live non-CMS systems carry imported owner values, and about 290 owner emails match no user record - drift that is already visible and can only grow.

## Proposal

Move both fields from the unscoped-only set into the OpDiv-writable set, following the #511 pattern exactly: same `guardManageFismaSystem` authorization, same three request states (omitted leaves stored, empty clears, value writes), same helper/unit/E2E test shape. They are contact fields, not system attributes, which is the distinction #511's helper rename established.

The UI companion should unlock the fields in the edit modal for scoped admins in the same pass.

jsos3-cms:
Agreed on the shape, and the no-derivation point makes this strictly worse than `isso_name`. There the stored column at least shadows something, which is why clearing the field is a real remedy; here the stored column is the only source, so an imported value is permanently wrong with no self-healing path. Not picking this up right now, but recording what the `isso_name` work turned up so it is not rediscovered.

## Where this overlaps #510

#510 is the read-side half of the same field family: the systems list read applies the `isso_name` COALESCE while the single-system read returns the raw column, so the two endpoints disagree for a system relying on the fallback. That issue is currently scoped to `isso_name`, but the decision it makes constrains this one.

**The premise that no derivation exists for the owner fields is true today and is not forced.** `system_owner_email` is the structural analogue of `issoemail`: both are addresses that can be matched against `users.email` to resolve a display name. So a COALESCE for `system_owner` off `system_owner_email` is possible in exactly the way the `isso_name` one is. Whether it is *desirable* is a real question, since a system owner is often not a ZTMF user and the lookup would miss more often than it hits, but "no fallback exists" should be read as a current-state fact rather than a constraint.

That gives two branches, and they interact with #510 differently:

[...]

PR #514, opened 2026-08-03
feat(fismasystems): OpDiv-writable owner contact fields + one shared isso_name derivation

### LINKED ISSUE #510: Systems list and single-system GET disagree on isso_name
### Description

The systems list and the single-system GET resolve `isso_name` differently, so the same system can show a name in one place and a blank in another.

`FindFismaSystems` rewrites the `isso_name` column into a COALESCE that falls back to the ISSO's user record:

```sql
COALESCE(fismasystems.isso_name,
         (SELECT fullname FROM users WHERE LOWER(email) = LOWER(fismasystems.issoemail) LIMIT 1))
```

`FindFismaSystem` selects the raw column list with no such rewrite, and the existing comment on the list path states the narrowing deliberately: "Applied to the list only; the single-system GET returns the stored value unchanged."

The frontend normally never notices, because the System Details page resolves its system out of the shared context array populated from the list endpoint rather than from the single GET. The single GET is reached only as a fallback, when the requested system is absent from that array. In practice that means a **decommissioned** system, since the list fetch is active-only, or a direct page load before the list has populated.

So the visible symptom is narrow: open a decommissioned system whose `isso_name` is NULL and whose `issoemail` matches a user record, and the ISSO Name renders blank, while the same system shows a resolved name anywhere it is served from a list.

### Steps to Reproduce

1. Find a decommissioned system whose `isso_name` is NULL and whose `issoemail` matches a user record
2. Navigate directly to that system's detail page, so it is fetched individually rather than read from the active-systems list
3. Observe ISSO Name empty
4. Compare against a list view that includes the system, which shows the resolved name

### Expected Behavior

One system shows one ISSO name, whichever endpoint served it.

### Actual Behavior

The list resolves the fallback and the single-system GET does not, so a system fetched individually shows a blank ISSO Name.

### Acceptance Criteria

- [ ] Decide deliberately whether the single-system GET should resolve the fallback, or whether the derivation should move to one place both reads share
- [ ] Whichever way it goes, the list and the single-system GET agree for the same system
- [ ] The edit form does not pre-fill a derived name into the stored-value field in a way that lets an admin persist an override without intending to. This is the reason the inconsistency was left alone when the field was unlocked, and it is the constraint any fix has to respect
- [ ] Tests cover a system with a stored `isso_name` and a system relying on the fallback, through both endpoints

### Environment

all

### Severity

Low - Minor issue or cosmetic

### Affected Component

Backend API

### Screenshots

_No response_

### Logs / Error Messages

```shell

```

### Workaround

Read the name from a list view that includes the system.

danielbowne:
Closes #512, closes #510. Stacked on #511 - the base is `fix/isso-name-opdiv-writable` and GitHub will retarget to `main` when it merges; only the top commit is new here.

## #512 - owner contact fields become OpDiv-writable

`system_owner` and `system_owner_email` leave the unscoped-only set and follow the #511 pattern exactly: same `guardManageFismaSystem` authorization, same three request states (omitted leaves the stored value, `""` clears, a value writes), same helper/unit/E2E test shape. The protected set shrinks to 9.

On the grouping question raised in the #512 thread: the remaining 9 do share one rule. They are facts about the system itself - impact level, hosting, operator model, lifecycle - while the three fields now outside the set are pointers at people, which is precisely what OpDiv admins exist to maintain. The set is not being whittled field by field; it converged on the rule the #511 rename named.

One divergence from `isso_name` that the UI companion must not paper over: there is no derivation behind the owner fields, so clearing them just clears them. The "clearing restores the name from the user account" help text pattern cannot be copied across (per the #512 thread).

## #510 - one shared isso_name derivation

The list read and the single-system GET disagreed on `isso_name` for systems relying on the email fallback. The derivation now lives in one helper (`resolveISSONameColumn`) used by both reads:

- the list resolves always, as before
- the single-system GET resolves via a new opt-in `ResolveISSOName` input, set only by the display read
- every internal fetch that feeds a write path (`guardManageFismaSystem`, `guardViewFismaSystem`, `FindFismaSystemByUUID`) keeps the raw column, so a derived name can never enter an entity that flows toward `Save` - the acceptance constraint from #510

A new integration test covers a fallback system and a stored-override system through both reads, and pins the raw read for write-path safety. If a derivation is ever wanted for the owner fields, there is now exactly one place to put it.

Known and accepted, not fixed here: a third-party client that round-trips a full GET body into a PUT could persist a resolved name as a stored override. The first-party UI cannot - the field is read-only in the field config and both PUT paths send only the dirty diff.

Being precise about the delta, because an earlier draft of this description undersold it. This does not merely inherit an exposure the list endpoint already had; it **widens that exposure to the single-system fallback path**. The frontend falls back to `GET /fismasystems/:id` when a system is absent from the list array - in practice a decommissioned system, or a direct URL load before the list populates. That path previously returned the raw stored column, so a resolved name could not reach the edit form there at all. Both consequences follow from the same change:

- the em-dash-after-clearing case on that path disappears, closing a residual recorded on #510 with no frontend change
- the accidental-override exposure now extends to that path, where it did not exist before

A separate `isso_name_resolved` response field is the clean fix if it is ever wanted, and that widening is what motivates it.

## Testing

- unit: helper set tests extended to the contact-field trio, including omitted-key semantics
- integration: new `TestFismaSystemISSONameResolutionIntegration` (stored vs fallback, both reads, raw-read pin)
- E2E: emberfall Test B extended - scoped-admin create and update now take contact-field values while the 9 attributes stay protected; full suite 187/187
- `make generate-openapi` run; spec committed

## Coordination

The UI companion for the owner fields should ride with ztmf-ui#661 (same field-config flag covers create modal and the System Details edit view, which is the surface that matters), with its own help text per the divergence above.

### DIFF
--- a/backend/cmd/api/internal/controller/fismasystems.go
+++ b/backend/cmd/api/internal/controller/fismasystems.go
@@ -76,7 +76,10 @@ func ListFismaSystems(w http.ResponseWriter, r *http.Request) {
 func GetFismaSystem(w http.ResponseWriter, r *http.Request) {
 	user := model.UserFromContext(r.Context())
 	vars := mux.Vars(r)
-	input := model.FindFismaSystemsInput{}
+	// A display read: resolve the ISSO display name the same way the list
+	// does, so the two endpoints cannot disagree on the same system
+	// (ztmf#510). Internal fetches that feed write paths keep the raw column.
+	input := model.FindFismaSystemsInput{ResolveISSOName: true}
 
 	if v, ok := vars["fismasystemid"]; ok {
 		var fismasystemID int32
@@ -104,7 +107,7 @@ func GetFismaSystem(w http.ResponseWriter, r *http.Request) {
 	respond(w, r, fismasystem, nil)
 }
 
-// clearUnscopedOnlyFields nils the 11 system-attribute fields only an
+// clearUnscopedOnlyFields nils the 9 system-attribute fields only an
 // unscoped-write admin may set. Called on INSERT when the acting user lacks
 // unscoped read access.
 func clearUnscopedOnlyFields(fs *model.FismaSystem) {
@@ -116,19 +119,21 @@ func clearUnscopedOnlyFields(fs *model.FismaSystem) {
 	fs.CloudVendor = nil
 	fs.SystemOperator = nil
 	fs.GocoCocGoGo = nil
-	fs.SystemOwner = nil
-	fs.SystemOwnerEmail = nil
 	fs.Legacy = nil
 }
 
-// preserveUnscopedOnlyFields overwrites the 11 system-attribute fields on
+// preserveUnscopedOnlyFields overwrites the 9 system-attribute fields on
 // incoming with the values already stored on existing. Called on UPDATE when
 // the acting user lacks unscoped read access, so a full-form PUT from a tier
 // that may not write these fields cannot wipe them.
 //
-// The set covers system attributes only. isso_name is a contact name that an
-// OpDiv-scoped admin may write on systems in their granted OpDivs, so it is
-// not part of this set and passes through from the request.
+// The set covers system attributes only. The contact fields - isso_name,
+// system_owner, system_owner_email - are writable by an OpDiv-scoped admin on
+// systems in their granted OpDivs (ztmf#511, ztmf#512), so they are not part
+// of this set and pass through from the request. The owner fields matter more
+// than a display preference: unlike isso_name they have no COALESCE fallback
+// to a user record, so the stored column is the only source there is, and
+// no onboarding load refreshes them for non-CMS OpDivs.
 func preserveUnscopedOnlyFields(existing, incoming *model.FismaSystem) {
 	incoming.HVA = existing.HVA
 	incoming.FIPS = existing.FIPS
@@ -138,8 +143,6 @@ func preserveUnscopedOnlyFields(existing, incoming *model.FismaSystem) {
 	incoming.CloudVendor = existing.CloudVendor
 	incoming.SystemOperator = existing.SystemOperator
 	incoming.GocoCocGoGo = existing.GocoCocGoGo
-	incoming.SystemOwner = existing.SystemOwner
-	incoming.SystemOwnerEmail = existing.SystemOwnerEmail
 	incoming.Legacy = existing.Legacy
 }
 
@@ -251,7 +254,7 @@ func SaveFismaSystem(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// Only OWNER and HHS_ADMIN may write the 11 system-attribute fields
+	// Only OWNER and HHS_ADMIN may write the 9 system-attribute fields
 	// (HasUnscopedRead gates this; HHS_READONLY_ADMIN is already blocked by
 	// IsAdmin() above). Every scoped admin can READ all of them - the list and
 	// GET reads return every column and only filter rows by OpDiv - so this is
--- a/backend/cmd/api/internal/controller/fismasystems_issoname_integration_test.go
+++ b/backend/cmd/api/internal/controller/fismasystems_issoname_integration_test.go
@@ -1,0 +1,51 @@
+package controller
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// The model-layer test in internal/model pins that FindFismaSystem CAN resolve
+// the ISSO display name. This one pins that GetFismaSystem actually ASKS it to.
+// Without this, dropping ResolveISSOName from the handler leaves the model test
+// green while ztmf#510 silently regresses: the single-system GET would go back
+// to returning the raw column and disagreeing with the list.
+//
+// Seed system 1002 carries isso_name NULL with an issoemail matching the
+// Admiral Piett user record, so a resolved response can only come from the
+// COALESCE fallback.
+func TestGetFismaSystemResolvesISSONameIntegration(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping database integration test")
+	}
+
+	r := httptest.NewRequest("GET", "/api/v1/fismasystems/1002", nil)
+	r = mux.SetURLVars(r, map[string]string{"fismasystemid": "1002"})
+	r = withUser(r, adminUser)
+	w := httptest.NewRecorder()
+
+	GetFismaSystem(w, r)
+	require.Equal(t, http.StatusOK, w.Code)
+
+	var resp struct {
+		Data struct {
+			FismaSystemID int32   `json:"fismasystemid"`
+			ISSOName      *string `json:"isso_name"`
+			ISSOEmail     *string `json:"issoemail"`
+		} `json:"data"`
+	}
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &resp))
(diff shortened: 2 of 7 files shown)


jsos3-cms:
Continuity note on the frontend side, since this PR's correctness now partly depends on it. Not a review, that is running separately.

## The owner fields need no UI companion

Correcting the coordination note: `system_owner` and `system_owner_email` are already editable and already in the write payload. Only `isso_name` carried the `readOnly` flag, so it was the sole extended field excluded from `EXTENDED_METADATA_KEYS`:

```
isso_name, hva, fips, system_type, cloud_system, cloud_service_model,
cloud_vendor, system_operator, goco_coco_gogo, system_owner,
system_owner_email, legacy
```

An OpDiv admin has always been able to edit those two and always got a success toast, with the backend discarding the value. This deploys and the existing UI starts working. So #512 is a pure backend fix and the companion work drops.

Help text is the only frontend nicety left, and not a dependency. Your point holds: "clearing restores the name from the user account" is true only where a derivation exists, so those fields need their own wording or none.

## What protects this change on the frontend

Two things, both on ztmf-ui#661, which is now a full PR with you on it.

**A strict dirty-diff, which this PR makes load-bearing.** The payload carries only fields whose value differs from the loaded system (`buildExtendedDiff`). Once the single-system read resolves `isso_name`, every detail page pre-fills a resolved name into the field the form writes back, so omission-when-untouched is what stops an admin's unrelated save from persisting that resolved name as a stored override.

The protection already existed. What changed on #661 is that it is no longer incidental: the dependency is documented at the function, and a test pins by name that an untouched resolved `isso_name` stays out of the payload, so a change to how that field is compared fails there rather than in someone's data.

Worth saying plainly: after this PR the frontend dirty-diff is the only thing between an ordinary save and an accidental permanent override, since nothing server-side rejects a write of `isso_name` equal to the resolved value. Reasonable place to land, but it is now a cross-repo invariant rather than a frontend detail.

**A post-save refetch.** The page used to echo its local draft into shared state after a successful PUT, which is accurate for a field whose stored value is the value sent. `isso_name` is the first one that is not: clearing it stores NULL and the read resolves a name, so the draft held an empty string where the saved system carries one, and the read view showed an em dash. It now refetches instead.

## This PR fixes a residual I documented on #510

The refetch left one path unresolved, recorded on #510: for a decommissioned system viewed with the toggle off, the refetch drops it from the active-only list, the fallback re-populates from `GET /fismasystems/:id`, and the raw column rendered as an em dash. There was no clean frontend fix, because no read available to the client returned a resolved name for a system outside the current list view.

**Making the single-system read resolve fixes that**, which is worth knowing since it was not the stated goal. The fallback fetch now returns the resolved name, so clearing the field on a decommissioned system displays correctly. If the review turns up a reason to narrow the single-read change, that residual comes back, so the two are coupled.

## Ordering

ztmf-ui#661 must not merge before #511 is deployed, or an OpDiv admin gets an editable field and a discarded write. This PR is stacked on #511, so the sequence is #511 merge and deploy, then #661, then this. If this lands before #661, nothing breaks; the frontend simply keeps sending owner-field values that now stick.

jsos3-cms:
Reviewed the top commit in isolation. The logic is sound: authorization parity with #511 on both create and update, three-state semantics identical across all three contact fields, the emberfall extension avoids both bad shapes #511 had to fix, and `openapi.yaml` regenerates byte-for-byte identical to what is committed. The 9-field rule holds, and better than the description argues: `system_operator` reads person-like but is enum-validated against the reference vocabulary, so the three fields that left are the only free-text ones in the set.

Two things to action, neither a code defect.

## 1. This will need a rebase after #511 merges

Not a "the diff looks bigger than it is" artifact. It produces real conflicts. Simulated by cherry-picking `0e0a574` onto `main` as a squash equivalent and merging this branch on top:

```
CONFLICT (content):  backend/cmd/api/internal/controller/fismasystems.go
CONFLICT (add/add):  backend/cmd/api/internal/controller/fismasystems_metadata_test.go
CONFLICT (content):  backend/emberfall_tests.yml
```

`model/fismasystems.go`, `openapi.yaml`, and the new integration test merge clean. The three above conflict because `061c5c7` edits the same lines `0e0a574` introduced: the renamed helper and test functions, the same doc comments, the same emberfall fixture block. Against the shared merge base each side looks like a different version of those hunks, so git cannot auto-resolve them.

Steps, in this order:

1. Merge #511.
2. `git rebase --onto main fix/isso-name-opdiv-writable feature/512-owner-fields-opdiv-writable`
3. Resolve the three files, keeping this PR's versions where they supersede #511's.
4. Force-push. GitHub will have retargeted the base to `main`.
5. Re-request review. The rebase dismisses any approval on this PR, since `dismiss_stale_reviews_on_push` is on for `main`.

Worth doing immediately after #511 lands rather than later, because until the rebase the diff view bundles all of #511 into this PR and makes further review cycles harder to read.

## 2. One line in the description undersells the exposure

The description says the single-system read resolving `isso_name` carries an exposure that already existed on the list endpoint. That is true for the ordinary detail-page load, which is list-fed, but not for the path that matters most here.

The frontend falls back to `GET /fismasystems/:id` when a system is absent from the list array, which in practice means a decommissioned system or a direct URL load before the list populates. That path previously returned the raw stored column, so it was the one path where a resolved name could not reach the edit form at all. `GetFismaSystem` now sets `ResolveISSOName: true`, so it returns the resolved name into the same field the form writes back.

Both consequences follow from the same change, and the good one is worth stating too:

- the em-dash-after-clearing case on that path disappears, which closes a residual recorded on #510 with no frontend change
- the accidental-override exposure extends to that path, where it did not exist before

The frontend guard holds either way: `buildExtendedDiff` omits any field equal to its baseline, so an untouched `isso_name` is never sent. But the delta is real. Before this, a broken diff on the fallback path would have written the raw value, a no-op. After this, it writes the resolved value, a genuine override. Suggest saying "widens the exposure to the single-system fallback path" rather than "already existed," so the follow-up for a separate resolved field is clearly motivated.

## Non-blocking

`TestFismaSystemISSONameResolutionIntegration` exercises `FindFismaSystem` directly rather than going through `GetFismaSystem`. Correctly skipped under `-short`, verified. But if someone dropped `ResolveISSOName: true` from the controller, that test still passes and #510 silently regresses. One assertion through the handler would pin the wiring.

## Disposition

Not requesting changes: there is no defect to fix, and the rebase cannot be done until #511 merges, so a block would be a signal you cannot act on yet. Leaving this as a comment rather than an approval, because the required rebase would dismiss an approval anyway. Happy to approve the rebased head once it is up, though a reviewer who did not write #511 looking at the combined result is probably healthier, since the base commit here is mine.

danielbowne:
Actioned both non-rebase items; pushed as 1f1a85e on top rather than amending, so the SHA you reviewed stays intact.

**Description corrected.** You are right that "already existed on the list endpoint" undersold it. It now says this widens the exposure to the single-system fallback path, and states both consequences of the same change - the em-dash residual on that path closing, and the override exposure reaching a path where it did not exist before. That is the motivation for a separate resolved field, so it should not read as pre-existing.

**Handler wiring now pinned.** Added `TestGetFismaSystemResolvesISSONameIntegration` in the controller package, going through `GetFismaSystem` rather than the model. Your point was exactly right: I removed `ResolveISSOName` from the handler and the model test stayed green. It fails now, and passes with the flag restored - verified rather than assumed.

Also caught while in there: the model-layer test checked out a pooled connection and never released it. Release moved into the same cleanup as the restore, since a bare defer would release it before `t.Cleanup` runs the restore Exec.

Suites after the change: integration green, emberfall 187/187.

On the rebase - agreed, and I will not start it until #511 lands. Your step list matches what I would do. Also agreed on the reviewer point: someone who did not write #511 looking at the combined result is healthier than either of us signing off on our own base.

danielbowne:

PR #515, opened 2026-08-03
feat(scores): expose status on reads and add an explicit confirm endpoint

costacalvin:
## Problem

FY2026 pre-fills every questionnaire answer from the prior cycle (`scores.status = 'not_started'`), and a question only counts toward Data Call Progress once it reaches `status = 'done'`. Agreeing with a pre-filled answer changes no answer field, so `Save`'s audit-preserving no-op guard (#412/#413) correctly refuses to record it — "reviewed all 40 and agreed", the most common data-call action, produced **no write at all**. An ISSO who did exactly that reported completion while the dashboard showed 14/40 (CMS-Enterprise/ztmf-ui#659). This repeats across the ~900 systems with carried-forward answers.

Two hard requirements are in tension, so neither can be relaxed:
- **Browsing must stay write-free** — viewers must never be stamped as editors (#413's guarantee, enforced server-side as defense in depth).
- **Agreeing must produce a write** — or progress never moves.

The resolution: agreement becomes its own explicit, attributable act.

## Changes

**1. Expose `scores.status` on the read paths.** Added to `model.Score`, the `FindScores` select/scan, and the `Save`/`Confirm` RETURNING clauses. The questionnaire can now mark carried-forward answers from the *same persisted fact* the progress count reads — deriving that state client-side from the absence of an edit event is what let the badge and the count disagree. No migration: the column and its values already exist (ztmf#435).

**2. `PUT /api/v1/scores/{scoreid}/confirm`.** Flips `status` to `'done'` and touches nothing else:

- `notes_is_ai_summary` survives — agreeing with an AI-drafted justification is not authoring it (the ordinary PUT clears the flag on edit, which is correct for edits)
- Runs through `queryRow`, so the existing `recordEvent` hook stamps the confirming user's identity and timestamp — the attribution is the point
- Same past-deadline rule as `Save` (`validateDeadline`, extracted so the two paths share it): non-admins cannot confirm into a closed cycle
- Idempotent on an already-`done` row
- **Authorization runs against the row's own `fismasystemid`** — the row is loaded first (`FindScoreByID`, extracted from `scoreUpdateIsNoOp`'s inline query); a client asserts nothing but the scoreid. Read-only tiers are rejected before any DB access, preserving the property `rbac_enforcement_test.go` pins
- `SaveScore`'s three write guards are extracted into `guardScoreWrite` and shared, so the confirm path cannot drift from the save path's rules

The ordinary `PUT /scores/{scoreid}` is untouched: its no-op guard remains unconditional, so no client can re-stamp an editor with an unchanged answer.

## Testing

- **Unit:** `guardScoreWrite` non-admin paths; `ConfirmScore` read-only rejection with no DB (`authorization_test.go`, `rbac_enforcement_test.go`)
- **Integration:** `TestConfirmScoreIntegration` — carried row (`not_started`, AI-flagged) → confirm → `status = 'done'`, answer fields byte-identical, `notes_is_ai_summary` preserved, confirming user stamped, `FindScoreProgress.questionsupdated` +1 exactly, idempotent re-confirm, past-deadline rejection, `ErrNoData` → 404 contract
- **E2E (emberfall):** 8-case HTTP matrix — owner confirm 200 with body (flag survival pinned), ISSO own-system 200 with attribution, ISSO unassigned 403, both read-only tiers 403, OPDIV_ADMIN outside grant 403, missing row 404
- `make test-unit`, `make test-integration`, `make test-e2e` all green; `openapi.yaml` regenerated via `make generate-openapi` + `lint-openapi`

## Acceptance criteria (backend half of ztmf-ui#659)

- [x] **Agreement is an explicit act, distinct from navigation**: the confirm endpoint exists solely for it; the ordinary PUT's audit-preserving no-op guard is untouched, so browsing/read-through can still never produce a write (#412/#413 preserved)
- [x] **Confirm flips `not_started` → `done` with the confirming user's own identity and timestamps** — attribution rides the same `recordEvent` → audit-lateral path as every save (`TestConfirmScoreIntegration`)
- [x] **No data migration**: nothing beyond the existing `not_started` state is needed — the column and values already exist; this PR only exposes them on reads
- [x] **Answer fields untouched by a confirm**: notes, option, and `notes_is_ai_summary` byte-identical after confirming (integration + emberfall pins)
- [x] **Progress counting unchanged**: `scoreprogress.go` is not modified — a confirmed row counts as updated because `status` flips, through the query that already exists
- [x] **Authorization parity with SaveScore**: shared `guardScoreWrite`; row loaded first and checked against its own `fismasystemid`; read-only tiers rejected before any DB access; same past-deadline rule as Save

## Deploy order

Deploy before CMS-Enterprise/ztmf-ui#664 (the UI degrades cleanly against a backend without this change — no badges, prior behavior — but the feature is inert until this ships).

### DIFF
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -233,6 +233,47 @@ func TestSaveScore_ISSONotAssignedForbidden(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
 
+// --- ConfirmScore ---
+
+func TestConfirmScore_ReadonlyAdminForbidden(t *testing.T) {
+	// The role-only rejection must fire before the row load, preserving the
+	// "read-only tiers are blocked before any DB access" property SaveScore
+	// pins - this test runs without a database.
+	r := httptest.NewRequest("PUT", "/api/v1/scores/123/confirm", nil)
+	r = mux.SetURLVars(r, map[string]string{"scoreid": "123"})
+	r = withUser(r, readonlyAdmin)
+	w := httptest.NewRecorder()
+
+	ConfirmScore(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
+// guardScoreWrite is the shared authorization for SaveScore and ConfirmScore.
+// The non-admin branches are pure (no DB), so pin them here; the admin branch
+// delegates to guardManageFismaSystem (DB) and is covered by the Emberfall
+// matrix.
+func TestGuardScoreWrite_NonAdminPaths(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("ReadonlyAdminForbidden", func(t *testing.T) {
+		assert.ErrorIs(t, guardScoreWrite(ctx, readonlyAdmin, 1), ErrForbidden)
+	})
+
+	t.Run("ISSOUnassignedForbidden", func(t *testing.T) {
+		assert.ErrorIs(t, guardScoreWrite(ctx, issoUser, 999), ErrForbidden)
+	})
+
+	t.Run("ISSOAssignedAllowed", func(t *testing.T) {
+		assigned := &model.User{
+			UserID:               "44444444-4444-4444-4444-444444444444",
+			Email:                "isso2@test.com",
+			Role:                 "ISSO",
+			AssignedFismaSystems: []*int32{int32Ptr(7)},
+		}
+		assert.NoError(t, guardScoreWrite(ctx, assigned, 7))
+	})
+}
+
 // --- SaveFismaSystemTargetMaturity (#398) ---
 
 func TestSaveFismaSystemTargetMaturity_ReadonlyAdminForbidden(t *testing.T) {
--- a/backend/cmd/api/internal/controller/rbac_enforcement_test.go
+++ b/backend/cmd/api/internal/controller/rbac_enforcement_test.go
@@ -79,3 +79,12 @@ func TestSaveScore_OpDivReadonlyForbidden(t *testing.T) {
 	SaveScore(w, r)
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
+
+// --- ConfirmScore: read-only tiers are blocked before any DB access ---
+
+func TestConfirmScore_OpDivReadonlyForbidden(t *testing.T) {
+	r := withUser(httptest.NewRequest("PUT", "/api/v1/scores/123/confirm", nil), opdivReadonly)
+	w := httptest.NewRecorder()
+	ConfirmScore(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
--- a/backend/cmd/api/internal/controller/scores.go
+++ b/backend/cmd/api/internal/controller/scores.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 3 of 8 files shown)


jsos3-cms (dismissed):
## Verdict

The design is right and the authorization on the new endpoint is right. One thing needs fixing before this merges: `Confirm` has no equivalent of `Save`'s audit-preserving guard, so confirming a row that is already `done` writes another event and moves the displayed editor to whoever confirmed last. That is a small fix and it is the only thing I would hold the merge for.

## The one blocking item

`Confirm` in `internal/model/scores.go:182` runs its UPDATE unconditionally. It never checks the current status, and because it goes through `queryRow`, `recordEvent` always fires. Since `lookupScoreAudit` projects the newest event, a confirm on a row that someone else already answered this cycle moves `last_edited_by` and `last_edited_at` to the confirming user. The answer content is untouched, and the `events` rows are append-only so the history is still there, but every read of that row now names the wrong person as the editor.

For an attribution field on FISMA compliance answers, that is the wrong default. It is also the exact case `Save` already defends against, in the same file, and the comment there states the reasoning explicitly: the guard is enforced server-side "as defense in depth even if the client adds its own dirty check." The frontend does currently hide the Confirm button once a row reads `done`, so hitting this needs a stale view, two users on the same system, or a direct API call. That is what makes it a small bug rather than a large one, but it is the same shape of client-guards-a-server-invariant that has bitten this codebase before, and the philosophy in `Save` is already the answer.

Suggested fix: short-circuit when the loaded row is already `done`, returning it unchanged through the same `lookupScoreAudit` path. That keeps the "repeat confirm returns 200" contract the integration test's third phase asserts, and removes the re-stamp. It mirrors `scoreUpdateIsNoOp` closely enough to read as the same pattern.

Worth adding a test that confirms twice as two different users and asserts the editor did not move. The current idempotency assertion only checks that status stays `done`, which passes either way.

## What I verified rather than assumed

**The new endpoint's authorization is sound, and it avoids a trap the save path does not.** `ConfirmScore` decodes no body at all. It loads the row through `FindScoreByID` and authorizes against that row's own `fismasystemid`, so a client asserts nothing except the scoreid. Read-only rejection happens before any database access, preserving the property the RBAC test pins. This is the pattern I would want everywhere.

**The three extractions carry no behavior drift.** `guardScoreWrite`, `validateDeadline`, and `FindScoreByID` are faithful line-for-line moves: same ordering, same error values, `ErrForbidden` and `ErrNoData` unwrapped as before, `trapError` untouched. I compared the pre- and post-refactor `SaveScore` specifically because a security-relevant refactor is where a dropped guard hides, and there is none.

**`notes_is_ai_summary` is genuinely preserved.** The column is absent from the SET clause entirely rather than written back from a possibly stale read, so Postgres never touches it. Correct, and the reasoning in the doc comment is the right one.

**No NULL-scan risk on the new column.** `status` is `NOT NULL DEFAULT 'not_started'` with a two-value CHECK from migration 0048, and that migration backfilled existing rows. So the non-pointer `Status string` scan is safe, and Confirm only ever writes a literal inside the CHECK domain.

**The tests are not vacuous.** Assertions are field by field rather than a struct compare, and the "PUT echoes a value that was already stored" problem does not apply here since Confirm sends no body. The emberfall matrix covers the tiers that matter.

## Follow-up, not blocking

A caller who is authenticated but not authorized for the target row can currently tell 403 apart from 404, because the row loads before the assignment check. That is a minor existence oracle on sequential integer ids, and it is outside the property the RBAC test pins. Not worth reordering the loads for on its own, but worth knowing.

## Merge coordination

This merges clean onto main after #511 lands, and clean again after #514, so it needs no rebase coordination with either. I verified that by simulating the squash rather than inferring it. The three files it shares with those branches are edited in disjoint regions.

costacalvin:
@jsos3-cms Addressed the audit-restamp feedback. Re-confirming an already `done` score now returns the existing row without writing another event, preserving the original editor and timestamp. The update also atomically requires `status <> 'done'`, covering concurrent requests that both loaded `not_started`. The integration test now confirms as a second user from a stale read and verifies attribution, timestamp, event count, and progress remain unchanged. Unit and integration suites pass.

danielbowne:
Tested this locally against the paired frontend branch, with the API container rebuilt on this branch and a full carry-forward fixture (~38.8k not_started rows in the open call).

Exercised as two different ISSOs on two different systems - one CMS system on the CMS-Cloud-AWS question edition, one HRSA system on the MAG edition - covering both the API directly and the UI flow. Confirm flips the row to done, the status now surfaces on the score read, progress moves off 0/40, the dashboard chip flips from Not updated to Updated, and the carried-forward score is preserved rather than blanked. Attribution landed on the confirming user in both cases.

Looks good end to end. The Confirm audit-guard point above is the only thing I would hold for.

jsos3-cms (approved):
Approving. The audit-restamp fix holds up on both the already-refreshed path and the genuine race, the conditional update cannot turn a legitimate second confirm into a 404, and the new test fails against the reverted fix and passes with it. Details in my dismissed review above.

The two cleanups noted there are worth picking up whenever this file is next open, and neither needs its own change: the doc comment that still describes recordEvent stamping on every call, and the unreachable fallthrough in the race-recovery branch.

Nice shape on the endpoint itself. Loading the row and authorizing against its own fismasystemid, rather than anything the client asserts, is the right pattern.

danielbowne:

PR #516, opened 2026-08-04
fix(scores): authorize score updates against the stored row

danielbowne:
Closes CMS-Enterprise/ztmf-misc#263 (internal - detail, impact, and severity rationale live there).

Stacked on #515: base is `fix/659-score-confirm-endpoint` and GitHub will retarget this to `main` when that merges. Only the top commit is new here.

## What changes

Score updates now authorize against the stored row instead of values carried in the request body, and the two parent columns become immutable on update.

- `PUT /scores/{scoreid}` loads the row, authorizes against its stored `fismasystemid`, and pins both parents onto the receiver. `POST` still authorizes against the body, which is correct there because no row exists yet - the asymmetry is deliberate.
- `fismasystemid` and `datacallid` leave the UPDATE's SET list. Saving an answer is not a reparenting operation.
- The deadline is evaluated against the cycle the row belongs to rather than one named by the caller.
- `ConfirmScore` in #515 is the shape this follows; it already loaded-then-authorized, and the two paths now agree.

## Hardening found in review

- The UPDATE is bound to `(scoreid, fismasystemid, datacallid)`, so a future caller that forgets to pin the parents updates zero rows and fails cleanly rather than writing a row it never authorized.
- Read-only tiers are rejected before the row lookup, matching `ConfirmScore` and the property pinned in `rbac_enforcement_test.go`. Without it a tier that must never reach the database could distinguish existing score ids from missing ones by the 403/404 difference.
- `Save` takes the already-loaded row via `WithCurrentScore`, so the added authorization read is offset rather than doubling reads on the questionnaire's per-Next PUT.

## Tests

- **emberfall**: the cross-system update at the HTTP layer, which is where this lived. The existing ISSO case only covered `POST`, the path this change deliberately leaves alone. Returns 403 now, 204 before. Suite 197/197.
- **integration**: the model-layer refusal, plus a correctly-parented save in the same test so it cannot pass against a `Save` that never updates anything.
- **unit**: the read-only ordering, asserted against a nonexistent scoreid so it discriminates. Verified to fail without the fix rather than assumed.

## Behavior notes for reviewers

- A receiver disagreeing with the stored row now errors instead of silently applying the answer half of the write. Stronger, and pinned by the integration test.
- For a caller who simply lacks the assignment, 404-before-403 on a missing id is unavoidable once authorization reads the stored row, and matches `ConfirmScore`.
- A client that submits a differing `fismasystemid`/`datacallid` has those values silently ignored rather than rejected. Returning 400 would be more honest to API consumers; the frontend never does this, so it is noted as a follow-up rather than changed here.

### DIFF
--- a/backend/cmd/api/internal/controller/rbac_enforcement_test.go
+++ b/backend/cmd/api/internal/controller/rbac_enforcement_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+	"github.com/gorilla/mux"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -78,6 +79,31 @@ func TestSaveScore_OpDivReadonlyForbidden(t *testing.T) {
 	w := httptest.NewRecorder()
 	SaveScore(w, r)
 	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
+// The UPDATE path has to be pinned separately from the create path above,
+// because a PUT that authorizes against the stored row must look that row up
+// and the lookup is easy to place before the role check.
+//
+// The assertion uses a scoreid that cannot exist, which is what makes this
+// test discriminate. Rejecting on role first is a 403. Looking the row up
+// first is a 404 - and that difference is exactly the leak: a tier that must
+// never touch the database could otherwise probe which score ids exist by
+// reading the status code. Asserting 403 on an existing id would prove
+// nothing, since guardScoreWrite rejects read-only tiers a few lines later
+// either way.
+func TestSaveScore_OpDivReadonlyForbiddenOnUpdate(t *testing.T) {
+	const missingScoreID = "2147483600" // no fixture allocates anything near this
+	body := jsonBody(t, map[string]any{"fismasystemid": 1002, "functionoptionid": 1, "datacallid": 3})
+	r := httptest.NewRequest("PUT", "/api/v1/scores/"+missingScoreID, body)
+	r = mux.SetURLVars(r, map[string]string{"scoreid": missingScoreID})
+	r = withUser(r, opdivReadonly)
+	r.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+	SaveScore(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code,
+		"a read-only tier must be rejected on role before the stored-row lookup, "+
+			"otherwise the 403/404 split reveals which score ids exist")
 }
 
 // --- ConfirmScore: read-only tiers are blocked before any DB access ---
--- a/backend/cmd/api/internal/controller/scores.go
+++ b/backend/cmd/api/internal/controller/scores.go
@@ -80,8 +80,14 @@ func SaveScore(w http.ResponseWriter, r *http.Request) {
 		log.Println(err)
 	}
 
-	if err := guardScoreWrite(r.Context(), user, score.FismaSystemID); err != nil {
-		respond(w, r, nil, err)
+	// Role-only rejection before any DB access, matching ConfirmScore and
+	// preserving the property pinned in rbac_enforcement_test.go ("read-only
+	// tiers are blocked before any DB access"). Without this the update path
+	// below would look the row up first, which both breaks that invariant and
+	// lets a read-only caller tell an existing scoreid from a missing one by
+	// the 404-vs-403 difference. guardScoreWrite re-checks this; harmless.
+	if user.IsReadOnlyAdmin() {
+		respond(w, r, nil, ErrForbidden)
 		return
 	}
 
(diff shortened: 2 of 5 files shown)


jsos3-cms:
Review in progress, no findings yet. Posting the merge mechanics ahead of it, since they affect how this lands rather than what it says.

## Retargeting after #515: rebase, do not use "Update branch"

When #515 squash-merges, GitHub retargets this to `main`, and at that moment the branch's history still points at the pre-squash commit. The merge base falls back to the commit before #515, so anything that *merges* the two sides sees both of them having edited the same regions of `controller/scores.go` and `model/scores.go` and conflicts.

I simulated both routes against the current heads:

- Plain merge, which is what the **"Update branch"** button does: conflicts on `controller/scores.go`, `model/scores.go`, and `rbac_enforcement_test.go`. It also produces a merge commit, which `required_linear_history` rejects on `main`.
- `git rebase --onto main fix/659-score-confirm-endpoint fix/misc263-score-write-authz`: **clean**, replaying exactly the one new commit.

So the rebase is the route, and it costs nothing:

```
git fetch origin
git rebase --onto origin/main origin/fix/659-score-confirm-endpoint fix/misc263-score-write-authz
git push --force-with-lease
```

Worth doing straight after #515 lands rather than later. Until then the diff view here bundles #515's whole change into this PR, and any approval collected in that window gets dismissed by the force-push anyway.

## On merge methods, since it came up

No repo setting needs changing. `ztmf` already permits squash, merge commits, and rebase merges; it is `ztmf-ui` that is squash-only. And rebase merges would not have helped the stacked case regardless, because a rebase rewrites commit hashes just as a squash does, so a child branch still fails to find its base in `main`. The only merge method that preserves ancestry for a stacked child is a true merge commit, and `required_linear_history` excludes that on `main`. A local `rebase --onto` is both the correct answer and a sufficient one.

## This PR currently has no gates

Because the base is a feature branch rather than `main`, the `main` ruleset does not apply to it. Its rules for required approval, required `dev` deployment, signatures, and linear history all scope to `refs/heads/main`. That is why no checks have reported here and why the merge state reads clean: the gates are absent, not satisfied.

Practically, this could be merged into #515's branch today with no approval, no CI, no dev deploy, and no signature check. Nobody appears to intend that, and reviewers are requested, so this is a note rather than a concern. It does mean the merge-state indicator on this PR carries no information until retargeting happens, at which point the full set applies and a `dev` deployment on the rebased head becomes necessary. Since `dev` is a single shared lane, a newer deployment from any other PR inactivates the one here and re-blocks it, so the deploy is best taken shortly before merging.

## Order across the two open stacks

`main <- #511 <- #514` and `main <- #515 <- #516`. Landing #511, then #514, then #515, then #516 works. #514 still needs its own `rebase --onto` after #511 and does conflict on three files, unlike this one.

## One small thing

`Closes CMS-Enterprise/ztmf-misc#263` will not auto-close across repositories. GitHub records no linkage for it here, so #263 needs closing by hand when this merges.

jsos3-cms (approved):
## Verdict

Approving. The defect is closed, and I verified that by reproduction rather than by reading: with the controller and model files reverted to the base while keeping your tests, the cross-system PUT succeeds with 204, and with the fix in place it returns 403. All seven acceptance criteria on the tracking issue are met.

Two things to pick up. Neither blocks the merge, and I would not hold a verified security fix for either, but both matter more than the usual follow-up and I would not want them to fall off. One is a missing regression test for the criterion most likely to silently reopen this. The other is a comment that now asserts the opposite of what the code does.

## Criteria

| # | Criterion | Result |
|---|---|---|
| 1 | PUT outside the caller's scope refused even when the body names a system they hold | Met, reproduced at the HTTP layer |
| 2 | Body `fismasystemid` mismatch does not move the row | Met, covered by the new integration test |
| 3 | Body `datacallid` mismatch does not move the row, deadline evaluated against the stored cycle | Met in code; the deadline half is untested, see below |
| 4 | Non-admin cannot write into a closed cycle by naming an open call | Met in code, verified by hand, not covered by any test |
| 5 | POST still authorizes against the body | Met, path untouched |
| 6 | `scoresEqualForUpdate` no longer compares immutable columns, or documents why | Met by documentation |
| 7 | Existing save and confirm behavior unchanged | Met, full unit and integration suites plus emberfall 197/197 |

## The one gap worth closing: criterion 4 has no test

Nothing in the PR exercises a closed-cycle row against an open-cycle claim in the body. The integration test cannot cover it, because `validateDeadline` exempts admins and that test runs as an admin, so the deadline branch never executes. The emberfall case reuses the same `datacallid` on both sides, so it never touches a closed cycle either.

The code is correct. Confirmed by hand against a closed-cycle row: a non-admin ISSO naming an open `datacallid` in the body gets 403 with the deadline error and the row is unchanged. So this is a coverage gap, not a defect.

It is worth closing because of what would silently reopen it. Moving the `score.DataCallID = stored.DataCallID` pin to after `Save` is called, or changing how the receiver is populated, would restore the bypass with every existing test still green. That is the failure mode this whole change exists to prevent, and it is the one path with no guard on it.

One integration case covers it: non-admin, a row on a closed cycle, an open `datacallid` in the body, assert `ErrPastDeadline`. Since this branch needs a force-push to rebase after #515 lands anyway, adding it costs no extra review cycle.

## The comment at `scoresEqualForUpdate` now contradicts the code

Non-blocking, but the most important of the two to actually land, because a wrong comment on this function is worse than no comment.

The first paragraph is accurate and worth keeping. The second one is not:

> Note what this does NOT do: a mismatch only fails the equality check, so the write proceeds as a normal partial update of the answer fields.

That was true before this commit. It is not true now, because the same commit bound the UPDATE to `scoreid=? AND fismasystemid=? AND datacallid=?`. A receiver that disagrees with the stored row matches zero rows and fails as `ErrNoData`. It does not proceed as a partial update.

Your own integration test asserts exactly that two lines from the comment, and the "Acknowledged behavior changes" section of the description gets it right. Only the comment is stale. Worth fixing because a future reader reasoning from it would conclude that mismatches are silently tolerated, which is the opposite of what the code now guarantees, and that is a dangerous thing to believe about this particular function.

Worth keeping distinct from the follow-up you already disclosed. That one is a controller-level, client-facing question about whether to return 400 instead of silently discarding a differing body value. This is a model-level description that the commit itself superseded.

## Notes, no action needed

**`WithCurrentScore` is safe as built, and the doc warning is doing real work.** Calling `Save` without it falls back cleanly to a fresh `FindScoreByID`, with no panic and no skipped authorization. More importantly a future misuse stays bounded: the value feeds only the no-op comparison, never the WHERE binding, which uses the receiver fields the controller pins from the stored row. So a wrong row passed here cannot become an authorization bypass.

Worth being precise about what it could become, though, because it is a little sharper than the comment implies. The no-op path returns `current` as the response body, so a future caller passing an unrelated row that happened to compare equal would hand back that row's content, not just record a wrong audit outcome. The "MUST be the one Save is about to update" line is therefore load-bearing rather than advisory, and I would leave it exactly as strongly worded as it is.

**The existence oracle now sits on a hotter path.** 404 before 403 on a missing id is unavoidable once authorization reads the stored row, and it matches `ConfirmScore`, so this is the accepted trade-off rather than a new one. Worth one line in the description acknowledging that it now applies to `PUT /scores/{id}` rather than only the much rarer confirm action. It reveals only whether an id exists, never the owning system, the content, or the cycle.

## On the suite

Five integration failures showed up in the full run. All five are pre-existing and unrelated: three fail identically on the base branch before your commit and are date and seed dependent, and the fourth passes in isolation on both base and head, so it is a full-suite ordering artifact. Nothing here regressed anything.

Merge mechanics are in a separate comment above, since they affect how this lands rather than what it does.

danielbowne:

PR #517, opened 2026-08-04
FY2026 backend batch: ISSO name, owner contact fields, score confirm, score write authorization

### LINKED ISSUE #512: Let OpDiv-scoped admins maintain system owner contact fields

danielbowne:
Combines the four reviewed backend PRs into a single branch so the FY2026 work merges and deploys once instead of four times. No code changes beyond what was already reviewed - the commits are replayed as-authored, with authorship preserved.

## What is in here

| Original | Author | What it does |
|---|---|---|
| #511 | @jsos3-cms | Lets OpDiv-scoped admins write the ISSO name |
| #514 | @danielbowne | Owner contact fields OpDiv-writable; one shared isso_name derivation across both reads (closes #512, #510) |
| #515 | @costacalvin | Exposes score status on reads and adds an explicit confirm endpoint (closes ztmf-ui#659 backend half) |
| #516 | @danielbowne | Authorizes score updates against the stored row (closes CMS-Enterprise/ztmf-misc#263) |

Review state carried in: #511, #515 and #516 were approved. #514 was reviewed with no defect found - its only outstanding item was a rebase that this branch performs, and the two non-rebase follow-ups from that review are already applied.

## Why combine

`main` requires a successful DEV deployment of the PR head, and DEV is a single lane. Merging these separately means four deploy-then-merge cycles, each needing a rebase of everything above it - and #514 genuinely conflicts on three files when rebased after #511. Doing it once removes both the queue and the conflict.

## How it was assembled

Branched from #514's head (which already contained #511), then cherry-picked #515 and #516 on top. All five replayed clean, no conflicts. The two stacks touch mostly disjoint code - fismasystems versus scores - overlapping only in `emberfall_tests.yml`, `authorization_test.go` and `openapi.yaml`, all of which merged automatically.

## Verification on the combined head

- `make generate-openapi` reproduces the committed spec byte-for-byte, so the CI drift check passes
- unit and integration suites green
- emberfall **197/197**, up from 187 on main
- both stacks' distinctive cases confirmed present rather than assumed: the OpDiv-writable contact-field assertions, the confirm-endpoint cases, and the cross-system update case that returns 403

## After this merges

The four original PRs should be closed as superseded, and CMS-Enterprise/ztmf-misc#263 closed by hand since the cross-repo `Closes` does not fire.

### DIFF
--- a/backend/cmd/api/internal/controller/authorization_test.go
+++ b/backend/cmd/api/internal/controller/authorization_test.go
@@ -233,6 +233,47 @@ func TestSaveScore_ISSONotAssignedForbidden(t *testing.T) {
 	assert.Equal(t, http.StatusForbidden, w.Code)
 }
 
+// --- ConfirmScore ---
+
+func TestConfirmScore_ReadonlyAdminForbidden(t *testing.T) {
+	// The role-only rejection must fire before the row load, preserving the
+	// "read-only tiers are blocked before any DB access" property SaveScore
+	// pins - this test runs without a database.
+	r := httptest.NewRequest("PUT", "/api/v1/scores/123/confirm", nil)
+	r = mux.SetURLVars(r, map[string]string{"scoreid": "123"})
+	r = withUser(r, readonlyAdmin)
+	w := httptest.NewRecorder()
+
+	ConfirmScore(w, r)
+	assert.Equal(t, http.StatusForbidden, w.Code)
+}
+
+// guardScoreWrite is the shared authorization for SaveScore and ConfirmScore.
+// The non-admin branches are pure (no DB), so pin them here; the admin branch
+// delegates to guardManageFismaSystem (DB) and is covered by the Emberfall
+// matrix.
+func TestGuardScoreWrite_NonAdminPaths(t *testing.T) {
+	ctx := context.Background()
+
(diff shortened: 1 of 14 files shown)


jsos3-cms:
LGTM, for the record :)

PR #519, opened 2026-08-04
fix(#513): read only in-app edits for score-progress last-updated

### LINKED ISSUE #513: scoreprogress `lastupdatedat` lateral has no action filter, so imported events can bump it (and the codebase disagrees whether they exist)
### Description

**Impact.**

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

### Environment

### Acceptance Criteria

### Severity

### Affected Component

voidspooks:
Closes #513.

`LastUpdatedAt` on the score-progress query took the newest `events` row of any action, while the status backfill in `0048scoresstatus.go` counts only `created` and `updated`. A score row whose only provenance is an out-of-band load therefore reported the load's timestamp while `questionsupdated` stayed 0, so the two read the same events and disagreed about them.

The lateral now carries the same action allowlist. An allowlist rather than a `<> 'imported'` exclusion, so a `deleted` event, or any action added later, cannot bump the timestamp either.

## The comments were the other half of the ticket

`scoreprogress.go` described history loaded from outside the app as carrying no events at all, while `0048scoresstatus.go` described it as carrying `imported` provenance events. Both statements are about environments rather than about the code, and neither is checkable from this repo, so I rewrote the `scoreprogress.go` side to state the rule instead: imported provenance is not an edit, whether or not a given environment records it. That holds either way and is what the query now enforces.

`0048scoresstatus.go` is untouched. It is the side already stating that rule, and it is an applied migration.

## Scope beyond the ticket's file list

`openapi.yaml` is a regeneration, not a hand edit. swag sources the `lastupdatedat` description from the struct field comment, so editing that comment moves the spec. Description text only, no schema or contract change, and `redocly lint` is clean at the usual 8 warnings with 2 explicitly ignored.

`_test_data_empire.sql` is a comment-only change. The data is unchanged, and it remains the fixture this fix is tested against. The old comment justified the imported events by saying last-updated shows the import instead of blank, which is the behavior being removed here, so it would have documented the opposite of the code.

## Visible behavior

A row whose entire event history is non-edit now reports no `lastupdatedat` instead of the load's timestamp. The field is `omitempty`, so it is absent rather than null. On the dashboard such a row's tooltip stops reading "Last updated <import date>" and falls through to its state-based text, and in `aggregateScores.ts` its `lastUpdatedMs` becomes -1, which puts it back in the tie where the has-score tie-break decides. That is the invariant the comment at `aggregateScores.ts:95` says the chosen-call logic depends on.

Rows with a genuine in-app edit are unaffected, since the edit is the newest matching event either way. The counts do not move at all; this changes only the timestamp.

## Tests

New `TestScoreProgressLastUpdatedIgnoresImportedEventsIntegration` drives the real `FindScoreProgress` path and asserts an imported-only system reports a nil `LastUpdatedAt` while still reporting its answers. It resolves its subject systems from the events table rather than hardcoding ids, so it cannot pass vacuously against a data call that simply has no events, and it mirrors the query's own exclusion of decommissioned systems. Verified to fail against the unfixed lateral, where it returns the import timestamp rather than nil.

`TestBuildScoreProgressSQL_Shape` gains one assertion pinning the action filter. The existing `TestImportedScoresKeepProvenanceWithoutDoneStatusIntegration` is unchanged and still passes; it pins the status half.

## Gate

`make test-unit` green across all 12 packages, run with `-count=1` rather than from cache. `make test-integration` green. `make test-e2e` ran 197 with 0 failures. Pre-commit and pre-push hooks both green, no `--no-verify`.

Snyk and the deploy checks will show red here: this is a fork PR and those jobs need org secrets. Expected, and they pass once the branch is pulled in-repo.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1001,11 +1001,11 @@ SELECT DISTINCT ON (s.scoreid)
 -- Import provenance for the FY2020 archives cycle (datacallid 6): a service
 -- account plus one 'imported' event per archived score, mirroring how
 -- externally-loaded history is attributed in real environments (ztmf#435). These
--- give the archived rows a who/when (last-updated shows the import instead of
--- blank) but use action 'imported', which the status-sync below deliberately
--- excludes - so the archived answers keep status 'not_started' despite carrying
--- events. This is the shape that proves imported != updated: answers present,
--- provenance present, never 'done'.
+-- give the archived rows a who/when in the audit log but use action 'imported',
+-- which both the status-sync below and the score-progress last-updated lateral
+-- deliberately exclude - so the archived answers keep status 'not_started' and
+-- report no last-updated, despite carrying events. This is the shape that proves
+-- imported != updated: answers present, provenance present, never 'done'.
 INSERT INTO public.users (userid, email, fullname, role, deleted, identity_provider)
 VALUES ('00000000-0000-4000-8000-000000000001', 'svc-importer@empire.test',
         'Imperial Archives Import Service', 'HHS_READONLY_ADMIN', false, 'okta')
--- a/backend/internal/model/scoreprogress.go
+++ b/backend/internal/model/scoreprogress.go
@@ -21,11 +21,11 @@ import (
 //     chip. A carried-forward answer does not count until re-saved (ztmf#299).
 //   - QuestionsAnswered ("has an answer at all") is the completion signal a
 //     PAST call needs. A closed cycle stops accumulating updates, and history
-//     imported from outside the app never had any (no events to backfill), so
-//     QuestionsUpdated says nothing about whether a historical call was in
-//     fact complete - QuestionsAnswered does (ztmf-ui#537). Cycles genuinely
-//     worked in-app keep their backfilled updated counts; accurate history,
-//     just not a completion signal.
+//     loaded from outside the app never counts as updated even where it
+//     carries provenance events, so QuestionsUpdated says nothing about
+//     whether a historical call was in fact complete - QuestionsAnswered does
+//     (ztmf-ui#537). Cycles genuinely worked in-app keep their backfilled
+//     updated counts; accurate history, just not a completion signal.
 //
 // Both are now read from persisted state (scores.status and score-row
 // presence) rather than reconstructed from the events audit log; see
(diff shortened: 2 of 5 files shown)


jsos3-cms:
@voidspooks rehomed this in-repo as #520 so the Snyk and deploy gates can run, since those need org secrets a fork PR cannot reach. Your commit is carried over unchanged: same SHA content, your authorship, and the signature still verifies. Nothing was rewritten.

Leaving this PR open rather than closing it now. It closes once #520 merges, pointing here, so your thread stays the place discussion lives until the change actually lands.

The review that happened off-PR is recorded on #520 so it is not lost, including the two non-blocking notes: the lateral scanning a row's events for imported-only rows rather than short-circuiting at `LIMIT 1`, and the cross-repo `aggregateScores` tie-break claim being worth confirming on dev once CI is green.

CI is already running on #520, which includes the openapi drift check that matters here given the spec regeneration.

Two process notes, neither about your change:

`Closes #513` has to live on #520 rather than here, because this PR will close without merging and the keyword would never fire.

You are not a collaborator on this repo, so I could not add you as an assignee on #520 and an approval from you would not satisfy the required review. You are credited in the body instead. Same as on ztmf-ui#654.

jsos3-cms:
@voidspooks converting this to draft, matching how ztmf-ui#646 is being held. Nothing is wrong with the change; this is about making the state say what is actually happening.

Your commit is queued in #521, the backend code-freeze batch, carried over unchanged with your authorship and signature intact. During the freeze approved work is held in that batch and merged as one, rather than PR by PR. So this PR is not the thing that merges, and draft makes that visible rather than leaving it looking like it is waiting its turn to merge on its own.

It stays open until #521 merges, then closes pointing there. Discussion still belongs here.

Review is done: a live-data pass plus an independent reproduction, both recorded on #521, and all eleven checks passed in-repo on your identical commit including the openapi drift check and the three Snyk jobs that a fork PR cannot reach.

jsos3-cms:
@voidspooks shipped. Your commit merged in #521, the backend code-freeze batch, as `b23c0a6`, carried over unchanged with your authorship and signature intact. #513 closed via the keyword declared there.

Thanks for this one. The review turned up something the ticket did not have: because the lateral had no action filter at all, a row with a genuine in-app edit could have its timestamp overwritten by a *later* import, so the fix reaches further than imported-only rows.

Closing by hand, since the batch squash-merged and GitHub cannot tell this PR's content landed.

PR #520, opened 2026-08-04
fix(#513): read only in-app edits for score-progress last-updated (rehome of #519)

### LINKED ISSUE #513: scoreprogress `lastupdatedat` lateral has no action filter, so imported events can bump it (and the codebase disagrees whether they exist)

jsos3-cms:
Rehome of #519 by @voidspooks, brought in-repo so the Snyk and deploy gates can actually run. The commit is his, unchanged and still signed. Original PR stays open until this merges.

Closes #513

## What it fixes

`LastUpdatedAt` on the score-progress query took the newest `events` row of any action, while the status backfill in `0048scoresstatus.go` counts only `created` and `updated`. A score row whose only provenance was an out-of-band load therefore reported the load's timestamp while `questionsupdated` stayed 0: the two read the same events and disagreed about them.

The lateral now carries the same action allowlist. An allowlist rather than excluding `imported`, so a `deleted` action, or any action added later, cannot bump the timestamp either.

The comment rewrite is the other half of the original ticket. `scoreprogress.go` described externally loaded history as carrying no events, while `0048scoresstatus.go` described it as carrying `imported` provenance events. Both are claims about environments rather than about the code, so the `scoreprogress.go` side now states the rule the query enforces: imported provenance is not an edit, whether or not a given environment records it. The migration is untouched, being both applied and already correct on that point.

`openapi.yaml` is a regeneration. swag sources the `lastupdatedat` description from the struct comment, so editing the comment moves the spec. Description text only. `_test_data_empire.sql` is comment-only: the old comment justified the imported events by saying last-updated shows the import rather than blank, which is the behavior being removed.

## Review already done, recorded here because it happened off the PR

Independent review checked this against live data and cleared it. Confirmed rather than assumed: the action allowlist exactly matches the 0048 backfill, the audit index keys only on scoreid and createdat so the filter rides on top of the index descent as the author noted, `openapi.yaml` regenerates byte-for-byte identical to what is committed, and the new integration test genuinely fails with the filter removed and passes with it. Integration suite green.

**Blast radius is historical only.** The affected rows are imported-only ones in closed cycles. No system in the open FY2026 data call is affected, so nothing can change for anyone mid-data-call. Exact figures were kept off this public PR and can be recorded internally if useful.

## Two non-blocking notes from that review

For an imported-only row the lateral now matches nothing and scans that row's events rather than short-circuiting at `LIMIT 1`. Bounded per row, but worth eyeballing an archives-year dashboard load rather than assuming it is free.

The author's claim that a nil last-updated feeds the `-1` tie-break in `aggregateScores.ts` is a cross-repo behavioral claim currently supported only by backend tests. Cheap to confirm on dev once CI here is green, and worth doing because the neighbouring invariant in that same function turned out to be upheld by a backend guarantee rather than by the frontend logic.

## Tests

New `TestScoreProgressLastUpdatedIgnoresImportedEventsIntegration` drives the real `FindScoreProgress` path and asserts an imported-only system reports a nil `LastUpdatedAt` while still reporting its answers. It resolves its subject systems from the events table rather than hardcoding ids, so it cannot pass vacuously against a data call with no events. `TestBuildScoreProgressSQL_Shape` gains an assertion pinning the action filter. The existing imported-provenance test is unchanged and still pins the status half.

Rows with a genuine in-app edit are unaffected, since the edit is the newest matching event either way. No counts move; this changes only the timestamp.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1001,11 +1001,11 @@ SELECT DISTINCT ON (s.scoreid)
 -- Import provenance for the FY2020 archives cycle (datacallid 6): a service
 -- account plus one 'imported' event per archived score, mirroring how
 -- externally-loaded history is attributed in real environments (ztmf#435). These
--- give the archived rows a who/when (last-updated shows the import instead of
--- blank) but use action 'imported', which the status-sync below deliberately
--- excludes - so the archived answers keep status 'not_started' despite carrying
--- events. This is the shape that proves imported != updated: answers present,
--- provenance present, never 'done'.
+-- give the archived rows a who/when in the audit log but use action 'imported',
+-- which both the status-sync below and the score-progress last-updated lateral
+-- deliberately exclude - so the archived answers keep status 'not_started' and
+-- report no last-updated, despite carrying events. This is the shape that proves
+-- imported != updated: answers present, provenance present, never 'done'.
 INSERT INTO public.users (userid, email, fullname, role, deleted, identity_provider)
 VALUES ('00000000-0000-4000-8000-000000000001', 'svc-importer@empire.test',
         'Imperial Archives Import Service', 'HHS_READONLY_ADMIN', false, 'okta')
--- a/backend/internal/model/scoreprogress.go
+++ b/backend/internal/model/scoreprogress.go
@@ -21,11 +21,11 @@ import (
 //     chip. A carried-forward answer does not count until re-saved (ztmf#299).
 //   - QuestionsAnswered ("has an answer at all") is the completion signal a
 //     PAST call needs. A closed cycle stops accumulating updates, and history
-//     imported from outside the app never had any (no events to backfill), so
-//     QuestionsUpdated says nothing about whether a historical call was in
-//     fact complete - QuestionsAnswered does (ztmf-ui#537). Cycles genuinely
-//     worked in-app keep their backfilled updated counts; accurate history,
-//     just not a completion signal.
+//     loaded from outside the app never counts as updated even where it
+//     carries provenance events, so QuestionsUpdated says nothing about
+//     whether a historical call was in fact complete - QuestionsAnswered does
+//     (ztmf-ui#537). Cycles genuinely worked in-app keep their backfilled
+//     updated counts; accurate history, just not a completion signal.
 //
 // Both are now read from persisted state (scores.status and score-row
 // presence) rather than reconstructed from the events audit log; see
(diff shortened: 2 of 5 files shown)


jsos3-cms:
Second independent review pass, run against this branch rather than reading the diff: SQL traced by hand, the fix reverted against a live database to confirm the new test catches it, `make generate-openapi` re-run, emberfall driven directly, and the frontend suite executed against ztmf-ui's actual `origin/main`. Verdict is the same, **go**. It corroborates the earlier live-data review on all four of its confirmed points, and adds two things worth having on the record.

## This fixes more than imported-only rows

The ticket and the description both frame this as rows whose *entire* provenance is a load. It is broader than that. Because the old lateral took the newest event of any action, a row with a genuine in-app edit could have its displayed timestamp overwritten by a **later** out-of-band import. So an ISSO's real edit could be relabelled with an ETL run's timestamp.

The fix closes that too, and for the same reason: the action filter is in the `WHERE`, so the descent picks the newest *matching* event regardless of how imports interleave. Verified in both orderings rather than by example. Worth stating because it makes the change more valuable than advertised, and because "only affects historical cycles" is true of the reported symptom but the interleaving case was not bounded to closed calls.

## The cross-repo claim holds, and the compensating logic is already tested

The author's note that a nil last-updated feeds the `-1` tie-break in `aggregateScores.ts` is correct, and it is already pinned by a passing frontend test: `keeps the score when the scored call row has a null lastupdatedat`, whose own comment names the imported-call case explicitly. Confirmed against ztmf-ui `origin/main` at its current tip, not a stale checkout, which mattered because a combined PR landed there mid-review and rewrote both of the consumers involved.

Worth being precise about which side guarantees what, since a neighbouring invariant in that same function turned out earlier to be upheld by the backend rather than the frontend. The "scoreless implies null last-updated" property comes from the join structure and is untouched here. What this PR newly establishes is the parallel "imported-only implies null last-updated", which is what makes the imported-call case actually reach the tie-break it was written for. Before this, that case could win outright on a recent import timestamp and bypass the tie-break entirely.

One concern raised and then closed rather than left hanging: whether ztmf-ui#654, which retires a score-presence proxy, might remove that tie-break. It does not touch `aggregateScores.ts` at all. The `hasScore` identifier remaining in that file is an unrelated local variable in the tie-break loop, and the branch at `aggregateScores.ts:104` is intact.

## Non-blocking suggestion

`('created', 'updated')` is now written literally in four places: the migration, this query, and two raw SQL fragments in the integration test. Nothing is wrong today, and the migration is applied history that will not itself gain an action. But two independent SQL fragments silently disagreeing about which actions count is exactly what produced the original bug, so a shared Go constant referenced by the query and the tests would remove the possibility of a future action value updating one copy and not the others. Cheap, and it can follow separately rather than holding this.

## Independently reproduced

Unit suite green. Full integration suite 0 failures, including the three tests that are sometimes date-dependent on main. Emberfall `Ran: 197 / Failed: 0`. `make generate-openapi` produces a file byte-identical to the committed spec, so the drift check here will pass. `_test_data_empire.sql` confirmed comment-only with the insert below it untouched. The comment about the audit index is accurate: `events_score_audit_idx` keys on scoreid and `createdat DESC` with no action column, so the filter is a post-scan row filter as stated rather than index-native.

jsos3-cms:
Superseded by #521, which carries the identical commit on `freeze/backend-approved-batch`.

Closing this rather than reframing it in place. A pull request cannot change its head branch, so leaving the backend freeze batch on a branch named `rehome/519-...` would have misdescribed it for anyone auditing the freeze later. #521 sits on a branch named to match ztmf-ui#674, so the two repositories' freeze artifacts read the same way.

Nothing is lost. @voidspooks' commit is byte-identical and still verifies, the review recorded here is carried over in full, and this PR's CI run (`30941730744`, eleven checks green including the three Snyk jobs, the openapi drift check, integration, and the dev deploy) is quoted on #521 as the proof that the in-repo move did what it was for.

#519 stays open and closes when #521 merges.

PR #521, opened 2026-08-04
Code freeze batch: backend approved fixes (score-progress last-updated)

### LINKED ISSUE #513: scoreprogress `lastupdatedat` lateral has no action filter, so imported events can bump it (and the codebase disagrees whether they exist)

jsos3-cms:
Backend counterpart to the frontend freeze batch in CMS-Enterprise/ztmf-ui#674. Running batch of approved work, draft until the remaining approvals land, then converted manually.

Closes #513

## What is in it

| Source | Author | Change | Closes |
| --- | --- | --- | --- |
| #519 | @voidspooks | Read only in-app edits for score-progress last-updated | #513 |

@voidspooks' commit is carried unchanged from his fork: same content, his authorship, and the signature still verifies. Nothing was rewritten or re-signed.

This supersedes #520, which was opened as a standalone rehome before the batch approach was settled. Same branch content, renamed so the freeze artifacts match across the two repositories.

## Why this is more than the ticket described

`LastUpdatedAt` took the newest `events` row of any action, while the status backfill in `0048scoresstatus.go` counts only `created` and `updated`. So a row whose only provenance was an out-of-band load reported the load's timestamp while `questionsupdated` stayed 0.

Review surfaced a second case nobody had stated: because the filter was absent entirely, a row with a **genuine in-app edit** could have its displayed timestamp overwritten by a **later** import. An ISSO's real edit could be relabelled with an ETL run's timestamp. That case is not bounded to closed cycles the way the reported symptom is, so the fix is worth more than its description claims.

The allowlist is applied in the `WHERE`, so the descent picks the newest matching event regardless of how imports interleave. Verified in both orderings rather than by example.

## CI proof, which is the point of bringing this in-repo

A fork PR cannot reach org secrets, so on #519 the Snyk jobs, the deploy, and the integration tests never ran. In-repo they do, and all eleven checks pass on the identical commit (run `30941730744`):

| Check | Result |
| --- | --- |
| Analysis / snyk test | pass, 22s |
| Analysis / snyk code test | pass, 37s |
| Analysis / snyk iac test | pass, 34s |
| Analysis / openapi spec up to date | pass, 41s |
| Analysis / lint go | pass, 1m8s |
| Analysis / lint terraform | pass, 7s |
| Analysis / detect backend changes | pass, 5s |
| Check for Changes | pass, 7s |
| Backend / Integration Tests | pass, 1m36s |
| Backend / Build and Smoke Test | pass, 2m24s |
| Infrastructure / Deploy | pass, 3m15s |

The openapi drift check passing matters specifically: `openapi.yaml` here is a regeneration, since swag sources the `lastupdatedat` description from the struct comment. Two independent reviews confirmed it regenerates byte-for-byte identical to what is committed, and CI now agrees.

## Review already done, recorded because it happened off the PR

Reviewed against live data and cleared. Confirmed rather than assumed: the action allowlist exactly matches the `0048` backfill, the audit index keys only on scoreid and createdat so the filter rides on top of the index descent as the author noted, the spec regenerates identically, and the new integration test genuinely fails with the filter removed and passes with it.

**Blast radius is historical only.** The affected rows are imported-only ones in closed cycles, and no system in the open FY2026 data call is affected, so nothing changes for anyone mid-data-call. Exact figures were kept off this public PR and can be recorded internally.

A second independent pass reproduced all of it, including emberfall at `Ran: 197 / Failed: 0`, and confirmed the cross-repo claim: a nil last-updated feeds the `-1` tie-break in ztmf-ui's `aggregateScores.ts`, which is already pinned there by a passing test naming the imported-call case explicitly.

## Non-blocking notes carried forward

For an imported-only row the lateral now matches nothing and scans that row's events rather than short-circuiting at `LIMIT 1`. Bounded per row, but worth eyeballing an archives-year dashboard load rather than assuming it is free.

`('created', 'updated')` is now written literally in four places: the migration, this query, and two raw SQL fragments in the integration test. Two independent SQL fragments disagreeing about which actions count is exactly what produced the original bug, so a shared Go constant would close that off. Can follow separately.

## How this branch is maintained

Members are reviewed work. **The approval lives on this batch, not on the member PRs**, and it is collected once the queue is final rather than per addition, because any push here dismisses an approval already given. So a member PR sitting at `REVIEW_REQUIRED` is expected and is not a gap: during the freeze the batch is the thing that gets approved and merged, and the member is the thing that got reviewed. Each member's review is recorded on this PR so the trail does not depend on where the conversation happened.

Additions are appended, never reordered, which keeps the push a fast-forward, avoids re-hashing contributors' commits, and preserves the record of what joined when.

This PR stays a draft while the queue fills. That is deliberate, not an oversight: it means no CI has run here yet, since draft PRs skip both the analysis and deploy jobs.

After merge, #519 closes without merging, so its own `Closes #513` never fires. That is why the keyword is declared here.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1001,11 +1001,11 @@ SELECT DISTINCT ON (s.scoreid)
 -- Import provenance for the FY2020 archives cycle (datacallid 6): a service
 -- account plus one 'imported' event per archived score, mirroring how
 -- externally-loaded history is attributed in real environments (ztmf#435). These
--- give the archived rows a who/when (last-updated shows the import instead of
--- blank) but use action 'imported', which the status-sync below deliberately
--- excludes - so the archived answers keep status 'not_started' despite carrying
--- events. This is the shape that proves imported != updated: answers present,
--- provenance present, never 'done'.
+-- give the archived rows a who/when in the audit log but use action 'imported',
+-- which both the status-sync below and the score-progress last-updated lateral
+-- deliberately exclude - so the archived answers keep status 'not_started' and
+-- report no last-updated, despite carrying events. This is the shape that proves
+-- imported != updated: answers present, provenance present, never 'done'.
 INSERT INTO public.users (userid, email, fullname, role, deleted, identity_provider)
 VALUES ('00000000-0000-4000-8000-000000000001', 'svc-importer@empire.test',
         'Imperial Archives Import Service', 'HHS_READONLY_ADMIN', false, 'okta')
--- a/backend/internal/model/scoreprogress.go
+++ b/backend/internal/model/scoreprogress.go
@@ -21,11 +21,11 @@ import (
 //     chip. A carried-forward answer does not count until re-saved (ztmf#299).
 //   - QuestionsAnswered ("has an answer at all") is the completion signal a
 //     PAST call needs. A closed cycle stops accumulating updates, and history
-//     imported from outside the app never had any (no events to backfill), so
-//     QuestionsUpdated says nothing about whether a historical call was in
-//     fact complete - QuestionsAnswered does (ztmf-ui#537). Cycles genuinely
-//     worked in-app keep their backfilled updated counts; accurate history,
-//     just not a completion signal.
+//     loaded from outside the app never counts as updated even where it
+//     carries provenance events, so QuestionsUpdated says nothing about
+//     whether a historical call was in fact complete - QuestionsAnswered does
+//     (ztmf-ui#537). Cycles genuinely worked in-app keep their backfilled
+//     updated counts; accurate history, just not a completion signal.
 //
 // Both are now read from persisted state (scores.status and score-row
 // presence) rather than reconstructed from the events audit log; see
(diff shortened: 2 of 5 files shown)


jsos3-cms:
@danielbowne requesting your review on this one. It is the backend release train for the freeze, and the approval is the only thing left.

**All 11 checks green** on `65131e85` from the post-conversion run `30954672674`, including the three Snyk jobs, the openapi drift check, integration tests, and the dev deploy. The dev deployment is the active one on this PR's own branch ref, so the `required_deployments` gate is satisfied rather than merely stale-green.

One commit, @voidspooks' unchanged from his fork, authorship and signature intact. It fixes the score-progress `lastupdatedat` lateral to read only in-app edits.

What is worth your attention rather than re-reviewing the diff:

The review already done is recorded in the description and in a follow-up comment, including one finding the ticket did not have. The original framing was imported-only rows, but with no action filter at all, a row with a **genuine in-app edit** could have its timestamp overwritten by a **later** import. So an ISSO's real edit could be relabelled with an ETL run's timestamp, and that case is not bounded to closed cycles the way the reported symptom is.

The `Closes #513` keyword lives here rather than on #519, because #519 closes without merging and would never fire it.

Two non-blocking notes carried on the PR: the lateral now scans an imported-only row's events rather than short-circuiting at `LIMIT 1`, worth eyeballing on an archives-year dashboard load; and `('created', 'updated')` is now written literally in four places, so a shared Go constant would remove the drift that caused the original bug.

danielbowne (approved):
LGTM. Content is byte-identical to #519, which I reviewed in depth, and Cameron's authorship is intact on the commit.

Carrying that verification forward: the action allowlist exactly matches the 0048 status backfill; the audit index keys on scoreid and createdat only, so the note about the filter riding on top of the index descent is accurate rather than assumed; openapi regenerates byte-for-byte identical to what is committed; and the new integration test genuinely fails when the filter is removed and passes with it, which I checked by removing it rather than trusting the description.

Blast radius is historical only - roughly 135k imported-only score rows, and zero systems in the open FY2026 call, so nothing here can affect anyone answering right now.

Two things I raised on #519 that remain non-blocking and unchanged: for imported-only rows the lateral finds nothing matching and scans that row's events rather than short-circuiting at LIMIT 1 (bounded per row, but worth a look at an archives-year dashboard load), and the claim that a nil last-updated feeds the -1 tie-break in aggregateScores is a cross-repo behavioural claim verified only by backend tests.

All CI green including Snyk and the DEV deploy, which is what the rehome was for.

PR #522, opened 2026-08-04
feat(users): expose last_seen on the users list

danielbowne:
**This is the backend piece only.** It exposes the data; how it gets shown is deliberately left open. Filed separately as CMS-Enterprise/ztmf-ui#675 so the team can decide the presentation. Until then the field is queryable through `/api/v1/users`, which is enough to answer the question that prompted this.

## What prompted it

There is currently no way to tell whether a provisioned account has ever been used. `users` has no `last_login` column, so "has this person shown up" can only be answered by querying the events log by hand — which is what we have been doing during the FY2026 call. With ~1,250 provisioned respondents and most accounts created by a bulk load rather than by signup, that question comes up constantly.

## What this adds

Two halves that meet in the middle:

**A writer.** `RecordLogin` appends a `session / created` event when the post-OIDC session handler mints a cookie. Until now every event in the log was a side effect of a *write*, so activity was only visible for users who changed something — and a privileged account that signs in and only reads looked identical to one that had never signed in at all. That is the exact pair you need to separate to find a stale account, and it was the one pair the log could not distinguish.

It is fire-and-forget: the session already exists and the cookie is already set by that point, so failing the redirect over an audit write would deny a user access for a bookkeeping problem. Logged rather than returned, matching what `recordEvent` already does for write-derived events. An empty user id writes nothing.

**A reader.** `last_seen` on the users list: the newest `events` row for that user, derived on read rather than stored. Nothing to backfill, and it cannot drift from the log it summarises. Because it is `MAX(createdat)` over *all* events, logins land in it for free — no change to the query.

The remaining caveat is narrower than it was: the log is not retrospective, so null on an account that predates this means "nothing since then", not "never". That ambiguity ages out as people sign in. The field comment says so plainly.

## Migration 0054 — the part not to skip

`events` had no `userid` index. It carries one on `(resource, createdat)` and a partial one on the scoreid expression, neither of which serves a lookup keyed on `userid`.

Without an index the correlated subquery is a sequential scan of `events` per user row — roughly 200k events against 1,250 users, on a page administrators load routinely. That is precisely the regression migration 0037 was written to avoid for the score audit lateral. 0054 adds `(userid, createdat DESC)` so the `MAX` is an index-only descent to the first matching entry rather than an aggregate over the whole partition.

Not partial: unlike the score audit index there is no single resource worth narrowing to, because `last_seen` deliberately counts activity of any kind.

## `findUser` selects a literal NULL, and that is load-bearing

`findUser` backs `FindUserByID` / `FindUserByEmail`, which run in the **auth middleware on every authenticated request**. It must not pay for a correlated subquery that only the list view renders.

But the column still has to appear there. That path scans strictly by name (`pgx.RowToStructByName`), so omitting it fails the scan outright rather than leaving the field nil — which would have broken every authenticated request. My first pass did exactly that, on the assumption the path scanned laxly; the integration test below caught it before it left the branch.

## Tests

Both suites resolve their subjects from the database rather than hardcoding ids, so neither can pass vacuously against a fixture that happens to lack the relevant rows.

`TestFindUsersLastSeenIntegration`:
- a user with activity reports the newest event for that user, checked against a direct `MAX(createdat)`
- a user with no activity reports nil rather than a zero time
- `findUser` does **not** populate the field, pinning the hot-path exclusion so a later "make it consistent" edit has to be deliberate

`TestRecordLoginIntegration` runs against a user chosen for having *no* events at all, so the assertions cannot be satisfied by pre-existing activity:
- the event lands with the right action, resource, and payload
- it surfaces as `last_seen` **without that user ever writing anything** — the whole point of the change
- an empty user id is a no-op, so we never fabricate an event with no initiator

## Review notes

`last_seen` intentionally has no `omitempty`. A never-used account is the case this field exists to identify, and omitting the key there would make the answer indistinguishable from the field not being served at all. Null is the answer, so it ships. (Caught in review — the first version had `omitempty` and silently dropped the field for exactly the population it was built to surface.)

Suites: unit, integration, and emberfall 197/197 all green. `make generate-openapi` committed.

### DIFF
--- a/backend/cmd/api/internal/auth/session.go
+++ b/backend/cmd/api/internal/auth/session.go
@@ -176,6 +176,15 @@ func SessionHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	SetSessionCookie(w, token)
+
+	// Record the login. Fire-and-forget by design: the session is already
+	// minted and the cookie already set, so failing the redirect over an audit
+	// write would deny a user access for a bookkeeping problem. Logged, not
+	// returned - the same trade recordEvent makes for write-derived events.
+	if err := model.RecordLogin(r.Context(), user.UserID); err != nil {
+		log.Printf("login: failed to record session event for %s: %s\n", user.UserID, err)
+	}
+
 	http.Redirect(w, r, "/", http.StatusFound)
 }
 
--- a/backend/cmd/api/internal/auth/session.go
+++ b/backend/cmd/api/internal/auth/session.go
@@ -177,10 +177,14 @@ func SessionHandler(w http.ResponseWriter, r *http.Request) {
 
 	SetSessionCookie(w, token)
 
-	// Record the login. Fire-and-forget by design: the session is already
-	// minted and the cookie already set, so failing the redirect over an audit
-	// write would deny a user access for a bookkeeping problem. Logged, not
-	// returned - the same trade recordEvent makes for write-derived events.
+	// Record the login. Fire-and-forget by design: the session is minted and
+	// the Set-Cookie header staged, so failing the redirect over an audit write
+	// would deny a user access for a bookkeeping problem. Logged, not returned -
+	// the same trade recordEvent makes for write-derived events.
+	//
+	// The failure direction is the safe one: a dropped write makes an active
+	// account look stale, which costs a needless review, and can never make a
+	// stale account look active.
 	if err := model.RecordLogin(r.Context(), user.UserID); err != nil {
 		log.Printf("login: failed to record session event for %s: %s\n", user.UserID, err)
 	}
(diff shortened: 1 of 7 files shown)


MackOverflow (approved):
clean, no nits

jsos3-cms:
@danielbowne shipped. Your three commits merged in #529, the backend code-freeze batch, as `96bf66a`, with your authorship and signatures intact. Prod deploy confirmed live.

Closing by hand: the batch squash-merged, so GitHub cannot match this PR's content and would leave it open indefinitely.

Two notes for the record. This PR declared no closing issue by design, since it is the backend piece and presentation stays open on CMS-Enterprise/ztmf-ui#675. And your `chore(ci): retrigger DEV deploy under a fresh image tag` commit was deliberately skipped rather than carried: it existed to get a fresh image tag past the immutable ECR repo, and the batch got its own SHA anyway.

PR #524, opened 2026-08-05
chore(model): single home for event-action and score-status values

### LINKED ISSUE #445: Tech-debt follow-ups from #437: shared event-action constant + scores.Save status seam
### Type

Chore

### Description

Two non-blocking follow-ups surfaced during the review of #437 (which implemented #435, persisting per-answer `scores.status`). Neither blocks anything that shipped; capturing them so they aren't lost.

**1. Share the score-event action strings instead of repeating literals.**

The values `'created'` / `'updated'` / `'imported'` are hand-repeated as bare string literals in three places that must stay in agreement:

- `backend/internal/model/events.go` — `recordEvent` assigns `"created"` / `"updated"`.
- `backend/cmd/api/internal/migrations/0048scoresstatus.go` — the status backfill filters `action IN ('created','updated')`.
- `backend/_test_data_empire.sql` — the seed status-sync repeats the same predicate, plus `'imported'` for provenance rows.

There is no shared source of truth, so a rename or a new action value could silently desync the backfill from what the app actually writes — the backfill would quietly mark the wrong rows (or miss rows) with nothing to catch it. Introduce a single Go definition for these action values (and reference it from the migration/seed comments) so the contract lives in one place.

**2. Give `scores.Save` a status path a future importer can reuse.**

`scores.Save` hardcodes `status = 'done'` on every create/update (`backend/internal/model/scores.go`), and `copyPreviousScores` sets `'not_started'`. A future importer that wants to load history as `not_started` plus an `'imported'` provenance event has no seam in `Save` — it would have to bypass `Save` with raw SQL, exactly as the seed and the 0048 backfill do today. When the real import feature is built, either add an explicit status/provenance parameter to the write path, or document which of the three score-mutating paths it should extend.

Both items are code hygiene / forward-looking; no user-facing impact.

Refs #437, #435.

danielbowne:
Closes #445.

Both follow-ups from the #437 review, taken as a no-behavior-change chore: every diff line outside comments is a literal-to-constant substitution, and the emitted SQL and stored values are byte-identical.

## Event actions get one home

`created` / `updated` / `deleted` / `viewed` / `imported` now live as one const block in `internal/model/events.go`, used by every Go writer (`recordEvent`'s switch, `RecordQuestionView`, and the three transaction-manual audit inserts in `fismasystems.go` and `users.go`).

The doc comment carries the part that actually matters: these are stored data, not an internal enum. Migration 0048's backfill selects on `action IN ('created','updated')` and has already run in every environment; the seed status-sync and the progress query repeat the same literals. Renaming a **value** is therefore a data migration — backfill the events table, update every SQL predicate — while renaming the Go identifier is free. The SQL sites can't reference Go constants, so they get comments pointing back at the const block as the source of truth.

`eventActionImported` has no Go writer by design (it marks provenance for data loaded outside the app), so it carries a `lint:ignore U1000` explaining itself. One caveat verified empirically: staticcheck will NOT flag the directive as unmatched if a Go writer ever appears — U1000 ignores are exempt from its unmatched-directive check — so the comment says to delete it by hand. This is the one judgment call in the diff worth a reviewer's eye.

## Score statuses too, plus the importer contract on Save

`done` / `not_started` get the same treatment in `scores.go` (the 0048 CHECK constraint pins the set, so the same "changing a value is a migration" warning applies).

Per the issue's either/or, `Score.Save` documents which path a future importer extends rather than speculatively redesigning the write path: Save is the interactive-human write path — it flips status to done and records created/updated, exactly what the 0048 backfill treats as "a human answered this cycle" — so an importer either gets an explicit status/provenance `ScoreSaveOption` or writes its own path recording `imported`. Two deliberate keeps are commented in place: the `copyPreviousScoresOpDivHardcoded` statement stays a compile-time const string, and the `copyPreviousScores` select fragment interpolates the constant with `fmt.Sprintf` rather than a bind parameter (squirrel would misorder the placeholder args; commented at the site).

## Known convergence with #522

#522 adds `RecordLogin`, which passes `"created"` as a literal — invisible from this branch since both fork from main. Whichever lands second (or the batch branch carrying both) converts that one call site to `eventActionCreated`. One line, noted here so it isn't lost.

## Testing

- `go build`, `go vet`, `staticcheck` — clean
- `make test-unit`, `make test-integration` — pass
- `make generate-openapi` — zero diff (no API change, so no emberfall changes either)

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1029,6 +1029,12 @@ SELECT '00000000-0000-4000-8000-000000000001', 'imported', 'public.scores',
 -- FY2020 archives block) are excluded via the same action filter the migration
 -- uses, so imported history stays not_started. Keeps this consistent
 -- automatically as the events seed above changes.
+--
+-- The action values here (and in the 'imported' events seeded above) are SQL
+-- literals because SQL cannot reference a Go constant; they are the same values
+-- defined as eventAction* in internal/model/events.go, which is the source of
+-- truth for the set. Changing an action value there means updating these
+-- literals and backfilling stored rows, not a rename.
 UPDATE public.scores s
 SET status = 'done'
 WHERE EXISTS (
--- a/backend/cmd/api/internal/migrations/0048scoresstatus.go
+++ b/backend/cmd/api/internal/migrations/0048scoresstatus.go
@@ -41,6 +41,13 @@ ALTER TABLE public.scores
 -- loaded outside the app is attributed with 'imported' provenance events so it
 -- carries a who/when, but an import is not a human answering this cycle - those
 -- events must NOT flip the row to done, so they are excluded here.
+--
+-- The action values below are SQL literals because SQL cannot reference a Go
+-- constant; they are the same values defined as eventAction* in
+-- internal/model/events.go, which is the source of truth for the set. This
+-- migration has already run in every environment, so its SQL is frozen -
+-- changing an action value means backfilling the events table and updating
+-- every SQL predicate that repeats these literals, not editing this file.
 UPDATE public.scores s
 SET status = 'done'
 WHERE EXISTS (
--- a/backend/internal/model/events.go
+++ b/backend/internal/model/events.go
@@ -52,16 +52,25 @@ const (
 	// eventActionImported marks provenance for score data loaded outside the
 	// application - bulk imports and seed SQL, not a human answering a
 	// questionnaire. Nothing in this package writes it today; the value lives
-	// here so there is one authoritative home for it, because the readers that
-	// exclude it (0048's backfill, the seed status-sync, the last-updated
-	// lateral in scoreprogress.go) all depend on the exact spelling. A future
-	// bulk importer must write this action rather than reusing the in-app
-	// create/update path - see Score.Save.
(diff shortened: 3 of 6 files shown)


MackOverflow (approved):
looks good, a nit that we can place as a P3 for backlog: 

SQL comment fan-out. The source-of-truth comments now live in ~4 places (events.go const block, scores.go const block, migration 0048, _test_data_empire.sql). If a value ever does change, someone must find all of them. Not worth changing now, but a one-line grep hint (grep -rn "action IN ('created'") in the events.go block could save a future editor a search. Take it or leave it.

jsos3-cms:
@danielbowne shipped. Both commits merged in #529, the backend code-freeze batch, as `96bf66a`, authorship and signatures intact. #445 closed via the keyword declared on the batch. Prod deploy confirmed live.

Worth noting this landed alongside #528 in the same batch, and the two touch adjacent ground: an independent review of #528 had suggested exactly this consolidation, flagging that `('created', 'updated')` was written literally in four places and that two SQL fragments disagreeing about which actions count is what produced the original bug. Your constants remove that.

Closing by hand, since the batch squash-merged and GitHub cannot match this PR's content.

Issue #376, opened 2026-06-30
API login canary (CloudWatch Synthetics) for the auth/lookup -> /login/entra path


jsos3-cms:
Add a CloudWatch Synthetics API canary (Terraform, infrastructure/) that continuously exercises the unauthenticated login entry path so a regression alerts instead of going unnoticed.

Each run:
- GET /api/v1/auth/lookup?email=<synthetic, non-existent address> -> expect 200 and the {data:{idp}} envelope the UI depends on.
- GET /login/entra -> expect a 302 whose Location points at login.microsoftonline.com.

Wiring and decisions:
- Failure alarm routed to the existing ztmf-alarms SNS topic.
- Single canary; run frequency configurable via a tfvar (default every 20 minutes, ~$4/mo).
- Needs an artifact S3 bucket and a scoped execution IAM role; confirm the current Synthetics runtime at build time (the provider examples list deprecated runtimes).

Scope and limits: catches backend, ALB, and CloudFront routing/redirect breakage and regressions that do not depend on the UI feature flag. Does NOT catch the user-not-assigned / consent class at the identity provider (rejected before any callback, so there is no AWS-side signal); that is covered by the Entra console runbook and the companion UI canary in CMS-Enterprise/ztmf-ui.

Background: follow-up from the dev login investigation that added ALB access logs, ELBAuth alarms, and SNS alert routing. Not urgent.

jsos3-cms:
Companion UI canary: https://github.com/CMS-Enterprise/ztmf-ui/issues/443

danielbowne:
Refinement, grounding the assertions in the actual backend contract.

**Endpoint contract — `GET /api/v1/auth/lookup?email=`**
- Unauthenticated (it's on the ALB `allow` path, no token), rate-limited to 5 req/s, burst 10, over a 10-minute window. A 20-minute canary is nowhere near that, but treat a `429` as a soft retry, not a hard fail.
- Always returns the `{data: {...}}` envelope. Body shape: `{"data":{"idp": <string|null>}}`.
- **Key correction for the synthetic email:** a non-existent (or soft-deleted) email returns **200 with `"idp": null`** by design — a missing user is deliberately indistinguishable from a provisioned-but-unrouted one (privacy). So the canary must **not** assert a non-null `idp` for a synthetic address; it will always be null.

**So the lookup assertion should be:**
- HTTP 200, `Content-Type: application/json`
[...]

PR #525, opened 2026-08-05
fix(monitoring): require 3-of-6 windows on the login auth-failure alarm

danielbowne:
Tunes the prod login auth-failure alarm from "any single failure in a 5-minute window" to **3 breaching windows out of 6** (30 minutes).

## Why

Five weeks of alarm history (Jul 1 - Aug 5): ~3 ALARM transitions per week, every single one a lone `ELBAuthFailure` datapoint surrounded by healthy `ELBAuthSuccess` traffic — the most recent had 12 successful logins in the same hour. That pattern is a user hitting back onto the OIDC callback or abandoning a login, not a configuration problem, and each occurrence sends an ALARM + OK email pair to the shared inbox. At data-call traffic levels this only gets louder, and the alert loses its meaning right when we need it watching the Entra rollout.

A real problem — a misconfigured IdP app, a user who cannot get in and keeps retrying — produces repeated failures. Those land 3 breaching 5-minute windows inside half an hour and still page well within one working session. Singletons never fire.

The companion `ELBAuthError` alarm (ALB-side auth breakage: IdP unreachable, bad client secret) deliberately keeps its single-window trigger: that class is rare and severe, and it fired zero times over the same five weeks.

## What changed

`evaluation_periods` 1 → 6, `datapoints_to_alarm` 3 (new). Threshold, period, statistic, missing-data handling, and both notification routes unchanged. Dev and prod both pick it up via `var.environment`.

## Deploy note

Holding as draft until the current backend batch merges — marking ready triggers the DEV apply and the lane is busy today.

### DIFF
--- a/infrastructure/monitoring-login-auth.tf
+++ b/infrastructure/monitoring-login-auth.tf
@@ -63,12 +63,22 @@ resource "aws_cloudwatch_metric_alarm" "ztmf_login_elb_auth_error" {
 
 # ELBAuthFailure: the IdP response was REJECTED by the ALB (invalid id_token,
 # issuer/audience mismatch, or missing/invalid authorization code - i.e. a
-# redirect-URI or consent problem at the IdP). >0 in a 5-min window is
-# actionable and tells Batch 2 exactly what to look for in the IdP console.
+# redirect-URI or consent problem at the IdP).
+#
+# 3-of-6 rather than a single >0 window: five weeks of history (Jul-Aug 2026)
+# showed ~3 fires/week, every one a lone failure surrounded by successful
+# logins - back-button onto the callback, canceled login - and each costing an
+# ALARM + OK email pair to the shared inbox. A real problem is never a
+# singleton: a misconfigured IdP app or a user who cannot get in produces
+# repeated failures, which lands 3 breaching 5-min windows inside 30 minutes
+# and still pages well within one working session. The companion
+# ELBAuthError alarm above deliberately stays on a single window: ALB-side
+# auth breakage is rare and severe, and it fired zero times in the same span.
 resource "aws_cloudwatch_metric_alarm" "ztmf_login_elb_auth_failure" {
   alarm_name          = "ztmf-login-elb-auth-failure-${var.environment}"
   comparison_operator = "GreaterThanThreshold"
-  evaluation_periods  = "1"
+  evaluation_periods  = "6"
+  datapoints_to_alarm = "3"
   metric_name         = "ELBAuthFailure"
   namespace           = "AWS/ApplicationELB"
   period              = "300"


jsos3-cms:
@danielbowne  I like this, and I think it makes sense. Out of an abundance of caution, can we make the window a little tighter? It's a good signal right now, while we're watching, to potentially see an increased volume during the first week. Moved this to a full PR as there's no traffic in Dev, thank you for keeping the lane clear~!

jsos3-cms (approved):
Approving. The three-of-six math is exactly right, the routing is untouched, and the apply is an in-place update rather than a replacement, so there is no window where the alarm goes missing during rollout. No tfvars override alarm parameters, so all three environments pick this up uniformly, and `terraform fmt` is clean.

I checked the assumption the tuning rests on, and it holds by documentation rather than by inference. The ALB metrics reference states that Elastic Load Balancing reports metrics only when requests are flowing through the load balancer, and that a metric with no data is not reported at all. So `notBreaching` is the correct treatment for the ordinary case where nobody has failed a login.

The split you chose is also the right one. The access-log reason-code table maps the token-endpoint and userinfo-endpoint failures and timeouts to `ELBAuthError`, which keeps the "the IdP is unreachable or misconfigured" class on the sibling alarm at one-of-one. Only the "this individual response was bad" class slows down. Given zero `ELBAuthError` fires in five weeks against roughly fifteen `ELBAuthFailure` singletons, that is the correct thing to slow down.

Three costs worth stating plainly so nobody is surprised by them later.

A short total outage under about ten minutes now produces no alert at all, because it can only touch two five-minute buckets and the alarm needs three. The ten to fifteen minute range depends on where the outage falls relative to bucket boundaries. So "still pages within one working session" is true for sustained problems, not for brief ones.

Because the metric is only published when a login actually fails, accumulating three datapoints requires three failed attempts. In a low-traffic window, a real but low-cadence outage can take considerably longer to reach three, or go unnoticed until traffic resumes.

The pre-existing `treat_missing_data = "notBreaching"` means that if the ALB stops emitting entirely, the alarm reads OK rather than alarming or going to insufficient data. This change does not cause that and no windowing would fix it, but it is worth being clear that this alarm family cannot catch a silent load balancer. That is already tracked in #376, the CloudWatch Synthetics login canary: a canary generates its own traffic, so it detects a dead load balancer by construction rather than waiting for a real user to happen to hit it. Better home for it than widening this diff.

Unrelated to this change: the failing dev deploy is the ECS `tfSTABLE` timeout caused by a task-definition replacement from a concurrent deploy on the shared dev lane. It needs a re-run once Actions is healthy again.

jsos3-cms:
Folded into #541, the backend freeze batch, so the approved PRs share one dev deploy and one prod deploy. The commit went in unchanged.

This closes no issue, so nothing needed re-declaring on the batch.

Its own DEV deploy was re-run first and came back green, including `Infrastructure / Deploy`, which had previously failed on an ECS stabilisation timeout caused by a concurrent deploy on the shared lane rather than by anything in this change.

jsos3-cms:
Landed in `main` as part of #541, squash commit 469b87e. The alarm change went in unchanged.

Closing as superseded rather than abandoned. Thanks @danielbowne.

PR #527, opened 2026-08-05
fix(#526): export the full questionnaire per system, not just answered rows

### LINKED ISSUE #526: [P0] Data call export omits systems with no score rows: FindAnswers is anchored on scores, so a never-started system contributes nothing to the xlsx
## Problem

## Proposal

answer render identically.

## Acceptance criteria

      for the same system and call.
      zero.
      and `make test-e2e` green.

## References

  produces a complete export.

voidspooks:
## What and why

The data call export was anchored on the `scores` table, so a system with no score rows for the call contributed nothing to the xlsx and no warning said so. After the HHS onboarding and FY26 inventory load that hides most of an OpDiv's inventory from the export. This re-anchors `FindAnswers` on the questionnaire: per system, every function either applicable to its environment or answered by it in the call, left joined to scores. An unanswered applicable function now exports as a blank row instead of being absent.

Closes #526. Frontend half is CMS-Enterprise/ztmf-ui#678.

## Changes

- `FindAnswers` drives from `fismasystems` and the applicable-or-answered function set, left joining scores, replacing the scores-anchored inner-join chain. Applicability is resolved through the `datacenterenvironments` scoring vocabulary with pillars taken through questions, matching the questionnaire and `/scores/progress` exactly.
- `Answer.OptionDescription`, `OptionName`, `Score`, and `Notes` become pointers, nil for an unanswered function; the spreadsheet renders nil as a blank cell, so a never-answered function is distinguishable from one genuinely answered at score 0.
- Two guards keep the file a superset of the old output: decommissioned systems appear only for functions they actually answered, and answered-but-no-longer-applicable functions (orphaned by an environment change) are retained rather than dropped.

## Behavior

A call's export now covers every participating system as a full questionnaire, blank where unanswered, rather than only systems with scores. Fully-answered, decommissioned, and env-orphaned systems keep exactly the rows they produced before. Scope (OpDiv/user) behavior is unchanged.

## Tests

- Unit: the blank-vs-real-zero rendering distinction.
- Integration: never-started inclusion with a per-system count matching `/scores/progress`, the answered-row regression, orphaned-answer retention, and the decommissioned guard, all against the seeded DB.
- Verified live against the dev stack: a never-started system returns its blank questionnaire and an env-orphaned system keeps its answers.
- `make test-unit`, `make test-integration`, `make test-e2e` green; no openapi drift.

Snyk and deploy checks show red on fork PRs (no repo secrets); those are re-run by maintainers after pulling the branch.

### DIFF
--- a/backend/cmd/api/internal/spreadsheet/spreadsheet.go
+++ b/backend/cmd/api/internal/spreadsheet/spreadsheet.go
@@ -34,10 +34,12 @@ func Excel(answers []*model.Answer) (*excelize.File, error) {
 		f.SetCellValue(sheet, fmt.Sprintf("D%d", row), a.Function)
 		f.SetCellValue(sheet, fmt.Sprintf("E%d", row), a.Description)
 		f.SetCellValue(sheet, fmt.Sprintf("F%d", row), a.Question)
-		f.SetCellValue(sheet, fmt.Sprintf("G%d", row), a.OptionDescription)
-		f.SetCellValue(sheet, fmt.Sprintf("H%d", row), a.OptionName)
-		f.SetCellValue(sheet, fmt.Sprintf("I%d", row), a.Score)
-		f.SetCellValue(sheet, fmt.Sprintf("J%d", row), a.Notes)
+		f.SetCellValue(sheet, fmt.Sprintf("G%d", row), derefString(a.OptionDescription))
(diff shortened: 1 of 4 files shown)


MackOverflow:
Moving this to #528 — fork PRs can't access the org secrets required for Snyk and the deploy CI gates, so those jobs can't run here. Your commit and authorship carry over unchanged. Thanks @voidspooks!

PR #528, opened 2026-08-05
fix(#526): export the full questionnaire per system, not just answered rows

### LINKED ISSUE #526: [P0] Data call export omits systems with no score rows: FindAnswers is anchored on scores, so a never-started system contributes nothing to the xlsx

MackOverflow:
> **Note:** In-repo copy of #527 by @voidspooks. Moved from the fork so Snyk
>
> Refs #527

---

## What and why

The data call export was anchored on the `scores` table, so a system with no score rows for the call contributed nothing to the xlsx and no warning said so. After the HHS onboarding and FY26 inventory load that hides most of an OpDiv's inventory from the export. This re-anchors `FindAnswers` on the questionnaire: per system, every function either applicable to its environment or answered by it in the call, left joined to scores. An unanswered applicable function now exports as a blank row instead of being absent.

Closes #526. Frontend half is CMS-Enterprise/ztmf-ui#678.

## Changes

- `FindAnswers` drives from `fismasystems` and the applicable-or-answered function set, left joining scores, replacing the scores-anchored inner-join chain. Applicability is resolved through the `datacenterenvironments` scoring vocabulary with pillars taken through questions, matching the questionnaire and `/scores/progress` exactly.
- `Answer.OptionDescription`, `OptionName`, `Score`, and `Notes` become pointers, nil for an unanswered function; the spreadsheet renders nil as a blank cell, so a never-answered function is distinguishable from one genuinely answered at score 0.
- Two guards keep the file a superset of the old output: decommissioned systems appear only for functions they actually answered, and answered-but-no-longer-applicable functions (orphaned by an environment change) are retained rather than dropped.

## Behavior

A call's export now covers every participating system as a full questionnaire, blank where unanswered, rather than only systems with scores. Fully-answered, decommissioned, and env-orphaned systems keep exactly the rows they produced before. Scope (OpDiv/user) behavior is unchanged.

## Tests

- Unit: the blank-vs-real-zero rendering distinction.
- Integration: never-started inclusion with a per-system count matching `/scores/progress`, the answered-row regression, orphaned-answer retention, and the decommissioned guard, all against the seeded DB.
- Verified live against the dev stack: a never-started system returns its blank questionnaire and an env-orphaned system keeps its answers.
- `make test-unit`, `make test-integration`, `make test-e2e` green; no openapi drift.

Snyk and deploy checks show red on fork PRs (no repo secrets); those are re-run by maintainers after pulling the branch.

### DIFF
--- a/backend/cmd/api/internal/spreadsheet/spreadsheet.go
+++ b/backend/cmd/api/internal/spreadsheet/spreadsheet.go
@@ -34,10 +34,12 @@ func Excel(answers []*model.Answer) (*excelize.File, error) {
 		f.SetCellValue(sheet, fmt.Sprintf("D%d", row), a.Function)
 		f.SetCellValue(sheet, fmt.Sprintf("E%d", row), a.Description)
 		f.SetCellValue(sheet, fmt.Sprintf("F%d", row), a.Question)
-		f.SetCellValue(sheet, fmt.Sprintf("G%d", row), a.OptionDescription)
-		f.SetCellValue(sheet, fmt.Sprintf("H%d", row), a.OptionName)
-		f.SetCellValue(sheet, fmt.Sprintf("I%d", row), a.Score)
-		f.SetCellValue(sheet, fmt.Sprintf("J%d", row), a.Notes)
+		f.SetCellValue(sheet, fmt.Sprintf("G%d", row), derefString(a.OptionDescription))
+		f.SetCellValue(sheet, fmt.Sprintf("H%d", row), derefString(a.OptionName))
+		if a.Score != nil {
+			f.SetCellValue(sheet, fmt.Sprintf("I%d", row), *a.Score)
+		}
+		f.SetCellValue(sheet, fmt.Sprintf("J%d", row), derefString(a.Notes))
 		// NULL target = no explicit assertion yet; the app-wide default is
 		// Advanced, and the export says so rather than leaving readers to guess.
 		targetTier := "Advanced (default)"
@@ -53,4 +55,13 @@ func Excel(answers []*model.Answer) (*excelize.File, error) {
 	}
 
 	return f, nil
+}
+
+// derefString returns the pointed-to string, or "" when the pointer is nil, so a
+// not-started answer renders as a blank cell rather than a zero value.
+func derefString(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
 }
--- a/backend/cmd/api/internal/spreadsheet/spreadsheet_test.go
+++ b/backend/cmd/api/internal/spreadsheet/spreadsheet_test.go
@@ -1,0 +1,67 @@
+package spreadsheet
+
+import (
+	"testing"
+
+	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func strptr(s string) *string { return &s }
+func intptr(i int) *int       { return &i }
+
+// TestExcelRendersNotStartedBlankVsRealZero pins the #526 distinction in the
+// output cells: a function the system never answered (nil option/score/notes)
+// renders as blank, while a function genuinely answered at score 0 renders "0".
+// The two must read differently, which is the whole point of carrying the answer
+// fields as nullable rather than coalescing an absent answer to zero.
+func TestExcelRendersNotStartedBlankVsRealZero(t *testing.T) {
+	answers := []*model.Answer{
+		{ // row 2: applicable function, never answered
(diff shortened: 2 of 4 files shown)


jsos3-cms (approved):
Approving. Reviewed alongside CMS-Enterprise/ztmf-ui#679, since the two only make sense together.

## Verified rather than read

The superset claim holds structurally, and both guards are genuinely exercised rather than skipped: the orphaned-answer and decommissioned-with-scores candidates exist in the seed, so the conditional assertions in `answers_integration_test.go` execute rather than falling through their log branches. Full integration suite green, including the three tests that are sometimes date-dependent on main.

Applicability now matches in all three places, compared join-chain by join-chain: the questionnaire (`FindQuestionsByFismaSystem`), `/scores/progress` (the `expected` CTE), and this export all resolve through `dce.scoring_key = functions.datacenterenvironment` and then `pillars.pillarid = questions.pillarid`. That is the part worth the most confidence and it checks out.

Nil handling is clean. `FindAnswers` is the only producer of `model.Answer`, and its only consumers are `Excel()` and the pass-through controller. `Excel()` derefs through a nil-safe helper and only sets the score cell when non-nil, so an unanswered function is a genuinely blank cell rather than `0`. The unit test pins that distinction.

## One consequence of the pillar change worth checking against prod

The description says pillars are now taken through questions, which is the correct alignment. What it does not say is what that changes for rows that already exported. Before this, the export resolved pillar from `functions.pillarid`; now it comes from the function's question. Those are two independent FKs to `pillars` (`0001initialschema.go:18` and `:52`), and `Save()` sets `functions.pillarid` from caller input rather than deriving it, so nothing has ever forced them to agree.

Where they disagree, the Pillar column changes value on rows that previously exported. That is outside what "superset" covers, since no row disappears, and it would be a silent change in a compliance artifact.

Cheap to settle before anyone relies on the file:

```sql
SELECT f.functionid, f.function, f.pillarid AS function_pillar, q.pillarid AS question_pillar
FROM functions f
JOIN questions q ON q.questionid = f.questionid
WHERE f.pillarid <> q.pillarid;
```

Empty means no exported value changes and this is purely additive. Non-empty means those functions' Pillar column shifts, and the new value is the right one, but it is worth knowing rather than discovering from a diffed spreadsheet. Not blocking, since the fix direction is correct either way.

## Non-blocking: one missing index

The `scores` LEFT JOIN's `functionoptionid IN (SELECT ... WHERE selected.functionid = functions.functionid)` cannot be folded into the hash join, so Postgres re-evaluates it per candidate row and sequentially scans `functionoptions` each time. `EXPLAIN (ANALYZE, BUFFERS)` against the seeded DB confirms the plan shape, at `loops=1188` on 31 systems.

Cost now scales with systems times applicable-functions times answered-functions rather than with total scores rows, which is the real change from the old query. On the seed this is nothing; extrapolated to hundreds of systems it is low-single-digit-millions of subplan evaluations per export. That is an extrapolation from a confirmed plan shape rather than a measurement, so treat the wall-clock estimate as soft.

`CREATE INDEX ON functionoptions(functionid)` removes it. Small static catalog table, safe, and it turns the per-row scan into a lookup. Worth doing, not worth holding this.

## Merge order

Land this first, or same-day with #679. The asymmetry matters:

This PR alone is fully safe. The backend starts returning full rows, and the pre-#679 frontend still gates selection on the scores map, so nothing is exposed yet.

#679 alone opens a narrow window where a never-started system's checkbox becomes selectable and the export button clickable, while the backend is still scores-anchored, so the download silently produces a header-only file. Its author already discloses that in an unchecked acceptance criterion, so it is known and accepted rather than hidden, but it argues against letting #679 sit alone for long.

jsos3-cms:
Shipped. Merged in #529, the backend code-freeze batch, as `96bf66a`, with @voidspooks' authorship preserved on the commit. #526 closed via the keyword declared on the batch. Prod deploy confirmed live at 16:36 Pacific, ahead of the frontend half.

Thanks @MackOverflow for rehoming it so the Snyk and deploy gates could actually run.

Two things from review are recorded on this PR and worth carrying forward rather than losing here. The fix reaches further than the ticket described: with no action filter at all, a row with a genuine in-app edit could have its timestamp overwritten by a later import, which was not bounded to closed cycles the way the reported symptom was. And the pillar source moved from `functions.pillarid` to the function's question, which is the correct alignment, but means the Pillar column changes value on any row where those two FKs had drifted. The drift query is in my review above and is still worth running against prod before the file is relied on.

The suggested `functionoptions(functionid)` index also remains outstanding, unticketed.

Closing by hand, since the batch squash-merged.

PR #529, opened 2026-08-05
Code freeze batch 2 (backend): last_seen, event-action constants, full-questionnaire export

### LINKED ISSUE #445: Tech-debt follow-ups from #437: shared event-action constant + scores.Save status seam

### LINKED ISSUE #526: [P0] Data call export omits systems with no score rows: FindAnswers is anchored on scores, so a never-started system contributes nothing to the xlsx

jsos3-cms:
Backend release train for the code freeze. Approved PRs only, cherry-picked so each contributor's authorship and signature survive. Frontend counterpart is CMS-Enterprise/ztmf-ui#682.

Closes #445
Closes #526

## What is in it

| Source | Author | Change | Closes |
| --- | --- | --- | --- |
| #522 | @danielbowne | Expose `last_seen` on the users list, plus a login event so it means last sign-in | — |
| #524 | @danielbowne | Single home for event-action and score-status values | #445 |
| #528 | @MackOverflow, commits by @voidspooks | Export the full questionnaire per system rather than only answered rows | #526 |

#522 declares no closing issue by design: it is the backend piece, and presentation is deliberately left open on CMS-Enterprise/ztmf-ui#675.

Ordering is by approval. #522 and #524 both touch `events.go` and `users.go`, and they merged without conflict because the edits sit in disjoint regions. #528 is disjoint from both.

One thing dropped on purpose: #522's tip was an empty `chore(ci): retrigger DEV deploy under a fresh image tag` commit, pushed to force a fresh image tag past the immutable ECR repo. It has no meaning in a batch that gets its own SHA, so it was skipped rather than carried.

## Verification on the combined branch

Verified the combination rather than trusting the parts, since three separately green PRs can still interact:

- `go build ./...` and `go vet ./...` clean
- `go test -short ./...` — 13 packages ok
- Full integration suite against a seeded database — 13 packages ok, zero failures, including the three tests that are sometimes date-dependent
- **`make generate-openapi` produces no drift**: the committed spec matches the generated one byte-for-byte. Worth doing here specifically because #522 regenerated the spec and #524 changed models independently, so the combination is the first time those two meet.

## Deploy note, and it is a real one

**Do not stack deploys on this branch.** #528's own SHA has two dev deployments two minutes apart: the first succeeded, the second failed with `waiting for ECS Service update: timeout while waiting for state to become 'tfSTABLE' (last state: 'tfPENDING', timeout: 20m0s)`. The second run raced the first while ECS was still rolling out. #524 lost a deploy the same way earlier in the day.

So this PR was opened ready rather than draft-then-ready, because that transition is what produced the double run. If the deploy here fails on a stabilise timeout, let ECS settle and re-run **one** deploy; do not push again or trigger a second in parallel.

## Merge order

**This merges first.** The frontend batch follows at least five minutes later, which is both a courtesy to the prod pipeline and a correctness requirement: #528's export change has to be live before the frontend's select-all makes never-started systems selectable, or a user can select them and download a header-only file.

## What is not in it

- #525 and #492 are drafts.
- #423 is a dependency bump with no approval.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1029,6 +1029,12 @@ SELECT '00000000-0000-4000-8000-000000000001', 'imported', 'public.scores',
 -- FY2020 archives block) are excluded via the same action filter the migration
 -- uses, so imported history stays not_started. Keeps this consistent
 -- automatically as the events seed above changes.
+--
+-- The action values here (and in the 'imported' events seeded above) are SQL
+-- literals because SQL cannot reference a Go constant; they are the same values
+-- defined as eventAction* in internal/model/events.go, which is the source of
+-- truth for the set. Changing an action value there means updating these
+-- literals and backfilling stored rows, not a rename.
 UPDATE public.scores s
 SET status = 'done'
 WHERE EXISTS (
--- a/backend/cmd/api/internal/auth/session.go
+++ b/backend/cmd/api/internal/auth/session.go
@@ -176,6 +176,15 @@ func SessionHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	SetSessionCookie(w, token)
+
+	// Record the login. Fire-and-forget by design: the session is already
+	// minted and the cookie already set, so failing the redirect over an audit
+	// write would deny a user access for a bookkeeping problem. Logged, not
+	// returned - the same trade recordEvent makes for write-derived events.
+	if err := model.RecordLogin(r.Context(), user.UserID); err != nil {
+		log.Printf("login: failed to record session event for %s: %s\n", user.UserID, err)
+	}
+
 	http.Redirect(w, r, "/", http.StatusFound)
 }
 
--- a/backend/cmd/api/internal/auth/session.go
+++ b/backend/cmd/api/internal/auth/session.go
@@ -177,10 +177,14 @@ func SessionHandler(w http.ResponseWriter, r *http.Request) {
 
 	SetSessionCookie(w, token)
 
-	// Record the login. Fire-and-forget by design: the session is already
-	// minted and the cookie already set, so failing the redirect over an audit
-	// write would deny a user access for a bookkeeping problem. Logged, not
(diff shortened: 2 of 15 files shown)


danielbowne (approved):
LGTM

Issue #530, opened 2026-08-05
Confirm the export's Pillar column did not shift for existing rows after #528


jsos3-cms:
## What to check

One query against prod, then either fix or close this. It exists because a change that shipped today could have moved a value in the data call export, and nobody has confirmed either way.

```sql
SELECT f.functionid, f.function, f.pillarid AS function_pillar, q.pillarid AS question_pillar
FROM functions f
JOIN questions q ON q.questionid = f.questionid
WHERE f.pillarid <> q.pillarid;
```

**Empty result:** nothing moved, this is purely additive, close it.

**Rows returned:** those functions' Pillar column changed value in the export as of today. Decide whether that needs communicating to whoever consumes the file, and whether `functions.pillarid` should be reconciled to its question's or dropped as a duplicate source.

## Why

#528 re-anchored `FindAnswers` and, as part of aligning applicability with the questionnaire and `/scores/progress`, moved the Pillar column's source from `functions.pillarid` to the pillar reached through the function's question. That alignment is correct: all three now resolve identically, which they did not before.

But those are two independent FKs to `pillars` (`0001initialschema.go:18` and `:52`), and `functions.Save()` sets its copy from caller input rather than deriving it, so nothing has ever forced them to agree. Wherever they drifted, a row that exported last cycle now exports with a different Pillar.

That is outside what the PR's superset guarantee covers. No row disappears, so the tests and the review both pass while a column value silently changes in a compliance artifact. The export is the data call deliverable, so an OpDiv comparing this cycle's file against last cycle's would see the change with no explanation available.

No evidence of drift was found in the schema or seed data during review, so the likely outcome is an empty result. The point is that likely is not the same as checked, and checking costs one query.

## Not in scope

The `functionoptions(functionid)` index also came out of that review. Separate concern, and the right first step there is measuring one real export at prod scale rather than adding the index on an extrapolation.

danielbowne:
Checked against live production this morning, using the exact query from the description:

```
 functionid | function | function_pillar | question_pillar
------------+----------+-----------------+-----------------
(0 rows)

total=360 agreeing=360
```

Empty result - all 360 functions carry the same pillarid as their question, so re-anchoring the export's Pillar column in #528 changed no existing row's value. Also confirmed on yesterday evening's snapshot copy with the same result, so this was true both before and after the FY26 activity in between.

[...]

PR #537, opened 2026-08-06
chore(functions): derive a function's pillar from its question (#534)

### LINKED ISSUE #534: Single source of truth for a function's pillar
### Type

Chore

### Description

`functions.pillarid` and `questions.pillarid` store the same fact independently - two separate FKs to `pillars`, and nothing forces them to agree. `functions.Save()` writes `pillarid` from caller input rather than deriving it from the function's question, so an admin edit or API call can create a disagreement at any time.

Today nothing user-facing reads the function's copy: #528 re-anchored the export's Pillar column through the question, matching the questionnaire and `/scores/progress`. #530 verified zero drift exists in production (360/360 agree). So this is latent, not live - but the stale copy sits there waiting to mislead the next query that reads it directly.

Options, in order of preference:

- Derive `pillarid` from the question inside `Save()` and ignore caller input (small, no migration)
- Drop `functions.pillarid` entirely and read through the question everywhere (cleaner, needs a migration plus an audit of remaining readers)

Refs #528, #530.

costacalvin:
Closes #534.

## Problem

`functions.pillarid` and `questions.pillarid` store the same fact independently — two separate FKs to `pillars`, with nothing forcing them to agree. `Function.Save()` wrote the function's copy straight from caller input rather than deriving it from the function's question, so an admin edit or a direct API call could create a disagreement at any time.

Nothing user-facing reads the function's copy today: #528 re-anchored the export's Pillar column through the question, matching the questionnaire and `/scores/progress`, and #530 verified zero drift in production (360/360 agree). So this was latent rather than live — but the stale copy sat there waiting to mislead the next query that read it directly.

## Approach

Option 1 from the issue: close the write path rather than drop the column.

`Save()` now resolves the function's question and writes that question's `pillarid`, discarding whatever the caller sent. `questionid` becomes required — a function with no question reaches no questionnaire (`FindQuestionsByFismaSystem` inner-joins `functions` on `questionid`) and has no pillar to derive from, so it is invalid input. With `pillarid` always derived, the `pillarid` check in `validate()` is dead for every valid input and comes out.

An unknown `questionid` returns `ErrNoReference`, which is the same 400 the caller already got — today that input reaches Postgres and comes back as a 23503 that `trapError` maps to `ErrNoReference` anyway — just detected before the insert.

### Requiring questionid also closes an accidental-NULL path

Worth calling out separately, because it is a second bug rather than a consequence of the first. The UPDATE in `Save()` sets every column unconditionally, so pre-PR a `PUT /functions/{id}` that omitted `questionid` — an ordinary partial-update shape — wrote NULL to the column and silently orphaned the function from every questionnaire, since `FindQuestionsByFismaSystem` inner-joins `functions` on `questionid`. The function would simply stop appearing for every system in its data center environment, with no error to the caller. Requiring `questionid` makes that request a 400 instead.

**No migration.** #530 showed production already agrees across all 360 functions, and writes are now self-correcting.

### Scope of the guarantee

This closes the `/functions` write path only: a caller-supplied `pillarid` can no longer put a function out of agreement with its question. It does **not** make the invariant total. `PUT /api/v1/questions/{id}` writes `questions.pillarid` with no cascade to dependent functions, so moving a question to a different pillar leaves its functions stale until each is next saved (at which point this change re-derives and self-heals it). That path is pre-existing and untouched here — see the follow-up below.

## Acceptance criteria

- [x] `POST`/`PUT /api/v1/functions` writes the question's `pillarid` regardless of the `pillarid` in the request body, on both insert and update.
- [x] A request whose `pillarid` disagrees with its question returns 201/200 with the corrected value, rather than persisting the disagreement.
- [x] `questionid` is required; a request without one returns 400 with `questionid` in the error payload, and nothing is written.
- [x] An unknown `questionid` returns 400 (`ErrNoReference`) and nothing is written.
- [x] A `PUT` omitting `questionid` returns 400 rather than NULLing the column and orphaning the function from every questionnaire.
- [x] No schema change and no data migration.
- [x] Existing readers are unaffected: the questionnaire, `/scores/progress`, the scoring aggregate and the export all already join `questions.pillarid`.

## Testing

- `internal/model/functions_pillarderive_integration_test.go` (new) — insert and update each correct a deliberately wrong `pillarid`; missing `questionid` → `InvalidInputError` carrying `questionid`; unknown `questionid` → `ErrNoReference`. Fixtures use the `AWS` environment, which no seeded system maps to, so they are never visible to a questionnaire while the test runs.
- `emberfall_tests.yml` — `functionData` / `updatedFunctionData` gained `questionid: 8001`; added a drift case (claims pillar 6 against a pillar-1 question → 201 with `pillarid: 1`), a questionless `POST` → 400, and a questionless `PUT` → 400 followed by a `GET` asserting the whole prior state, so a partial write of any field fails the case.
- `make test-unit` — pass
- `make test-integration` — pass, including all 4 new subtests
- `make test-e2e` — 201 ran, 0 failed
- `openapi.yaml` regenerated: `pillarid` is now `readOnly: true` with a description. Redocly lint is unchanged at 8 warnings / 2 ignored, verified against the baseline.

## Reviewer notes

- **Behavior change:** `POST`/`PUT /api/v1/functions` without a `questionid` now returns 400 where it previously succeeded. The frontend never writes to this endpoint — its only `functions` calls are `GET functions/{id}/options` in `QuestionnairePage` and `QuestionnareModal`, and there is no admin "manage functions" view — so the only affected callers would be hand-rolled scripts against the admin API.
- `Save()` repeats the nil check on `QuestionID` immediately before dereferencing it. `validate()` already rejects nil, so the branch is unreachable today; it is there so the deref does not silently depend on a guard sixteen lines up in another method.
- Worth confirming before merge that production has no `functions` rows with a NULL `questionid` (`SELECT count(*) FROM functions WHERE questionid IS NULL;`). Such rows are already orphaned from every questionnaire, but they would no longer be editable through the API without supplying a question. #530's 360/360 count only covered rows that join to a question, so it would not have surfaced them.

## Follow-up (post data call)

Deferring the remaining work to a single change after the current data call, rather than splitting it across cycles:

- Cascade or eliminate the question-edit path described above, so a question moving pillars cannot leave its functions stale.
- Option 2 from #534: drop `functions.pillarid` entirely and read through the question everywhere. This subsumes the cascade — with no second copy there is nothing to keep in sync. It needs `functions.questionid` made NOT NULL, plus updates to the `PillarID` filter in `FindFunctions`, the Empire seed's score generator, and the OpenAPI schema.

Doing both at once avoids writing a cascade that the column drop would immediately delete. Not scheduled during the data call because it touches the questionnaire's schema while systems are actively being scored.

Refs #528, #530.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -1268,6 +1268,37 @@ tests:
     expect:
       status: 400
 
+  # Same on update. The UPDATE sets every column, so before questionid was
+  # required this partial-update shape wrote NULL and orphaned the function from
+  # every questionnaire instead of failing.
+  - url: "http://localhost:8080/api/v1/functions/{{.createFunction.Response.data.functionid}}"
+    method: PUT
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        function: "Updated Identity Management"
+        description: "no questionid on update, must be rejected"
+        datacenterenvironment: "AWS"
+        order: 2
+        pillarid: 1
+    expect:
+      status: 400
+
+  # Nothing from the rejected update applied: the question is intact and the
+  # fields it did carry (a changed description) were not written either.
+  - url: "http://localhost:8080/api/v1/functions/{{.createFunction.Response.data.functionid}}"
+    method: GET
+    headers:
+      <<: *commonHeaders
+    expect:
+      status: 200
+      body:
+        json:
+          data:
+            <<: *updatedFunctionData
+
   # Function Options Endpoints
   - url: "http://localhost:8080/api/v1/functions/{{.createFunction.Response.data.functionid}}/options"
     method: GET
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -107,11 +107,14 @@ reactivatedFismaSystemData: &reactivatedFismaSystemData
   decommissioned_notes: "System migrated to cloud infrastructure"
   reactivation_notes: "Brought back online for follow-on work"
 
+# questionid is required, and pillarid is derived from that question rather than
+# taken from the request. Seed question 8001 is in pillar 1.
 functionData: &functionData
   function: "Identity Management"
   description: "Manage user identities and access"
   datacenterenvironment: "AWS"
   order: 1
+  questionid: 8001
   pillarid: 1
 
 updatedFunctionData: &updatedFunctionData
@@ -119,6 +122,7 @@ updatedFunctionData: &updatedFunctionData
   description: "Updated identity and access management"
   datacenterenvironment: "AWS"
   order: 2
+  questionid: 8001
   pillarid: 1
 
 userData: &userData
@@ -1222,7 +1226,48 @@ tests:
       status: 200
       headers:
         content-type: "application/json"
-  
+
+  # A caller-supplied pillarid that disagrees with the question is corrected, not
+  # persisted. Question 8001 is in pillar 1; the request claims pillar 6.
+  - id: createFunctionPillarDrift
+    url: http://localhost:8080/api/v1/functions
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        function: "Pillar Derivation Check"
+        description: "pillarid must follow the question, not the request body"
+        datacenterenvironment: "AWS"
+        order: 99
+        questionid: 8001
+        pillarid: 6
+    expect:
+      status: 201
+      body:
+        json:
+          data:
+            questionid: 8001
+            pillarid: 1
+
+  # A function with no question has no pillar to derive from and reaches no
+  # questionnaire, so it is rejected.
+  - url: http://localhost:8080/api/v1/functions
+    method: POST
+    headers:
+      <<: *commonHeaders
+      content-type: "application/json"
+    body:
+      json:
+        function: "Questionless Function"
+        description: "no questionid, must be rejected"
+        datacenterenvironment: "AWS"
+        order: 99
+        pillarid: 1
+    expect:
+      status: 400
+
   # Function Options Endpoints
   - url: "http://localhost:8080/api/v1/functions/{{.createFunction.Response.data.functionid}}/options"
     method: GET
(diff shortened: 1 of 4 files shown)


MackOverflow (dismissed):
Approving. Clean fix for a real latent integrity bug, and the write-only scope is documented honestly rather than papered over.

Traced the full path and the acceptance criteria hold:

Unknown questionid → ErrNoData → ErrNoReference → 400, same status the FK violation produced, just caught before the insert.
Missing/zero questionid rejected in validate() before any DB call, with questionid in the 400 payload.
Pillar derivation sits above the insert/update branch, so it applies to both.
One thing I'd add to the description: making questionid required also closes an accidental-null path. Pre-PR, a PUT omitting questionid would have set the column to NULL and orphaned the function from every questionnaire — this fixes that too, not just the pillar drift.

Two non-blocking follow-ups, whenever convenient:

pillarid is now an ignored input on write but the spec still advertises it as settable — worth marking read-only so a hand-rolled admin script doesn't think it does anything.
The e2e covers a questionless POST → 400 but not the equivalent PUT. Shares validate(), so low-risk, but it's the case most likely to affect an existing caller.
Only real gate before merge is the reviewer note you already flagged: SELECT count(*) FROM functions WHERE questionid IS NULL;. #530's 360/360 can't see NULL-questionid orphans since it only counted rows that join to a question, and those rows become un-editable via the API after this lands. If it comes back 0, ship it.

costacalvin:
Thanks @MackOverflow — folded both non-blocking follow-ups into this PR rather than deferring them as this will wait to be merged until later tonight. Pushed as 5f5b5e6, so the approval was dismissed and need a re-approval.

1. Accidental-NULL path added to the description as its own section, and to the AC list. Agreed it's a second bug, not a consequence of the first.
2. `pillarid` marked read-only — readonly:"true" on the field, spec regenerated to readOnly: true with a description. No behavior change: nothing reads the spec at runtime and the FE has no codegen, so a request still sending pillarid is accepted and ignored as before.
3. Questionless PUT → 400, plus a GET asserting the whole prior state via the updatedFunctionData anchor, so a partial write of any field fails the case. E2E 201 ran / 0 failed.

MackOverflow (approved):
C L E A N

Issue #482, opened 2026-07-27
Close unit-test gaps: internal/secrets and the API router


voidspooks:
## Problem

Two packages with zero test files sit on riskier paths than the other untested ones.

`internal/secrets` (167 lines) backs database credential fetch and refresh-on-rotation, including the `refreshLocked`/`once.Do` synchronization that the 28P01 auth-failure retry depends on. That concurrency is exactly the kind of code that regresses silently under a refactor.

`cmd/api/internal/router/router.go` (137 lines) registers every route, verb, and numeric regex constraint. A mis-registered method or a dropped `[0-9]+` constraint fails at runtime with nothing to catch it before the e2e suite, and the regex constraints are load-bearing (they are currently what keeps the unchecked `fmt.Sscan` parsing in #477 latent rather than live).

## Proposal

For secrets, a mocked Secrets Manager client exercising fetch, cache, and forced-refresh semantics. For the router, a table-driven test asserting method + path pattern + handler wiring for every registered route, so route changes show up in a diffable table. Secrets is the priority of the two.

Surfaced in a tech-debt survey ahead of the post-release push.

PR #538, opened 2026-08-06
techdebt(#484): move DB secret resolution from db/conn.go into config

### LINKED ISSUE #484: Move DB secret resolution from db/conn.go into config (existing TODO)
## Problem

## Proposal

voidspooks:
Resolves the standing TODO in `internal/db/conn.go` by centralizing DB secret resolution in `internal/config`, as #484 proposes. Config now owns fetching and caching the Secrets Manager secret through a `DbCreds()` resolver and a `RefreshDbCreds()` refresh path, and `db` consumes resolved credentials instead of running its own `secrets.NewSecret` and `once.Do` initialization. The env-var fallback, the unconditional once.Do happens-before discipline, and the 28P01 refresh-and-retry behavior are verbatim ports; no behavior change intended. The secret cache lives on the config struct rather than at package level so resolution is testable on a constructed config, and two unit tests cover the AWS-free paths (env fallback, refresh with no secret configured). The secret-backed path remains #482's coverage territory.

Closes #484

Local gate: unit, integration, and emberfall e2e (197 passed, 0 failed) all green. Snyk and the deploy checks will show red as usual for a fork PR; maintainers re-run them after pulling the branch.

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -26,6 +26,13 @@ type smtp struct {
 	CertIntermediateSecretID *string `env:"SMTP_CA_INT_SECRET_ID"`
 }
 
+// DbCreds holds resolved database credentials, either read directly from env
+// vars or unmarshalled from the Secrets Manager secret named by DB_SECRET_ID.
+type DbCreds struct {
+	Username string
+	Password string
+}
+
 // config is shared by all binaries with values derived from environment variables
 type config struct {
 	Env      string `env:"ENVIRONMENT" envDefault:"production"`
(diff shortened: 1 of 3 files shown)


jsos3-cms:
Update on where this went. The change is in-repo now and heading for main, just by way of two hops rather than one.

It was rehomed to #540 so the CI gates that need org-level secrets could run, which they cannot on a fork PR. That worked: Snyk and the Snyk code and IaC scans all passed there, and none of them had ever run here. The only run this PR attempted sat at `action_required`, the outside-contributor approval gate, so it never got coverage at all.

#540 has since been folded into #541, the backend freeze batch, so the batch is what will actually merge. Your commit is in it, unchanged and still authored by you, and `Closes #484` is re-declared there so this work still closes its issue.

Keeping this open until the batch merges so nothing is stranded if the batch needs rework, then closing it. Thanks @voidspooks.

jsos3-cms:
This is in `main` now, as of squash commit 469b87e.

The path was fork to in-repo rehome (#540) so the org-secret CI gates could run, then folded into the backend freeze batch (#541) so it shared one dev and one prod deploy with the other approved work. `Closes #484` was re-declared on the batch and fired, so ztmf#484 is closed.

Your commit went in unchanged and stayed authored by you throughout. It also picked up a second independent review on the batch and a human approval before merge.

Closing now that it has landed. Thanks @voidspooks.

PR #539, opened 2026-08-06
Fix questionnaire ordering end to end: answer choices, question sequence, export, diff (+ export index)

### LINKED ISSUE #531: Measure a prod-scale data-call export after #528, then decide on a functionoptions(functionid) index

danielbowne:
Closes #531. Fixes the full ordering family from ztmf-misc#279 (with #276/#277/#278) and supersedes #535, whose index migration rides here.

## The problem, in one paragraph

Nothing in the database or the API ever specified the questionnaire's order. Answer choices had no ORDER BY at all, so they rendered in Postgres heap order — which matched maturity order for years by accident of insertion, until the Aug 2 description cleanup relocated the updated rows and scrambled the choices on ~50 visible questions across 7 editions, 32 minutes before the FY26 call opened. Question and pillar sequence had ORDER BYs pointing at `ordr` columns that have been zero since they were created, so the xlsx export and the score diff returned questions in plan-dependent order; only the questionnaire UI looked right, because the frontend re-sorts against its own hardcoded map.

## What this PR does

**Answer choices** (`FindFunctionOptions`): `ORDER BY score, functionoptionid` — the CISA ZTMM stage sequence (traditional, initial, advanced, optimal). The integration test reproduces the production trigger rather than trusting fixture luck: it UPDATEs an option's description (physically relocating the row) and asserts the order survives.

**Question and pillar sequence** (migration 0056): populates `pillars.ordr` (1-6, CISA pillar order) and `questions.ordr` (`pillar_rank*100 + function_index`) by transcribing the frontend's own `PILLAR_ORDER` / `PILLAR_FUNCTION_MAP` — the order the tool has rendered in production for years, so this is a lift-and-shift of an already-approved sequence, not an editorial decision. Guarded `WHERE ordr = 0`: the legacy Future Plan questions keep their 901-907 band, and the empire seed's fictional function names stay unranked (test databases fall to the tiebreaker; emberfall order is unchanged, verified 201/201). Verified against a prod copy before writing: 40/40 questions rank, zero name mismatches, zero duplicates.

**Deterministic tiebreakers**: `FindAnswers`, `FindQuestionsByFismaSystem`, and the score diff all gain `questionid` at the end of their ORDER BY, so order can never again depend on physical layout even where `ordr` ties. The score diff also gains the pillars join its ordering silently lacked — it ordered by `functions.ordr`, which is zero on all 432 rows.

**The export index** (migration 0055, from #535): `functionoptions(functionid)`. Measured on the prod-scale copy: whole-call export 47.7s → 2.0s (the LEFT JOIN's per-row subquery ran 1.55M sequential scans). Shipping it in the same PR as the choices ORDER BY is deliberate — the index flips that query's plan, which would have reshuffled the visible choice order a third time had they landed separately.

**Rider**: `RecordLogin`'s action literal converts to `eventActionCreated`, the one-line convergence #524 deferred to whichever change landed after it.

## Verified end to end on a current prod copy

- Answer choices: `traditional, initial, advanced, optimal` via the live API on the reported question, and confirmed rendered correctly in the UI
- Questions API: `Identity → Devices → Networks → Applications → Data → CrossCutting`, contiguous, exact map sequence within each pillar — from the raw API with no frontend sort
- Export xlsx: same sequence in the actual spreadsheet rows, 0.4s for a single-system export
- Score diff: a 40-row diff returns in exact questionnaire order
- Stored answers untouched throughout: answers key by `functionoptionid`, and ascending id agrees with ascending score on all 432 functions — presentation order only, no re-assessment needed

## Testing

- `make test-unit`, `make test-integration`, `make test-e2e` (201/201) all green
- Migration 0056 has its own integration test executing the registry SQL against fixture rows (the migration is deliberately a no-op on empire-seeded test databases, so fixture-based coverage is the honest kind)
- `make generate-openapi` — zero diff; response shapes unchanged, no emberfall edits needed

## Deploy notes

Two new migrations (0055, 0056). Marking this ready auto-deploys to DEV and puts `dbversions` at 56 while `main` is at 54 — any other PR's deploy in between crash-loops the lane. Ready it only when it can merge promptly. Local/test databases keep unranked `ordr` by design; dev/prod get the real ordering because their data carries the real function names.

### DIFF
--- a/backend/cmd/api/internal/migrations/0055functionoptionsindex.go
+++ b/backend/cmd/api/internal/migrations/0055functionoptionsindex.go
@@ -1,0 +1,27 @@
+package migrations
+
+func init() {
+	appendMigration(
+		"add functionoptions(functionid) index for the data-call export",
+		`
+-- The data-call export (#528 re-anchor of FindAnswers) left-joins scores with
+-- a per-row subquery over functionoptions keyed on functionid. Postgres cannot
+-- fold that IN-subquery into the hash join, so it runs as a per-candidate-row
+-- Join Filter - and functionoptions has no functionid index (only its primary
+-- key), making each evaluation a sequential scan.
+--
+-- Measured on a prod-scale copy (ztmf#531, 1,530 systems / 192k scores): the
+-- whole-call FY26 export performs that scan 1,556,120 times, 47.7s of database
+-- time before the spreadsheet is even rendered - retry/abandon territory, and
+-- near typical gateway idle timeouts. With this index the same export runs in
+-- 2.0s (23x).
+--
+-- functionoptions is a small, static, read-mostly catalogue (1,728 rows; it
+-- changes only when a questionnaire edition is edited), so the write cost is
+-- negligible and no partial or composite form is warranted.
+CREATE INDEX IF NOT EXISTS functionoptions_functionid_idx
+    ON public.functionoptions (functionid);
+`,
+		`DROP INDEX IF EXISTS functionoptions_functionid_idx;`,
+	)
+}
--- a/backend/cmd/api/internal/migrations/0056orderingdata.go
+++ b/backend/cmd/api/internal/migrations/0056orderingdata.go
@@ -68,9 +68,18 @@ FROM (VALUES
 ) AS v(pillar, ordr)
 WHERE p.pillar = v.pillar;
 
+-- Deterministic by construction rather than by hope: a question is ranked by
+-- the MINIMUM matching rank among its functions' names. On prod data every
+-- question matches exactly one name (verified: 40/40 single-name), so min() is
+-- the identity; if a future edition ever forked a question's function name,
+-- min() still yields one stable, re-runnable answer instead of letting the
+-- planner pick a VALUES row arbitrarily.
 UPDATE questions q
-SET ordr = r.ordr
-FROM (VALUES
+SET ordr = ranked.ordr
+FROM (
+  SELECT f.questionid, min(r.ordr) AS ordr
+  FROM functions f
+  JOIN (VALUES
     ('Authentication-Users',                101),
     ('IdentityStores-Users',                102),
     ('RiskAssessment',                      103),
@@ -111,13 +120,12 @@ FROM (VALUES
     ('Cross-VisibilityAnalytics',           601),
     ('Cross-AutomationOrchestration',       602),
     ('Cross-Governance',                    603)
-) AS r(function_name, ordr)
-WHERE q.ordr = 0
-  AND EXISTS (
-      SELECT 1 FROM functions f
-      WHERE f.questionid = q.questionid
-        AND f.function = r.function_name
-  );
+  ) AS r(function_name, ordr) ON r.function_name = f.function
+  WHERE f.questionid IS NOT NULL
+  GROUP BY f.questionid
+) AS ranked
+WHERE q.questionid = ranked.questionid
+  AND q.ordr = 0;
 `,
 		`
 -- Reverses the population only. The legacy 901-907 band predates this migration
--- a/backend/cmd/api/internal/migrations/0056orderingdata.go
+++ b/backend/cmd/api/internal/migrations/0056orderingdata.go
@@ -1,0 +1,130 @@
+package migrations
+
+func init() {
+	appendMigration(
+		"populate pillars.ordr and questions.ordr with the canonical questionnaire order",
+		`
+-- pillars.ordr and questions.ordr have existed since 0002 but were never
+-- populated: every row is 0 except questions 41-47, the legacy "Future Plan..."
+-- questions that carry 901-907 so they sink to the bottom. With every other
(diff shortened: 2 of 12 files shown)


costacalvin (approved):
## Condensed review — `fix/functionoption-order`

### ACs

- **ztmf#531** — met, closes legitimately. Both ACs satisfied; measurement recorded on
  the issue (47.7s → 2.0s).
- **ztmf#535** — closed unmerged, so migration 0055 isn't duplicated on `main`. Confirmed
  absent from `upstream/main`.
- **ztmf#524 rider** — met. True no-op.
- **misc#277** — met, and it *corrects* the ticket (heap order, not insertion order).
  Post that back to the issue.
- **misc#278** — not met, defensibly. The enumeration AC is obsoleted by fixing it at the
  query level. Needs an explicit closing comment; it's an open P0.
- **Migration 0056 has no ticket** — the largest, riskiest part of the PR has no ACs to
  check against.

### Non-blocking

`FindFunctionOptions` interleaves all functions when unfiltered (unreachable today) ·
`FindQuestions` still unordered · `Function.Ordr` keeps the same clobber (nothing to destroy
yet) · `questions.go` newly gofmt-dirty · misindented comment in `answers.go` · dead
`IS NOT NULL` predicate in 0056 · `assertPillarsContiguous` passes on seed luck · orphaned
diff rows now sort last instead of first · three unrelated changes in one PR. 

### Verdict

Strong work, well documented, ACs met. Will verify through Dev environment that the questionnaire ordering is correct. Then it will be good to merge!

jsos3-cms (approved):
Approving. Calvin's pass already covers the ACs and the nits, so this is only what I verified separately.

Actions has been in a major outage since 15:22 UTC and this PR has no runs at all, so I ran the checks locally instead. Build and vet are clean, the full backend suite passes 13 of 13 packages against a freshly seeded database, the new integration tests pass, `make generate-openapi` produces zero drift, and `lint-openapi` validates. The branch already sits on top of current main, so there is no rebase and nothing to reconcile.

On verifying the questionnaire ordering through dev: most of that can be settled statically, and it holds. I compared migration 0056's ordering data against ztmf-ui's `PILLAR_ORDER` and `PILLAR_FUNCTION_MAP`, computing the rank each frontend entry implies and diffing in both directions. Forty function names on each side, nothing unique to either, zero rank mismatches. So the order the database will serve is identical to the order the frontend renders today, which means the export and the score diff will agree with the questionnaire rather than diverge from it. Dev is still worth an eyeball for the rendering itself, but the data is not the risk.

No frontend counterpart is needed either. The `ordr` clobber your comment warns about is not reachable from this client: ztmf-ui only reads functionoptions through `?include=functionoption` and writes only scores. There is no write path to functionoptions or questions.

Two smaller confirmations. pgx handles the pointer change correctly at the driver level, using a dedicated plan for pointer-to-pointer destinations where a NULL sets the outer pointer to nil, and `RowToStructByNameLax` is the right variant since the strict version errors on a column it cannot map. And plain `CREATE INDEX` was effectively the only option here rather than merely the right one: tern exposes transaction control through a marker in a SQL file, a global option, or a hand-built migration struct, and this repo registers everything through `AppendMigration(name, upSQL, downSQL)`, which surfaces none of them.

One correction to the description. It says this PR marks `pillarid` as `readOnly`. That shipped in #537 and is already on main. The only openapi change in this diff is the description block on `model.Question.order`. Worth fixing so nobody approves a claim the diff does not contain. Relatedly, `order` is a pointer now but the schema still describes it as a plain integer, so it does not document that a client may omit the field.

One request before this deploys. Could you re-run the "forty of forty non-legacy questions resolve to exactly one function name" audit against a fresh prod snapshot? The `AND q.ordr = 0` guard is what makes 0056 safe to re-run, but it also means a wrong-but-nonzero value cannot be corrected by running 0056 again. That would need a new forward migration. Cheap to check now, awkward to fix later.

And a note for wherever the rollback steps live: `order` is admin-writable now, so a down migration clears any hand re-ordering an admin has done, not only the values this migration wrote.

jsos3-cms:
Folded into #541, the backend freeze batch, so the approved PRs take one dev deploy and one prod deploy between them rather than three. Your commits went in unchanged and authorship is preserved, including the two `chore(ci)` retrigger commits.

`Closes #531` is re-declared on the batch, since a folded PR closes unmerged and its own keyword never fires.

Before folding, this branch got a green DEV deploy of its own via `Manual Deploy to DEV`, which is the first time migration 0056 has run against real data. The local test seed uses function names that never match the ordering map, so every rank stays at zero there and the curated sequence is not exercised. That covers the dev verification you wanted, @costacalvin.

jsos3-cms:
Landed in `main` as part of #541, squash commit 469b87e. Your commits went in unchanged and `Closes #531` fired from the batch, so ztmf#531 is closed.

Closing as superseded rather than abandoned. Thanks @danielbowne.

PR #540, opened 2026-08-06
techdebt(#484): move DB secret resolution from db/conn.go into config

### LINKED ISSUE #484: Move DB secret resolution from db/conn.go into config (existing TODO)

jsos3-cms:
In-repo version of #538 by @voidspooks, moved so the CI gates that need org-level secrets can run. Fork PRs do not receive `SNYK_TOKEN` or the AWS deploy role, so Snyk and the Infrastructure / Deploy jobs never ran there. The commit is pushed unchanged and authorship is preserved.

Closes #484
Refs #538

## What this does

Moves database secret resolution out of `backend/internal/db/conn.go` into `backend/internal/config`, replacing the package-level `sync.Once` and cached secret with fields on the config singleton. The relocated function bodies are identical to the originals, comments included. Only the receiver changed.

## Verification

The fork branch was cut one commit before #537, so it was rebased onto current `main` first. Everything below was run after the rebase rather than before it, since a clean rebase does not imply a working tree:

- `go build` and `go vet` clean, including no copylock finding from the newly embedded `sync.Once`
- Full backend suite passes 13 of 13 packages against a freshly seeded database, which exercises the real `pgxpool.BeforeConnect` path through `config.GetInstance().DbCreds()`
- `internal/config` clean under `-race`

Resolution is still lazy rather than eager, and a failed secret fetch still returns a hard error rather than an empty credential, matching the behaviour before the move.

Thanks @voidspooks.

### DIFF
--- a/backend/internal/config/config.go
+++ b/backend/internal/config/config.go
@@ -26,6 +26,13 @@ type smtp struct {
 	CertIntermediateSecretID *string `env:"SMTP_CA_INT_SECRET_ID"`
 }
 
+// DbCreds holds resolved database credentials, either read directly from env
+// vars or unmarshalled from the Secrets Manager secret named by DB_SECRET_ID.
+type DbCreds struct {
+	Username string
+	Password string
+}
+
 // config is shared by all binaries with values derived from environment variables
 type config struct {
 	Env      string `env:"ENVIRONMENT" envDefault:"production"`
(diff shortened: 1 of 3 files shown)


jsos3-cms:
Folded into #541, the backend freeze batch. The commit went in unchanged and is still authored by @voidspooks.

`Closes #484` is re-declared on the batch, since a folded PR closes unmerged and its own keyword never fires.

Worth stating plainly: this commit has never had an independent approval, and I cannot approve it because I opened the rehome. The batch body asks its reviewer to look specifically at this part rather than rubber-stamping the two changes they already know. So the review this needs happens at the batch.

jsos3-cms:
Landed in `main` as part of #541, squash commit 469b87e. `Closes #484` fired from the batch, so ztmf#484 is closed.

The commit is still authored by @voidspooks in the batch branch history. Note that because the batch was squash-merged, `main`'s log attributes the squash commit to me as the batch PR author, so the credit for this change lives here, in #541's body, and in #538.

Closing as superseded.

PR #541, opened 2026-08-06
Code freeze batch 3 (backend): questionnaire ordering, export index, login alarm tuning, DB secret resolution

### LINKED ISSUE #484: Move DB secret resolution from db/conn.go into config (existing TODO)

### LINKED ISSUE #531: Measure a prod-scale data-call export after #528, then decide on a functionoptions(functionid) index

jsos3-cms:
One batch carrying the approved backend PRs so they take a single dev deploy and a single prod deploy instead of three. The dev lane is shared and serialises, and three separate deploys tonight would have raced each other; that race has already cost three deploys today.

Closes #531
Closes #484

## What is folded in

| PR | Author | Approvals before folding |
| --- | --- | --- |
| #539 questionnaire ordering end to end, plus the export index | Daniel Bowne | costacalvin, jsos3-cms |
| #525 require 3-of-6 windows on the login auth-failure alarm | Daniel Bowne | jsos3-cms |
| #540 move DB secret resolution into config, rehome of #538 | Cameron Testerman | none |

Appended in approval order, then #540 last since it has no approval to order by. Nothing was reordered, so the record of what joined when survives.

**Please give the #540 commit real attention when reviewing.** It is the only part of this batch that has never had an independent approval. Its content was reviewed and came back clean, and it is a verbatim relocation rather than a rewrite, but a review verdict is not a recorded approval and I authored the rehome so I cannot approve it. Approving this batch is what puts a second pair of eyes on that commit.

## How it was built

Started from #539's head rather than cherry-picking its range, because #539 contains a merge commit (folding in `chore/functionoptions-index`) that a range pick cannot replay. #525's single commit and #540's single commit were then appended.

#539 carries two `chore(ci)` retrigger commits of its own. They were pushed to get past the immutable ECR tag on its original branch and are meaningless here, but removing them would have rewritten Daniel's commits, so they ride along.

This batch has since needed one of its own, for a reason worth writing down: the API image tag is derived from the commit, and ECR has tag immutability enabled, so a DEV deploy can never be re-run for a SHA that has already pushed an image. Refreshing the deployment requires a new commit. That is the mechanical reason the retrigger-commit pattern exists in this repo rather than a habit. Re-running only a *failed* deploy job does work, since the Docker push step has already succeeded and is skipped.

## Verification

Files are disjoint across the three changes, splitting into `internal/config` and `internal/db`, `internal/model` and the migrations, and `infrastructure`. Checked on the combined branch rather than trusting the individual runs:

- `go build` and `go vet` clean
- Full backend suite passes 13 of 13 packages against a freshly seeded database
- `internal/config` and `internal/db` clean under `-race`, which matters because #540 relocates a `sync.Once`
- `make generate-openapi` produces no drift, which #539 edits
- `terraform fmt -check` clean on the alarm file #525 edits

The suite run is the check worth having here: it exercises #540's relocated credential path and #539's new ordering and migration tests in the same process, which neither PR could do alone.

Each constituent also had its own green DEV deploy before being folded, so each piece is known to apply against real infrastructure. #539's deploy is the first time migration 0056 has run against real data, since the local seed's function names never match the ordering map and leave every rank at zero.

### DIFF
--- a/backend/cmd/api/internal/migrations/0055functionoptionsindex.go
+++ b/backend/cmd/api/internal/migrations/0055functionoptionsindex.go
@@ -1,0 +1,27 @@
+package migrations
+
+func init() {
+	appendMigration(
+		"add functionoptions(functionid) index for the data-call export",
+		`
+-- The data-call export (#528 re-anchor of FindAnswers) left-joins scores with
+-- a per-row subquery over functionoptions keyed on functionid. Postgres cannot
+-- fold that IN-subquery into the hash join, so it runs as a per-candidate-row
+-- Join Filter - and functionoptions has no functionid index (only its primary
+-- key), making each evaluation a sequential scan.
+--
+-- Measured on a prod-scale copy (ztmf#531, 1,530 systems / 192k scores): the
+-- whole-call FY26 export performs that scan 1,556,120 times, 47.7s of database
+-- time before the spreadsheet is even rendered - retry/abandon territory, and
+-- near typical gateway idle timeouts. With this index the same export runs in
+-- 2.0s (23x).
+--
+-- functionoptions is a small, static, read-mostly catalogue (1,728 rows; it
+-- changes only when a questionnaire edition is edited), so the write cost is
+-- negligible and no partial or composite form is warranted.
+CREATE INDEX IF NOT EXISTS functionoptions_functionid_idx
+    ON public.functionoptions (functionid);
+`,
+		`DROP INDEX IF EXISTS functionoptions_functionid_idx;`,
+	)
+}
--- a/backend/cmd/api/internal/migrations/0056orderingdata.go
+++ b/backend/cmd/api/internal/migrations/0056orderingdata.go
@@ -68,9 +68,18 @@ FROM (VALUES
 ) AS v(pillar, ordr)
 WHERE p.pillar = v.pillar;
 
+-- Deterministic by construction rather than by hope: a question is ranked by
+-- the MINIMUM matching rank among its functions' names. On prod data every
+-- question matches exactly one name (verified: 40/40 single-name), so min() is
+-- the identity; if a future edition ever forked a question's function name,
+-- min() still yields one stable, re-runnable answer instead of letting the
+-- planner pick a VALUES row arbitrarily.
 UPDATE questions q
-SET ordr = r.ordr
-FROM (VALUES
+SET ordr = ranked.ordr
+FROM (
+  SELECT f.questionid, min(r.ordr) AS ordr
+  FROM functions f
+  JOIN (VALUES
     ('Authentication-Users',                101),
     ('IdentityStores-Users',                102),
     ('RiskAssessment',                      103),
@@ -111,13 +120,12 @@ FROM (VALUES
     ('Cross-VisibilityAnalytics',           601),
     ('Cross-AutomationOrchestration',       602),
     ('Cross-Governance',                    603)
-) AS r(function_name, ordr)
-WHERE q.ordr = 0
-  AND EXISTS (
-      SELECT 1 FROM functions f
-      WHERE f.questionid = q.questionid
-        AND f.function = r.function_name
-  );
+  ) AS r(function_name, ordr) ON r.function_name = f.function
+  WHERE f.questionid IS NOT NULL
+  GROUP BY f.questionid
+) AS ranked
+WHERE q.questionid = ranked.questionid
+  AND q.ordr = 0;
 `,
 		`
 -- Reverses the population only. The legacy 901-907 band predates this migration
--- a/backend/cmd/api/internal/migrations/0056orderingdata.go
+++ b/backend/cmd/api/internal/migrations/0056orderingdata.go
@@ -1,0 +1,130 @@
+package migrations
+
+func init() {
+	appendMigration(
+		"populate pillars.ordr and questions.ordr with the canonical questionnaire order",
+		`
+-- pillars.ordr and questions.ordr have existed since 0002 but were never
+-- populated: every row is 0 except questions 41-47, the legacy "Future Plan..."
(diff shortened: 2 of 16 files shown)


jsos3-cms:
Notes on the config commit in this batch, carried over from reviewing it before the rehome so they do not die on a PR that closes unmerged. None are blocking. The relocation itself is verbatim and behaviour-preserving: the moved function bodies are byte-for-byte identical, only the receiver changed.

`DbCreds` is now an exported type with exported `Username` and `Password`, where the old unexported `dbCreds` was confined to `internal/db`. Any package importing `config` can now fetch live credentials through the singleton. A `String()` that redacts the password would make an accidental `%+v` harmless, and `config_test.go` already formats the struct that way, with a fake constant today.

The secret-backed resolution path has no test coverage. Removing its error handling entirely leaves every test green. #482 covers that, so this is noted rather than requested.

Pre-existing and unchanged by the move, listed so nobody reads them as new: a failed first secret fetch poisons the instance for the process lifetime because `sync.Once` never re-runs, startup is not fail-fast so a bad secret id surfaces on first connection rather than at boot, and a malformed payload can echo one character of itself into a log line.

jsos3-cms:
Recording the review of the config commit here, because it was done before the rehome and never made it onto a PR. That was my oversight: it existed only in local notes, and notes are not a review record.

What was checked, and how.

**The relocation is verbatim.** The moved function bodies are byte-for-byte identical to the originals in `origin/main`'s `internal/db/conn.go`, comments included. Only the receiver changed, from package-level `once` and `dbSecret` variables to fields on the config singleton.

**`go vet` is clean**, including no copylock finding despite the newly embedded `sync.Once`. Tests pass under `-race`. The full backend suite passes against a freshly seeded database, which exercises the real `pgxpool.BeforeConnect` path through `config.GetInstance().DbCreds()` end to end rather than in isolation.

**Concurrency was tested rather than reasoned about.** A throwaway fifty-goroutine stress test against the relocated `once.Do` confirmed the secret is still fetched exactly once, so the "call `once.Do` unconditionally, then re-check the field" discipline survived the move intact.

**One hypothesis was killed by testing it.** I suspected the `dbSecret` field, now embedded in the shared config singleton, could leak secret content through a naive `%+v` or JSON dump of the whole config. It cannot: `fmt` prints an unexported pointer-typed field as an address without dereferencing it, and `encoding/json` skips unexported fields entirely.

**Mutation testing found the one real coverage gap.** Removing the error handling around the secret fetch entirely leaves every test green, so the secret-backed resolution path has no coverage. That is #482's territory, and it is noted in the comment above rather than requested here.

**Failure behaviour is unchanged.** Resolution is still lazy rather than eager, and a failed fetch still returns a hard error rather than a zero-value credential, so a misconfigured secret fails the connection attempt instead of silently connecting with empty credentials.

Verdict: no critical issues, and nothing here that should block a merge.

A second independent pass over this batch as a whole is running now, including the question no constituent PR could answer on its own: whether relocating credential resolution interacts with the ordering migrations and integration tests that land alongside it. I will post that outcome here too, whichever way it comes back.

MackOverflow (approved):
LGTM!

jsos3-cms:
Second independent pass on the batch is in, as promised above. **No critical issues, and an explicit recommendation that nothing here should block a merge to main.**

It confirmed the config relocation line by line against `origin/main`: only the receiver and the package-level-var to struct-field syntax changed, and every conditional, log line, and error string is byte-identical. It also traced every error path and confirmed no secret value can reach a log line, error string, panic, or serialized struct, and that failure is still fatal at startup rather than a confusing later failure, unchanged from before the move.

**On the batch-interaction question, which no constituent PR could answer:** no functional interaction. Worth noting the two changes *do* share a code path, because `0056orderingdata_integration_test.go` calls `db.Conn`, which routes through the exact `BeforeConnect` to `DbCreds()` path the config commit relocated. Since the move is verbatim, there is no behavioural divergence for the ordering tests to trip over. Migrations 0055 and 0056 also both run single-threaded at startup, before any pool or concurrent path exists, so they never exercise concurrent secret resolution either. Migration numbering is sound, `make generate-openapi` is deterministic against a scratch copy, and the full suite passes all 13 packages against a seeded database.

**Correcting something I wrote in the PR body.** I claimed `internal/config` and `internal/db` are "clean under `-race`, which matters because #540 relocates a `sync.Once`." The first half is true and I re-verified it, but the second half oversells it: with `DB_SECRET_ID` empty, which the local and test harness requires, `DbCreds()` only ever takes the early-return branch and never reaches the `sync.Once`-guarded secret resolution. Neither new test calls `secrets.NewSecret`, and there is no mock for Secrets Manager in this repo. So the `-race` run does not exercise the code path I cited it as evidence for. The concurrency question is answered by static reasoning, not by the suite.

**Two non-blocking items worth a fast-follow, both pre-existing carry-overs rather than regressions:**

`RefreshDbCreds` (`internal/config/config.go:225`) reads `c.dbSecret` with a bare nil check outside `dbSecretOnce.Do`, which is exactly what `DbCreds()`'s own doc comment a few lines above warns against, since `once.Do` is what establishes the happens-before for that write. It is not reachable in practice today: `main.go` runs migrations synchronously before `ListenAndServe`, and that path warms the secret before any request goroutine exists. So startup ordering, not the `sync.Once`, is what makes the bare read safe. It is a latent trap for any future change to startup order, and it is structurally identical to what `db/conn.go` did before the move.

`DbCreds()` being exported widens credential access in a narrower way than I described earlier. The env-var path was already reachable, since `cfg.Db.User` and `cfg.Db.Pass` are exported fields on the singleton. What is newly reachable is the Secrets-Manager-resolved production credential, which previously required being inside `internal/db`. `internal/db/conn.go` remains the only caller today.

Happy to file those two as a follow-up if wanted; neither justifies holding the batch.

Issue #545, opened 2026-08-10
[P1] SaaS reduced questionnaire, phase 2: seed the pillar-exclusion rule and apply it to scoring and the questionnaire API (FY26 onward)


voidspooks:
### Context

Spun out of CMS-Enterprise/ztmf-misc#289. That work is split in two phases. @costacalvin is working Phase 1 now: the progress tracker counts SaaS systems out of 25 for FY26 onward, exports return the 25 in-scope rows for FY26 onward, and the frontend restores the Devices and Applications & Workloads pillars on historical calls. This ticket is Phase 2: make the backend authoritative for the reduced-pillar rule and carry it into scoring.

### Scope

1. Move the pillar-exclusion rule out of the hardcoded SQL literal (and the frontend constant) into a seeded table, so the rule is data and a future reduced-scope environment is a row, not a code change.
2. Score aggregate: systems on a reduced-pillar environment stop carrying Devices and Applications & Workloads pillar scores, FY26 onward, anchored on the data call deadline. System scores average over the pillars present.
3. Questionnaire endpoint: serve the reduced question set from the API so the frontend can delete its pillar filter (frontend deletion tracked separately in ztmf-ui).
4. Repoint the Phase 1 progress and export queries at the seeded table so the rule lives in exactly one place.

### Constraints

- FY26 onward only, anchored on the data call deadline. Closed cycles keep their existing pillar scores, system scores, tiers, and 40-row exports.
- Roughly 113 SaaS systems' current scores and tiers move when this lands mid-call, since a system score averages across whatever pillars are present. Needs a low-traffic deployment window and a heads-up before it ships.
- Decided 2026-08-11: CMS is not exempt and follows all other OpDivs in removing the two pillars from SaaS systems and their scoring (per Elizabeth, relayed in CMS-Enterprise/ztmf-misc#289). No OpDiv awareness needed anywhere in this change; the rule stays keyed on environment and data call only.

### Acceptance criteria

- [ ] The reduced-pillar rule is seeded data, not a SQL literal or a frontend constant
- [ ] FY26-onward reduced-scope systems carry no Devices or Applications & Workloads pillar scores
- [ ] The questionnaire endpoint returns 25 questions for reduced-scope systems and 40 for everything else
- [ ] Closed data calls are byte-for-byte unchanged: pillar scores, system scores, tiers, exports
- [ ] Emberfall e2e updated for the changed API contracts

[...]

PR #289, opened 2026-04-15
Add shared git hooks and expand gitignore coverage

danielbowne:
## Summary

- Added shared pre-commit and pre-push hooks in `.hooks/` that apply to all contributors automatically via `make dev-setup`
- Pre-commit hook blocks common build artifacts, Terraform state files, certificates, private keys, and files over 5MB from being committed
- Existing Go compilation and unit test checks preserved in pre-commit
- Expanded `.gitignore` to cover Terraform local files (`.tfstate`, `.terraform/`, `.tfplan`), build outputs (`bootstrap`, `*.zip`), and certificate formats (`.pem`, `.key`, `.crt`)

## How it works

- `make dev-setup` now runs `git config core.hooksPath .hooks` to point git at the shared hooks directory
- Hooks are committed to the repo so all contributors get them
- Pre-push hook runs the full test suite (unchanged from current behavior)

## Test plan

- [x] Pre-commit hook blocks `.tfstate` files from staging
- [x] Pre-commit hook blocks files >5MB
- [x] `.gitignore` rejects `*.tfstate` at `git add` time
- [x] Go compilation and unit test checks still run on `.go` file commits
- [x] Pre-push hook runs full test suite (76/76 E2E passing)

PR #546, opened 2026-08-10
fix(#289): scope SaaS progress counts and exports to 25 questions

costacalvin:
Part of CMS-Enterprise/ztmf-misc#289 — phase 1 of 2. **Deploy before ztmf-ui#690.**

## Problem

Devices and Applications do not apply to a SaaS-hosted system, so the consumer answers 25 questions, not 40. That is program-approved, but it was enforced only in the frontend (ztmf-ui#430) — every backend query still resolved 40. So the dashboard showed a permanent `25/40` that could never reach Complete (the chip is gated on `answered >= expected`, and the other 15 questions are unreachable in the UI), those systems stayed ranked as laggards by `progressSortValue` and matched the "Not updated" facet, and the export returned 15 questions nobody was asked.

## Approach

One predicate, `saasPillarScopeSQL` (`internal/model/pillarscope.go`), applied to **both** `buildScoreProgressSQL` CTEs and to `FindAnswers` as a top-level `WHERE`.

In scope from the FY26 cycle onward only: a cycle qualifies when it is named FY26 or its deadline falls after every FY26 cycle. Anchored on deadline, not `datacallid`, because ids are not chronological — `findPreviousDataCall` documents the same trap. Closed cycles were collected and reported against 40 and keep reading 40.

Three details are load-bearing rather than stylistic:

- **Both progress CTEs.** In `expected` alone, a denominator of 25 would face a numerator of 40 from carried-forward answers — breaking the documented answered-cannot-exceed-expected invariant and reporting a false "Complete" on a closed call.
- **Top-level `WHERE` in the export.** That functions join is applicable-OR-answered (#528), so filtering one branch would export 40 rows for a system whose excluded answers were carried forward and 25 for a freshly created one.
- **`COALESCE` on the scoring key.** It is nullable (`DECOMMISSIONED` maps to NULL) and `NOT(NULL AND …)` is NULL, which a `WHERE` reads as false. Without it, a decommissioned system carrying FY26 answers — which `FindAnswers` deliberately still exports — silently lost 15 rows, on a **non-SaaS** system.

**Scoring (`scores.go`) and the questionnaire endpoint (`questions.go`) are untouched, so no score moves.** Those, plus seeding the rule as data instead of a literal, are #545 — the issue's remaining two acceptance criteria (SaaS presenting exactly 25 questions from the API, and not being scored for the excluded pillars) belong to that PR, not this one. No rows are deleted here either: the retained answers stay in the database and stop counting.

## Acceptance criteria

- [x] Progress/completion calculates against 25 for SaaS systems — 25 for every SaaS system on FY2026, with the numerators drawn from the same set so `answered`/`updated` can never exceed it. A fully-answered closed call now reaches Complete instead of a permanent 25/40.
- [x] All other environments unchanged (40) — every other environment value, and non-SaaS export row counts byte-identical.
- [x] Closed historical calls unchanged — FY25 ZTM and FY2025 Q3 report 40 and export identical counts. The gate anchors on the *last* FY26 deadline rather than the first precisely so a mis-dated FY26 call cannot restate a closed cycle.
- [x] The export returns only the in-scope questions for FY26 onward (post-issue decision from Mack/Allie). See the row-count note below.

## Testing

- `make test-unit` — pass. `make test-integration` — pass.
- Against a production-shaped database: on FY2026 every SaaS system reports 25 and every other environment 40; on FY25 ZTM and FY2025 Q3 every system still reports 40; **no** row on any cycle has `answered`/`updated` exceeding `expected`; the FY2026 export drop matches the excluded-pillar count exactly, per system, with prior-cycle counts byte-identical; `/scores/aggregate` output unchanged.
- New: `pillarscope_test.go`; `pillarscope_integration_test.go`, which evaluates the predicate in Postgres because its correctness rests on three-valued logic no string assertion can pin — verified to fail when the `COALESCE` is removed; progress and export integration subtests covering the carried-forward case, a prior cycle, and a non-SaaS control.
- `EXPLAIN ANALYZE`: the `MAX(deadline)` and `EXISTS` plan as InitPlans with `loops=1`, evaluated once per query rather than per row. Export 1.79s vs 2.13s baseline (faster — fewer rows); whole-dashboard progress 57ms vs 25ms. No new indexes.

## Reviewer notes

- **CMS is deliberately not exempt.** The reduced scope is understood to be an HHS-OpDiv rule with CMS answering all 40, but that is unconfirmed — and exempting CMS here alone would give CMS SaaS systems a 40 denominator with 15 questions their ISSOs cannot see, since the frontend filter is OpDiv-blind. All SaaS systems read 25 for now; the clause to add is in `pillarscope.go` and the current behavior is pinned by `CMSAlsoReducedWhileExemptionIsUndecided` so re-adding it has to change a test.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1086,7 +1086,7 @@ END $$;
 BEGIN;
 
 -- Clean re-run: remove only what this file owns (scores cascade via datacall)
-DELETE FROM public.datacalls WHERE datacallid IN (101, 102, 103);
+DELETE FROM public.datacalls WHERE datacallid IN (101, 102, 103, 104);
 DELETE FROM public.fismasystems WHERE fismaacronym LIKE 'MOCK-%';
 
 -- The 3 historical HHS data calls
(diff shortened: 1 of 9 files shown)


MackOverflow (dismissed):
Looks good, provisionally approving, one question: 

1. Integration coverage likely skips in standard CI (worth confirming). The new behavioral tests all early-Skip when the DB has no SaaS functions or no FY26 cycle, and the code comments state the empire seed has neither. If make test-integration in CI runs against that seed, the executed guards reduce to the unit shape-tests — which only assert the SQL string contains expected fragments. Those catch a deleted COALESCE or a MIN-vs-MAX regression, but not a logic error that produces valid-looking SQL with wrong counts. The author validated against a production-shaped DB manually, which is great — but it'd be worth confirming whether CI has a seed that actually exercises these paths, or noting that these tests are effectively manual-run.

costacalvin:
Great catch @MackOverflow , I went ahead and added some extra data to our old seed data so our CI can run the tests correctly instead of being skipped. Verified all tests passing. Ready for re-review

voidspooks (changes requested):
See: https://aquiaworkspace.slack.com/archives/C090CCK3CE8/p1786394308333589?thread_ts=1786365329.102979&cid=C090CCK3CE8

Thoughts?

voidspooks (approved):
LGTM

Issue #268, opened 2026-02-18
Infrastructure: Verified Access multi-provider support


danielbowne:
# Infrastructure: Verified Access multi-provider support

**Epic**: Multi-IdP Authentication (#150)
**Future work** — Verified Access is currently commented out

## Summary

When AWS Verified Access is re-enabled, add a second trust provider for Entra ID alongside the existing Okta trust provider, and update Cedar policies to accept users from both IdPs.

## Current State

`infrastructure/verified-access.tf` is entirely commented out. The existing (commented) config:
- Hardcoded to `ztmf_idmokta` trust provider
- Cedar policy restricts to `*@cms.hhs.gov` emails only
- Single trust provider attachment to the Verified Access instance

## Requirements

- [ ] Add `aws_verifiedaccess_trust_provider` resource for Entra ID
- [ ] Update Cedar policy to accept emails from both CMS and partner domains
- [ ] Attach both trust providers to the Verified Access instance
- [ ] Parameterize domain restrictions in Cedar policy

## Acceptance Criteria

- [ ] Both Okta and Entra ID users can access ZTMF through Verified Access
- [ ] Cedar policy correctly restricts to configured email domains
- [ ] `terraform plan` shows clean changes

jsos3-cms:
We need to check if Verified access can be enabled in our environments before we can even start on this work. This would need more research/follow up with SecOps on whether this can be enabled in prod for our applications. Can be for post/after-datacall

PR #547, opened 2026-08-10
fix(datacalls): scope the data-call export by tier and stop query params redirecting scope

voidspooks:
Fixes two authorization defects in `GetDatacallExport`, both filed against the same handler, and retires the copy-pasted scope pattern that let the first one happen.

## The defects

**CMS-Enterprise/ztmf-misc#267** (OpDiv scope): the export used a two-way `HasAdminRead()` branch that dropped both OpDiv-scoped tiers into the unscoped path, so an OPDIV_ADMIN or OPDIV_READONLY_ADMIN exported every OpDiv's answers. Now scoped by the same three-way tier split the rest of the codebase uses, with a fail-closed OpDiv predicate added to `FindAnswers`: an OpDiv-restricted caller with no grants gets no rows.

**CMS-Enterprise/ztmf-misc#268** (query-param scope override): scope was set before the client query string was decoded onto the same struct, so `?UserID=<other>` redirected the export to another user's answers, and the scope pointer aimed at the session `User` meant decode mutated the authenticated identity itself. Now decode-then-scope (the ordering every other list handler documents), server-owned fields tagged `schema:"-"` so they cannot be bound from the query at all, and `UserID` copied off the session rather than pointed at its field. The path is now authoritative for the data call id.

## Retiring the pattern

#267 happened because the export predated the tier-scoping pattern and never adopted it. The same two snippets were hand-copied across the codebase: the tier-classification switch (six controllers) and the fail-closed OpDiv predicate (nine model queries). This PR extracts both into an embedded `OpDivScope` type:

- `ApplyTier(user)` owns the role classification (unscoped / OpDiv / self-scope). The self-scope default stays at the call site because it legitimately differs per endpoint (by UserID, by assigned systems, or a 403).
- `OpDivWhere` / `AppendRawFilter` own the fail-closed decision (restricted-and-empty means no rows) for squirrel and hand-built queries respectively. The grants predicate stays at the call site because the path from each query's base table to `opdiv_id` differs (direct column, `fismasystemid` subquery, `EXISTS` against `users_opdivs`).

The fields carry `schema:"-"`, so a server-owned scope field supplied in any list endpoint's query is now a 400 rather than silently honored. A new list endpoint can no longer mis-classify a tier or forget the fail-closed predicate.

The per-resource access guards (`usersopdivs.go`, `users.go`, single-system checks) are a different pattern and are left for CMS-Enterprise/ztmf#427.

## Behavior changes

- OpDiv-tier exports are now scoped to granted OpDivs; a zero-grant OpDiv caller gets an empty export.
- `?UserID=` and `?DataCallID=` on the export return 400 instead of being honored; the path is authoritative.
- The `schema:"-"` hardening is codebase-wide: server-owned scope fields on the scores, progress, diff, insights, fismasystems, and users list endpoints are likewise no longer bindable from the query (they were set from the session already; this makes an attempt a 400 rather than a silent no-op).
- Unscoped admin (OWNER, HHS_ADMIN, HHS_READONLY_ADMIN) and ISSO/ISSM behavior is unchanged.

## Tests

- `datacalls_export_authz_test.go`: export role matrix and the query-override attempt as an explicit negative case.
- `answers_integration_test.go`: OpDiv scoping, empty-grant fail-closed, and cross-OpDiv `fsids` returning nothing.
- Emberfall cases for an OpDiv-scoped export and the 400 rejection.
- The existing per-query OpDiv and progress tests are repointed to the embedded struct and still pass.
- `make test-unit`, `make test-integration`, `make test-e2e` green; staticcheck clean; OpenAPI spec unchanged.

## Sequencing

These are scope fixes and should land before CMS-Enterprise/ztmf#427 (the ACL centralization), whose "preserve existing authorization behavior" criterion would otherwise carry #268 forward. Touches the same `FindAnswers` function as #546, so whichever merges second takes a small rebase.

### DIFF
--- a/backend/cmd/api/internal/controller/datacalls.go
+++ b/backend/cmd/api/internal/controller/datacalls.go
@@ -60,20 +60,26 @@ func GetDatacallExport(w http.ResponseWriter, r *http.Request) {
 	user := model.UserFromContext(r.Context())
 	findAnswersInput := model.FindAnswersInput{}
 
-	if !user.HasAdminRead() {
-		findAnswersInput.UserID = &user.UserID
-	}
-
-	vars := mux.Vars(r)
-	if v, ok := vars["datacallid"]; ok {
-		fmt.Sscan(v, &findAnswersInput.DataCallID)
-	}
-
+	// Decode the client query FIRST, then set the path id and the caller's scope,
+	// so nothing client-supplied can widen or redirect them. This is the ordering
+	// every other list handler uses (see scopeScoreProgressInput); the export was
+	// the one place scope was set before decode, which let a query param overwrite
+	// it (ztmf-misc#268). The scope fields carry schema:"-" so a decode attempt on
+	// them is a 400, but the ordering is the real guard.
 	err := decoder.Decode(&findAnswersInput, r.URL.Query())
 	if err != nil {
 		respond(w, r, nil, err)
 		return
 	}
+
+	// The path is authoritative for the data call; set it after decode so
+	// ?DataCallID= cannot redirect the export.
+	vars := mux.Vars(r)
+	if v, ok := vars["datacallid"]; ok {
+		fmt.Sscan(v, &findAnswersInput.DataCallID)
+	}
+
+	scopeFindAnswersInput(user, &findAnswersInput)
 
 	answers, err := model.FindAnswers(r.Context(), findAnswersInput)
 	if err != nil {
@@ -106,6 +112,20 @@ func GetDatacallExport(w http.ResponseWriter, r *http.Request) {
 	// it so server-side observability catches mid-stream disconnects.
 	if err := file.Write(w); err != nil {
 		log.Printf("GetDatacallExport: error writing xlsx to response (datacallid=%d): %v", findAnswersInput.DataCallID, err)
+	}
+}
+
+// scopeFindAnswersInput applies the caller's tier to the export query input,
+// the same three-way split ListScores and scopeScoreProgressInput use: unscoped
+// admins (OWNER, HHS_ADMIN, HHS_READONLY_ADMIN) see every OpDiv; OpDiv tiers
+// fail-closed to their granted OpDivs; everyone else is limited to their
+// assigned systems. Before this the export used a two-way HasAdminRead() branch
+// that dropped both OpDiv tiers into the unscoped path, so an OpDiv admin
+// exported every OpDiv's answers (ztmf-misc#267). Extracted so the role matrix
+// is unit-testable without a database.
+func scopeFindAnswersInput(user *model.User, input *model.FindAnswersInput) {
+	if input.ApplyTier(user) {
+		input.UserID = user.UserIDPtr()
 	}
 }
 
--- a/backend/cmd/api/internal/controller/datacalls_export_authz_test.go
+++ b/backend/cmd/api/internal/controller/datacalls_export_authz_test.go
@@ -1,0 +1,129 @@
+package controller
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func int32PtrExport(v int32) *int32 { return &v }
+
+// TestScopeFindAnswersInput pins the role matrix the data-call export applies to
+// its query input. This is the security boundary for GET /datacalls/{id}/export:
+// the SQL is tested separately, so what matters here is that each tier populates
+// exactly the scope fields FindAnswers keys on, and that no tier is dropped into
+// the wrong branch (ztmf-misc#267).
+func TestScopeFindAnswersInput(t *testing.T) {
+	t.Run("OwnerUnrestricted", func(t *testing.T) {
+		input := model.FindAnswersInput{}
+		scopeFindAnswersInput(adminUser, &input)
+
+		assert.False(t, input.RestrictToOpDivIDs, "OWNER must not be OpDiv-restricted")
+		assert.Empty(t, input.OpDivIDs)
+		assert.Nil(t, input.UserID, "OWNER must not be limited to assigned systems")
+	})
+
+	t.Run("HHSReadonlyAdminUnrestricted", func(t *testing.T) {
+		input := model.FindAnswersInput{}
+		scopeFindAnswersInput(readonlyAdmin, &input)
+
+		assert.False(t, input.RestrictToOpDivIDs, "HHS_READONLY_ADMIN has unscoped read")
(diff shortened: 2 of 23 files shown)


jsos3-cms:
Heads up that `main` moved under this tonight, and in a way that matters more than a usual stale base.

`main` is now `80af5db`, which carries #546 (SaaS progress and export scoped to 25 questions from FY26 onward). This PR's base is `469b87e0`, so it predates that.

The reason it is worth rebasing before review rather than after: the two changes overlap on four files, including `internal/model/answers.go` and `internal/model/scoreprogress.go`. Both are adding scoping conditions to the same two query builders. #546 narrows by pillar scope for SaaS systems, this PR narrows by tier, and neither author could see the other's work. So the green checks here were earned against a tree with no SaaS pillar scoping in it, and they do not tell us how the two filters compose. Over-filtering an export is the failure mode that is easy to miss, since nobody notices rows that are absent.

Worth a specific look after the rebase at whether the tier scoping lands in the same branch of the applicable-OR-answered functions join that #546 had to use a top-level `WHERE` for. #546's description explains why that mattered there.

Suggested order, because `dismiss_stale_reviews_on_push` is on in this repo: rebase onto `80af5db` first, let CI re-run against the combined tree, then get the review. Reviewing first means the rebase throws the approval away and someone has to do it twice.

Nothing shipped tonight depends on this one, so there is no rush.

voidspooks:
Rebased onto 80af5db. I went through the four files that overlap with #546 rather than trusting the clean merge, the two changes compose fine. Full suite green locally, ready for review.

voidspooks:
The deploy red is not this branch: #549's PR deploy applied migration 0057 to the shared dev DB, and any image without 0057 now fatals on startup until #549 merges. Code checks are all green.

jsos3-cms (approved):
**GO.** Both filed defects are fixed at the query level and well covered. One non-blocking item below that I'd like folded into this PR rather than deferred — it's two lines.

## What I verified

Read every touched controller and model file at the PR head rather than just the diff hunks, specifically hunting for the "authorize on a caller-supplied field instead of the path or session row" class this repo has a filed precedent for. `go build ./...`, `go vet ./...` and `go test -short ./cmd/api/internal/controller/... ./internal/model/...` all pass clean — 463 tests across 2 packages.

ztmf-misc#267 is genuinely closed. The old branch keyed on `HasAdminRead()`, which is `IsAdmin() || IsReadOnlyAdmin()` — a union that swept `OPDIV_ADMIN` and `OPDIV_READONLY_ADMIN` into the "no filter" path. `ApplyTier`'s two-case switch (`opdivscope.go:169-178`) no longer has that union.

ztmf-misc#268 is closed by ordering rather than tagging alone. `GetDatacallExport` decodes the client query, then unconditionally overwrites `DataCallID` from the path (`datacalls.go:75-80`), then scopes (`:82`). I confirmed against the pinned `gorilla/schema v1.4.1` that `schema:"-"` produces a hard decode error rather than a silent no-op, and that `sanitizeErr` maps that error type to 400 — matching the emberfall assertion.

Fail-closed is enforced at the SQL layer, not post-filtered in Go: `OpDivWhere` returns a literal `FALSE` predicate when the grant list is empty.

## Non-blocking: `FindScoresInput` missed the hardening sweep

`backend/internal/model/scores.go:565`:

```go
type FindScoresInput struct {
	input
	FismaSystemID  *int32 `schema:"fismasystemid"`
	FismaSystemIDs []*int32
	DataCallID     *int32 `schema:"datacallid"`
	UserID         *string
	IncludePillars *bool `schema:"include_pillars"`
	OpDivScope
}
```

`UserID` and `FismaSystemIDs` carry no `schema:"-"`. Every analogous field in every other struct this PR touches got one here — `answers.go`, `fismasystems.go`, `scorediff.go`, `scoreprogress.go` and `systeminsights.go` all do. This is the one struct in the sweep that doesn't, which makes the description's claim that scope fields are no longer bindable on "the scores, progress, diff, insights, fismasystems, and users list endpoints" true for five of the six.

It's live rather than theoretical: `ListScores` decodes before `ApplyTier` runs, and `ApplyTier` only overwrites `UserID` for the ISSO/ISSM self-scope case — for OWNER, HHS_ADMIN, HHS_READONLY_ADMIN, OPDIV_ADMIN and OPDIV_READONLY_ADMIN it returns false, so whatever arrived in `?UserID=` survives into the query.

**It is not exploitable today, which is why I'm not blocking.** Both consumption sites are `INNER JOIN` (`scores.go:588` and `:852`), there is no `OrWhere` or `.Or(` anywhere in the file, and the OpDiv predicate composes as a separate AND-ed `Where`. A supplied `UserID` can therefore only narrow a result set the caller was already authorized to see.

The reason I'd still close it here rather than after: it's a defense-in-depth gap in the exact bug class this PR exists to close, sitting in the one file with no negative test. Any future change that turns one of those ANDs into an OR, or adds a consumer that isn't AND-guarded the same way, reopens #268's shape silently. The fix is two struct tags plus the same query-cannot-widen test that `TestScopeFindAnswersInput_QueryCannotWiden` already gives the export handler.

Entirely happy for you to land this as-is and follow up separately if you'd rather — flagging mainly that the follow-up lands in someone else's context by then.

## Informational, no action needed

`backend/internal/model/fismasystems.go:144-159` — the "widen to admit UserID-assigned systems when the caller is also OpDiv-scoped" branch has no reachable caller; every construction site sets either `UserID` or `OpDivIDs`/`RestrictToOpDivIDs`, never both. This predates the PR (confirmed against `main`) and was only mechanically rewrapped into the shared `OpDivWhere` call here. Noting it so it isn't mistaken for something this change introduced.

jsos3-cms:
Rebased this onto `9d9c12c` and merged it — flagging the changed SHA so it isn't a surprise.

The branch was based on `80af5db`, which predates #549. #549's own PR deploy had already applied migration 0057 to the shared dev database, so any image built without 0057 fataled on startup. That is why `Infrastructure / Deploy` here was red at 20m28s while every code check passed — your diagnosis was right. Merging #549 put 0057 on `main`, but this branch still had to pick it up before its image would start, so a deploy re-run alone would have failed identically.

The rebase applied cleanly as a single commit with your authorship intact (`8611f92` → `04546bc`). I re-ran the suite locally rather than trust a clean rebase, since a base change can break a branch git sees no conflict in: `go build ./...`, `go vet ./...` and `go test -short ./...` all clean, 622 tests across 19 packages. CI on the rebased commit came back green across all 11 checks with the deploy passing in 3m3s, and it merged as `e03d171`.

Nothing in your change was modified — only the base it sits on.

Worth noting for next time: this was a PR-branch deploy mutating shared dev state and breaking a sibling PR that touched no migrations. Any migration-bearing PR can do the same to whatever else is open.

PR #549, opened 2026-08-11
feat(#545): seed the reduced-pillar rule and apply it to scoring and the questionnaire

### LINKED ISSUE #545: [P1] SaaS reduced questionnaire, phase 2: seed the pillar-exclusion rule and apply it to scoring and the questionnaire API (FY26 onward)

voidspooks:
Closes #545. Phase 2 of CMS-Enterprise/ztmf-misc#289.

## What

The reduced-pillar rule moves from a name-matching SQL literal into a seeded table, and the two consumers #546 deliberately left untouched now honor it. Migration 0057 creates `reducedpillarscopes` (scoring_key, pillarid, effective_datacallid): a row excludes that pillar for systems scored under that key, on every data call whose deadline is on or after the anchor call's. The shared predicate now reads the table, so the progress counts, the export, the scoring aggregate and the questionnaire resolve the same rule from the same rows, and a future reduced-scope environment is an INSERT, not a code change.

Scoring: the aggregate's expected CTE applies the predicate, so an excluded pillar produces no pillar score and the system score averages the pillars that remain (the divisor already followed the actual pillar count). Carried-forward answers on excluded pillars stop counting because the answers CTE is only read through the join from expected.

Questionnaire: `GET /fismasystems/{id}/questions` accepts an optional `datacallid` and serves the reduced set for in-scope cycles, so the frontend can delete its client-side pillar filter (tracked separately in ztmf-ui). Without the param the response is byte-identical to today, which keeps the current frontend correct until it passes the param.

Seeding: in deployed environments migration 0057 locates the FY26 anchor by the same name prefixes the interim predicate matched, once, at migration time; after that the rule is pure data. The ephemeral local/test databases get their rule rows from the empire seed, which owns the fixture cycles and runs after migrations. A deployed environment with no FY26-named call at migration time seeds nothing and keeps the full set until a rule row is inserted alongside the cycle that needs it; that environment-dependence is inherent to rule-as-data and deliberate.

Per the decision recorded on ztmf-misc#289 (2026-08-11), CMS is not exempt and there is no OpDiv awareness anywhere in this change. The interim `CMSAlsoReducedWhileExemptionIsUndecided` pin is renamed to assert the now-permanent behavior.

## Visible behavior changes

`/scores/aggregate` returns four pillar entries and a four-pillar system score for SaaS systems on FY26-onward cycles. Roughly 113 SaaS systems' current scores and tiers move when this deploys mid-call, so it needs a low-traffic window and a heads-up before it ships (flagged on #545). The questionnaire endpoint changes only when the new param is passed. Progress and export behavior is unchanged, now sourced from the seeded table instead of name matching.

## Acceptance criteria

- [x] The reduced-pillar rule is seeded data, not a SQL literal or a frontend constant. No cycle name, environment, or pillar is compiled in; pinned by the shape tests.
- [x] FY26-onward reduced-scope systems carry no Devices or Applications pillar scores. Integration-pinned for the SaaS fixture and its CMS twin, with carried-forward excluded answers present.
- [x] The questionnaire endpoint returns the reduced set for reduced-scope systems and the full set for everything else, keyed by the new optional `datacallid` param.
- [x] Closed data calls are unchanged: pillar scores, system scores, tiers, exports, and question sets all pinned on a pre-anchor cycle.
- [x] Emberfall updated for the changed API contracts (five new pinned cases).

## Testing

`make test-unit`, `make test-integration`, and `make test-e2e` all green locally (e2e: 206 ran, 0 failed, 0 skipped). New integration coverage: aggregate reduction, questionnaire reduction with and without the param plus a non-reduced control system, and a pin that the seed wiring produced exactly the two SaaS rule rows. `openapi.yaml` regenerated via `make generate-openapi`; redocly lint valid.

### DIFF
--- a/backend/_test_data_empire.sql
+++ b/backend/_test_data_empire.sql
@@ -1137,6 +1137,17 @@ INSERT INTO public.datacalls (datacallid, datacall, datecreated, deadline) VALUE
  (104,'FY26 ZTM','2025-10-01 00:00:00+00','2026-09-11 00:00:00+00')
 ON CONFLICT DO NOTHING;
 
+-- The seeded reduced-scope rule (ztmf#545), anchored at the fixture FY26 cycle.
+-- Migration 0057 seeds these rows only in environments that already hold an
+-- FY26-named call when it runs; this file's data calls are inserted after
+-- migrations, so the local/test rule rows are owned here. The clean re-run
+-- DELETE above cascades them (effective_datacallid is ON DELETE CASCADE).
+INSERT INTO public.reducedpillarscopes (scoring_key, pillarid, effective_datacallid)
+SELECT 'SaaS', p.pillarid, 104
+  FROM public.pillars p
+ WHERE p.pillar IN ('Devices', 'Applications')
+ON CONFLICT DO NOTHING;
+
 -- One SaaS function per pillar, derived rather than hardcoded: ids 7101-7106 sit
 -- above every explicit id in this file, and each function borrows an existing
 -- question in its pillar so the applicability join (which reads
@@ -1166,9 +1177,9 @@ VALUES
  (2016,'5A1A5000-0000-4000-8000-000000002016','MOCK-SAAS','Mock SaaS Reduced Scope System','SaaS',
   'mock.isso2016@example.empire', TRUE,
   (SELECT opdiv_id FROM public.opdivs WHERE UPPER(code) <> 'CMS' ORDER BY opdiv_id LIMIT 1),'Mock ISSO 2016'),
- -- CMS counterpart: the reduced scope is understood to be an HHS-OpDiv rule with
- -- CMS keeping all 40, but that is undecided, so both currently read 25. Without a
- -- CMS SaaS system the subtest pinning that interim behaviour skips.
+ -- CMS counterpart: CMS follows every other OpDiv - no exemption, per the
+ -- decision recorded on ztmf-misc#289 (2026-08-11) - so both systems read 25.
+ -- Without a CMS SaaS system the subtest pinning that skips.
  (2017,'5A1A5000-0000-4000-8000-000000002017','MOCK-SAAS-CMS','Mock CMS SaaS Reduced Scope System','SaaS',
   'mock.isso2017@example.empire', TRUE,
   (SELECT opdiv_id FROM public.opdivs WHERE UPPER(code) = 'CMS' LIMIT 1),'Mock ISSO 2017')
--- a/backend/cmd/api/internal/controller/questions.go
+++ b/backend/cmd/api/internal/controller/questions.go
@@ -16,7 +16,9 @@ import (
 //	@Produce	json
 //	@Security	bearerAuth
 //	@Param		fismasystemid	path		int	true	"FISMA system ID"
+//	@Param		datacallid		query		int	false	"Apply the reduced-pillar rule for this data call; omitted returns the environment's full catalog"
 //	@Success	200				{object}	apiResponse[[]model.Question]
+//	@Failure	400				{object}	apiResponse[any]
 //	@Failure	404				{object}	apiResponse[any]
 //	@Failure	500				{object}	apiResponse[any]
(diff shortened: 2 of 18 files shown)


MackOverflow (approved):
looks solid, one question: 

Behavior change: this endpoint now rejects previously-ignored query params (medium)
Confirmed: the shared decoder = schema.NewDecoder() has no IgnoreUnknownKeys(true), and sanitizeErr maps schema.UnknownKeyError → 400. Before this PR, ListFismaSystemQuestions never decoded the query string, so GET /fismasystems/{id}/questions?anything=x returned 200 and ignored the extra param. After this PR, any unrecognized query param returns 400.

This aligns the endpoint with every other list handler (a consistency win), but it's a real contract tightening on a GET. The concrete risk: cache-busting params some HTTP clients/proxies append (?_=1699999999, ?cacheBust=…) would now 400 where they used to pass. Worth a quick confirmation that no FE caller, proxy, or monitoring probe hits this route with incidental params. If that's a concern, decoder.IgnoreUnknownKeys(true) on the shared decoder would preserve leniency — though that's a broader decision than this PR.

costacalvin (approved):
LGTM! Tested locally and in Dev, all looks good. Even removed the FE filtering just to verify and still worked as expected.

voidspooks:
@MackOverflow thanks for the comment. My response is below:

Checked all three frontend call sites and the axios instance, everything hits this route bare, no cache busting params anywhere and axios appends nothing on its own. Strict decode is already what the other 17 handlers do, so every list GET the frontend uses daily has this contract and this was the last lenient one. I'd keep it strict rather than add IgnoreUnknownKeys globally: a silently honored query param is exactly the shape of the misc#268 finding, and #547 leans on the unknown-key 400 to surface a ?UserID= attempt instead of eating it.

PR #551, opened 2026-08-11
fix: declare CloudFront price class and assert it

jsos3-cms:
## What

Declares `price_class` on the CloudFront distribution and adds a `lifecycle` postcondition asserting it. Two additions, twelve lines, no other change.

## Why

`price_class` was inherited from the AWS provider's schema default rather than declared. That default is not stated in the provider's published argument reference — it exists only in the SDK schema — so the effective value was implicit and dependent on an upstream choice, while `terraform.tf` permits any 5.x minor.

Declared as a literal rather than a variable so the value is identical in every environment. Placed adjacent to `restrictions` to match the canonical example in the registry docs.

The postcondition evaluates against refreshed state during `plan`, so it catches an out-of-band change as well as a bad diff, and it fails at plan time rather than at apply.

## Verification

- `terraform plan` against dev reports `No changes. Your infrastructure matches the configuration.` both before and after this commit. This is a no-op against live state.
- `price_class` carries no `ForceNew` in the provider schema, so it cannot trigger distribution replacement under any value change. No recreate, no new distribution ID or domain name, no propagation event.
- Postcondition negative-tested by falsifying the assertion — Terraform correctly fails the plan, confirming the guard is live rather than inert.
- `terraform fmt -check` clean.

## Scope

The `restrictions` block is untouched and unasserted. An earlier revision of this PR also asserted the geo restriction; that was dropped deliberately. Geo enforcement interacts with the WAF layer in ways that an assertion here would prematurely lock in, so the distribution's geo posture is left free to change independently of this PR.

Tracked in CMS-Enterprise/ztmf-misc#293.

### DIFF
--- a/infrastructure/cloudfront.tf
+++ b/infrastructure/cloudfront.tf
@@ -243,6 +243,12 @@ resource "aws_cloudfront_distribution" "ztmf" {
     error_caching_min_ttl = 10
   }
 
+  // Declared rather than inherited: the provider does default this to
+  // PriceClass_All, but that default is undocumented in the registry and lives
+  // only in the SDK schema, so leaving it implicit makes the effective value
+  // invisible here and dependent on an upstream choice we do not control.
+  price_class = "PriceClass_All"
+
   restrictions {
     geo_restriction {
       restriction_type = "whitelist"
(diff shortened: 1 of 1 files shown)


jsos3-cms:
Sequencing note, since #552 is open alongside this.

**Merge #552 first, then this.** The two are independent — this changes `infrastructure/cloudfront.tf` only, #552 changes `backend/internal/model/scores.go` and adds a controller test, so there is no shared file and no conflict in either direction. The reason for the order is that #552 was already based on current `main`, whereas this branch predated #547 and #549 and has just been updated onto `e03d171` so its checks re-run against the same base everything else merged on.

Nothing here blocks or is blocked by #552 beyond that.

jsos3-cms:
Retracting the sequencing note above — it no longer applies.

That note existed because this branch predated #547 and #549 and had not been verified against the base #552 was on. Both have since merged, and this branch is now rebased onto `e03d171` as a single commit (`fe6580b`), the same base as #552.

There is no ordering between the two. They touch disjoint files — this is `infrastructure/cloudfront.tf` only, #552 is `backend/internal/model/scores.go` plus a controller test — and both are now verified against identical current `main`. Merge them in whichever order review lands.

The one thing that does still apply is on #552 rather than here: closing CMS-Enterprise/ztmf-misc#296 by hand when it merges, since a cross-repo `Closes` does not fire.

costacalvin (approved):
LGTM!

PR #552, opened 2026-08-12
fix(scores): tag the server-owned fields on FindScoresInput as non-bindable

jsos3-cms:
## What

Adds `schema:"-"` to the two server-owned fields on `FindScoresInput`, and pins the boundary with tests.

```diff
 type FindScoresInput struct {
 	input
 	FismaSystemID  *int32   `schema:"fismasystemid"`
-	FismaSystemIDs []*int32
+	FismaSystemIDs []*int32 `schema:"-"`
 	DataCallID     *int32   `schema:"datacallid"`
-	UserID         *string
+	UserID         *string  `schema:"-"`
 	IncludePillars *bool    `schema:"include_pillars"`
 	OpDivScope
 }
```

## Why

`FindScoresInput` is the one query-input struct without this tagging. Its siblings — `FindAnswersInput`, `FindFismaSystemsInput`, `FindScoreDiffInput`, `FindScoreProgressInput`, `FindSystemInsightsInput` — all carry it.

Both fields are populated by the server, never by a caller: `UserID` from the session via `ApplyTier`, and `FismaSystemIDs` from `user.AssignedFismaSystems` (`controller/scores.go:322`, its only assignment anywhere). Since neither is a parameter callers are meant to send, decode should reject them rather than accept them silently. `ApplyTier` assigns `UserID` only for the tiers needing self-scope, so for any other tier a decoded value would otherwise persist into the query.

## Tests

Three cases, mirroring the shape `datacalls_export_authz_test.go` already uses for the export input:

- **Negative** — the two tagged fields fail to decode from the query string.
- **Positive** — `fismasystemid`, `datacallid` and `include_pillars` still decode normally, so the tagging doesn't cost callers anything they legitimately send.
- **Handler** — `ListScores` answers 400 for a request naming a server-owned field, rather than proceeding. Decode failing short-circuits before the query runs, so this needs no database.

## Verification

- `go build ./...` and `go vet ./...` clean
- `go test -short ./...` — **630 passed across 19 packages**, up from 622 by exactly the eight added here
- Confirmed `FismaSystemIDs` has no caller that sets it from a request, so the tag breaks nothing

Small and self-contained; no behavior change for any legitimate request.

### DIFF
--- a/backend/cmd/api/internal/controller/scores_authz_test.go
+++ b/backend/cmd/api/internal/controller/scores_authz_test.go
@@ -1,0 +1,73 @@
+package controller
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/CMS-Enterprise/ztmf/backend/internal/model"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestFindScoresInput_QueryCannotWiden pins that the scores list endpoints'
+// server-owned scope fields are not bindable from the query string.
+//
+// ListScores and GetScoresAggregate decode the client query onto the input
+// before ApplyTier runs, and ApplyTier only overwrites UserID for the tiers that
+// need self-scope. For every other tier it returns false and leaves the decoded
+// value in place, so the tag is what keeps a client-supplied value out of the
+// query in the first place.
+func TestFindScoresInput_QueryCannotWiden(t *testing.T) {
+	for _, q := range []string{"UserID", "FismaSystemIDs"} {
+		t.Run(q, func(t *testing.T) {
+			input := model.FindScoresInput{}
+			err := decoder.Decode(&input, url.Values{q: {"1"}})
+
+			assert.Error(t, err,
+				"a server-owned scope field must not be bindable from the query string")
+		})
+	}
+}
+
+// TestFindScoresInput_LegitimateQueryParamsStillDecode is the companion positive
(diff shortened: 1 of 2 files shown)


jsos3-cms:
Sequencing note, since #551 is open alongside this.

**Merge this one first.** The two are technically independent — this touches `backend/internal/model/scores.go` and a new controller test, #551 touches `infrastructure/cloudfront.tf` only, so there is no shared file and no conflict in either direction. The order is practical rather than technical: this branch is already based on current `main` (`e03d171`), while #551 was opened before #547 and #549 landed and needed a branch update to be verified against the same base.

**On merge, close CMS-Enterprise/ztmf-misc#296 by hand.** That issue tracks this gap, and a cross-repo `Closes` does not fire — the reference has to be resolved manually.

jsos3-cms:
Retracting the "merge this one first" in the sequencing note above — it no longer applies.

That ordering was practical rather than technical: #551 predated #547 and #549 and had not been verified against the base this branch was on. Both have since merged, and #551 is now rebased onto `e03d171` as a single commit, the same base as this PR. Both have green checks against identical current `main`.

So there is no ordering between them — merge in whichever order review lands.

The rest of that note stands: **on merge, close CMS-Enterprise/ztmf-misc#296 by hand**, since a cross-repo `Closes` does not fire.

costacalvin (approved):
LGTM! No nits

PR #555, opened 2026-08-14
Return assignedopdivids as an empty array instead of null

### LINKED ISSUE #346: Return assignedopdivids as an empty array instead of null for users with no OpDiv grants
### Feature Type
General Feature

### Workflow Description

Both the users list (`FindUsers`) and the user detail (`findUser`) return `assignedopdivids` as JSON `null` for a user with no OpDiv grants, because `ARRAY_AGG` over zero rows is SQL `NULL` and the `[]*int32` field scans a NULL into a nil slice. The two endpoints are consistent with each other, and the frontend already handles it (`row.assignedopdivids ?? []`), so this is not a bug. It is a contract-tidiness nit: an always-array shape (`[]` instead of `null`) is friendlier for any current or future consumer and removes the need for null handling.

Raised by MackOverflow on PR #344.

### Acceptance Criteria

- [ ] `assignedopdivids` is always a JSON array (`[]` when the user has no grants), never `null`
- [ ] Applied to BOTH `FindUsers` (list) and `findUser` (detail) so the two stay consistent
- [ ] Implemented with `COALESCE(ARRAY_AGG(opdiv_id), '{}')` in the subquery
- [ ] Emberfall assertion added so the shape cannot regress

### Additional Context

Low priority / tech debt. Pure response-shape cleanup, no behavior change. Follow-up from the users-list N+1 fix (#344).

costacalvin:
## Summary

Closes #346.

`ARRAY_AGG` over zero rows is SQL `NULL`, which scans into a nil slice and serializes as JSON `null`. A user holding no OpDiv grants was therefore indistinguishable from a response that omitted the field. Wraps the subquery in `COALESCE(..., '{}')` and lifts it to a shared const so the paths cannot drift.

The spec was already on the fixed side of this. `backend/openapi.yaml` has always declared the field as a plain array:

```yaml
assignedopdivids:
  items:
    type: integer
  type: array
```

No `nullable: true`. So the API was returning `null` for a field its own published contract declared non-nullable. This brings the implementation into conformance with the schema rather than the other way round — which also means the spec needs no regeneration (verified: regenerating is byte-identical).

## Changes

* **`internal/model/users.go`:** `assignedOpDivIDsSubquery` const, applied to the `/users` paths — `FindUsers`, `findUser`, `Save` (insert and update), and `RestoreUser`. Scans realigned where the `RETURNING` list grew. The delegate paths deliberately do not select it; see below.
* **`internal/model/users_opdivarray_integration_test.go` (new):** pins the contract on both the scanned slice and the marshalled JSON.
* **`emberfall_tests.yml`:** assertions on create, get, list, update, restore, and `/users/current`, plus a delegates-roster assertion pinning the field as `null` there.

## Why this stops at the /users resource

The delegate paths deliberately do **not** select the field, because populating it there would leak data an authorization gate already withholds:

* An ISSO is `403`'d on `GET /users` — user management is restricted, and that is where OpDiv membership is otherwise exposed.
* That same ISSO **can** read their own system's delegate roster: `ListDelegates` gates only on `CanAccessFismaSystem`.
* A delegate is granted the OpDiv of every system they are attached to (`AddSystemDelegate` inserts into `users_opdivs`), so a shared delegate's other-OpDiv memberships would become visible to a system-scoped viewer.

The counter-argument — that these responses already carry `assignedopdivids` as `null`, so selecting it merely corrects a value — is about schema conformance, not confidentiality. `null` discloses nothing; a populated array discloses real membership. Those are separate questions, and the confidentiality one wins.

So the delegate responses keep spelling the field `null`. The consequence is that a client's `?? []` stays load-bearing off the `/users` resource, which is the right trade. The emberfall roster assertion pins the `null`, so adding the subquery to those paths fails loudly — verified by adding it and watching the assertion go red.

## Acceptance criteria

* [x] `assignedopdivids` is always a JSON array (`[]` when no grants), never `null`
* [x] Applied to both `FindUsers` (list) and `findUser` (detail)
* [x] Implemented with `COALESCE(ARRAY_AGG(opdiv_id), '{}')` in the subquery
* [x] Emberfall assertion added so the shape cannot regress

## Testing

`go build`, `go vet`, `go test -short ./...`, `make test-integration`, and `make test-e2e` (211 cases) all clean.

Both layers verified non-vacuous by reverting the COALESCE:

| Check | Result |
| --- | --- |
| Go integration, COALESCE reverted | 4 subtests red |
| emberfall, COALESCE reverted | 23 red |
| emberfall, subquery re-added to a delegate path | roster assertion red |

The Go test resolves a grantless user by query rather than hardcoding an id, so it cannot pass vacuously against a fixture where everyone is granted, and `GrantsAreNotFlattened` cross-checks every seeded user's array length against the `users_opdivs` counts — so a COALESCE that flattened everyone to `[]` would fail too.

### DIFF
--- a/backend/emberfall_tests.yml
+++ b/backend/emberfall_tests.yml
@@ -135,7 +135,7 @@ updatedUserData: &updatedUserData
   fullname: "Updated Test User"
   role: "ISSO"
 
-# Response-only anchors, so the ztmf#346 always-array assertion can live here.
+# Response-only anchors, so the always-array assertion can live here.
 # userData / updatedUserData double as request bodies, so theirs is spelled at
 # each expect site instead.
 deletedUserData: &deletedUserData
@@ -278,7 +278,7 @@ tests:
             role: "OWNER"
             identity_provider: "okta"
 
-  # ztmf#346 on the caller's own record - the frontend reads this into
+  # The caller's own record - the frontend reads this into
   # callerGrantIds (ztmf-ui#642). Served from the auth middleware's context
   # user, so no /users test would catch a regression here. Delegate.User holds
   # no grants, so the expectation is stable against the opdivs serial.
@@ -518,8 +518,8 @@ tests:
       headers:
         content-type: "application/json"
 
-  # ztmf#346 stops at the /users resource on purpose. An ISSO is 403'd on GET
-  # /users but can read their own system's roster, so populating OpDiv
+  # The always-array contract stops at /users on purpose. An ISSO is 403'd on
+  # GET /users but can read their own system's roster, so populating OpDiv
   # membership here would leak exactly what that 403 withholds. Pinned as null
   # so re-adding the subquery to the delegate paths fails loudly.
   - url: http://localhost:8080/api/v1/fismasystems/1001/delegates
@@ -1393,7 +1393,7 @@ tests:
         json:
           data:
             <<: *userData
-            # ztmf#346: a brand-new user holds no grants; still [], not null.
+            # A brand-new user holds no grants; still [], not null.
             assignedopdivids: []
 
   - url: "http://localhost:8080/api/v1/users/{{.createUser.Response.data.userid}}"
@@ -1414,7 +1414,7 @@ tests:
             # passing an explicit identity_provider (see the override gate).
             identity_provider: "okta"
 
-  # ztmf#346 on the list path. Before the PUT below so the email filter still
+  # The list path. Before the PUT below so the email filter still
   # matches this user.
   - url: "http://localhost:8080/api/v1/users?email=New.User@example.com"
     method: GET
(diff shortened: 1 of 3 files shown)


tarratsco (dismissed):
LGTM approving  

Clean, well-scoped fix. Two optional nits, neither blocking, so up to you:

1. `hardDeleteUserByEmail`: worth a glance to confirm it cascades to (or scrubs) `users_opdivs`. The final grant insert in `TestAssignedOpDivIDsWritePathsIntegration` would otherwise leave a junction row behind after `Cleanup`. If there is an `ON DELETE CASCADE` on `users_opdivs.userid`, ignore.
2. `COALESCE(ARRAY_AGG(opdiv_id), '{}')` works cleanly today; `'{}'::integer[]` would be marginally more explicit if the constant ever moves next to a differently typed aggregate. Cosmetic.

Ship it 🚀

tarratsco (approved):

PR #558, opened 2026-08-19
Surface the migration startup fatal in deploy output instead of a bare Terraform timeout

voidspooks:
When a backend image starts against a database whose migration version is ahead of the image's registry, `migrations.Run()` hits tern's `BadVersionError` and `log.Fatal`s into CloudWatch, the task crash-loops, and the only symptom in Actions is the tfSTABLE timeout after 20 minutes, the same string a contention timeout produces. Grepping a failing run for the migration error finds nothing because the fatal never reaches the job output (ztmf-misc#298).

Three changes, per the ticket's two candidate shapes:

1. `migrations.Run()` now classifies tern's `BadVersionError` and fatals with a greppable `MIGRATION_VERSION_MISMATCH` message naming the version gap and the remedy (rebase onto latest main). Only this error class is reachable in a deployed environment since `Migrate()` always targets the registry length; other errors pass through unchanged.
2. `infrastructure.yml` gains a `Deploy failure diagnostics` step (`if: failure()`) that prints the ECS service events, stopped-task reasons and exit codes, and any `MIGRATION_VERSION_MISMATCH` lines from the last 30 minutes of the log group. This repo is public, so raw application logs are never dumped into job output; anything beyond the controlled token points at CloudWatch by name. A crash-looped startup fatal and a genuine contention timeout now look different in the job output.
3. `orchestration-dev.yml` fails fast, in the existing Check-for-Changes job, when main carries migration files the branch lacks, naming the mismatch and the remedy in seconds instead of after a 20-minute burn. This is a git approximation: it cannot see a migration another open PR deployed to dev without merging, which is the case changes 1 and 2 cover.

Gate: `go build`, unit, integration, and emberfall e2e all green locally. The fail-fast check was simulated from the parent of the newest migration-bearing commit (flags the missing migration and fails) and from HEAD (passes clean). No API contract change, so no `openapi.yaml` or `emberfall_tests.yml` changes.

### DIFF
--- a/.github/workflows/infrastructure.yml
+++ b/.github/workflows/infrastructure.yml
@@ -47,3 +47,47 @@ jobs:
 
       - name: Terraform apply
         run: terraform -chdir=infrastructure apply -no-color -var-file="tfvars/${ENV,,}.tfvars" -auto-approve
+
+      # On a failed apply (typically the 20m ECS steady-state timeout) pull what
+      # Terraform cannot show: why tasks stopped and whether the app named a
+      # known fatal. A migration-version crash loop prints its
+      # MIGRATION_VERSION_MISMATCH line here; genuine deploy contention shows no
+      # crash-looping stopped tasks. This repo is public, so raw application
+      # logs are never dumped into job output: only the controlled mismatch
+      # token is printed, everything else points at CloudWatch by name.
+      # Context: ztmf-misc#298.
+      - name: Deploy failure diagnostics
+        if: failure()
+        run: |
+          set +e
+          SFX=""; USFX=""
+          if [ "${ENV,,}" = "impl" ]; then SFX="-impl"; USFX="_impl"; fi
+          CLUSTER="ztmf${SFX}"
+          SERVICE="ztmf-api${SFX}"
+          LOG_GROUP="ztmf_api${USFX}"
+
+          echo "::group::ECS service events ($CLUSTER/$SERVICE)"
+          aws ecs describe-services --cluster "$CLUSTER" --services "$SERVICE" \
+            --query 'services[0].events[:15].[createdAt,message]' --output text
+          echo "::endgroup::"
+
+          echo "::group::Stopped task reasons"
+          TASKS=$(aws ecs list-tasks --cluster "$CLUSTER" --service-name "$SERVICE" \
+            --desired-status STOPPED --query 'taskArns[:10]' --output text)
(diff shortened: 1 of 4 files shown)


jsos3-cms (approved):
Verified the parts of this that carry the most risk, all clean:

The suffix derivation in the diagnostics step matches the terraform locals it mirrors, for every environment the workflow can be called with, and the log group it reads is the one the task actually writes to.

The steady-state wait is the clock on the failure this is meant to explain, so the log window is comfortably wider than what it has to catch rather than marginal.

The error classification is provably narrow rather than just narrow by convention: tern's migrate target is the registry length, so the only reachable case is the one the message describes, and the error is returned unwrapped so the match holds. `go vet` is clean, and I mutation-tested the new test in both directions - dropping the classification, and applying it to every error - and each fails on exactly the assertion that should catch it.

Two non-blocking notes:

1. The fail-fast check runs on every non-draft PR, but the image is only rebuilt when the backend job runs. A PR that does not rebuild the image cannot hit the mismatch, so gating the step on the existing backend-changes output would avoid asking for a rebase that changes nothing. Only one of the last twelve merges had that shape, so this is a judgment call rather than a defect - a cheap forced rebase is defensible too.

2. `grep -v '_test'` matches the substring anywhere in the path. It is load-bearing today, since a numbered integration test would otherwise trip the check, but `grep -v '_test\.go$'` is the same length and cannot swallow a migration whose name happens to contain `_test`.

Approving on the code.

jsos3-cms:
Separate from this PR, so it does not get read as a change request: the red snyk gate here is a Go standard-library advisory, not anything in this diff. `backend/go.mod` pins 1.25.11 and the advisory is fixed in 1.25.13.

That matters for merge order rather than for the review. The backend and infrastructure jobs are both gated on `!failure()`, so they skip while that scan is red, which means this PR cannot reach the dev deploy the ruleset requires until the toolchain moves. A one-line bump on its own branch would clear it for the whole backend queue rather than just for this PR.

jsos3-cms (approved):
Re-approving after the rebase onto the toolchain bump. Confirmed before re-approving that the change itself is untouched: the diff of this branch against its base is byte-identical to the version reviewed earlier, so the only difference is the rebase.

Worth flagging for the rest of the queue, since it will recur: the Update branch rebase replayed the commits without signatures, and this branch's commit came back as unsigned. Since main requires signatures, the PR was unmergeable until the commits were re-signed. The branch has been force-pushed with the same content, signed, and the signature now verifies.

PR #562, opened 2026-08-20
Bump excelize to v2.11.0 (Dependabot alert 44)

danielbowne:
Bumps `github.com/xuri/excelize/v2` from v2.9.0 to v2.11.0, resolving Dependabot alert 44 (GHSA-h69g-9hx6-f3v4, high): unbounded row-index allocation in the worksheet parser that can OOM or panic when opening a crafted xlsx.

The vulnerable code path is not reachable in this codebase - our only excelize usage is the write-only data-call export (`NewFile()` + `SetCellValue` in `backend/cmd/api/internal/spreadsheet/spreadsheet.go`); we never parse untrusted spreadsheets. This is hygiene to close the alert, not an urgent patch.

Both APIs we call are signature-identical between v2.9.0 and v2.11.0; the documented breaking changes in this range are chart/shape APIs we don't touch. Transitive upgrades (mscfb, msoleps, efp, nfp, x/crypto, x/net, and the mohae/deepcopy to tiendc/go-deepcopy swap) are excelize's own upstream dependency changes. The `go 1.25.14` directive already satisfies v2.11.0's Go 1.25 floor, so no toolchain or Dockerfile change is needed.

## Testing

- `go build ./...` clean
- `make test-unit` green
- `go mod tidy -diff` produces no diff (go.mod/go.sum are canonical)
- `go vet` on the spreadsheet package clean
- No OpenAPI, migration, or emberfall changes required: dependency-only bump with no request/response surface change


voidspooks (approved):

jsos3-cms (approved):
Confirmed this lands the same dependency version as the Dependabot PR it replaces: both move excelize to v2.11.0, and the diff here is limited to go.mod and go.sum.

The reason this one is worth taking instead is that it can actually prove itself. Every check is green, including the build and smoke test and the integration suite, none of which the Dependabot run could reach because its Snyk jobs failed on authentication before the pipeline got that far.

Closing #561 in favor of this.

PR #566, opened 2026-08-20
feat(#564): paginate GET /events with stable ordering and time-range filters

### LINKED ISSUE #564: Backend: ordering, pagination, and time-range filtering on GET /events
Part of epic CMS-Enterprise/ztmf-ui#183 (backend half).

`GET /events` exists and is restricted to unscoped admins, but `FindEvents` returns the full unordered result set: no ORDER BY, no limit, no pagination, no time-range filter. At roughly 200k rows in prod that is the whole table in arbitrary order, which no page can sit on.

Scope: order results by `createdat` DESC, add a limit with a cursor (a createdat-based cursor rides the existing `events_resource_createdat_idx` and `events_user_activity_idx` indexes from migrations 0037 and 0054, so no new index should be needed), and add from/to time-range query params. Regenerate the OpenAPI spec (`make generate-openapi`, never hand-edit) and update `emberfall_tests.yml` for the changed contract in the same branch.

Blocks the epic's frontend tickets (CMS-Enterprise/ztmf-ui#711): the table consumes this contract.

voidspooks:
Closes #564. First slice of the admin events page epic (CMS-Enterprise/ztmf-ui#183).

`GET /events` returned the entire events table unordered: no ORDER BY, no limit, no time filter, roughly 200k rows in prod. This makes it a paginated contract: results order by `createdat` DESC with `eventid` as tiebreaker, `limit` (default 50, capped at 500) and `offset` page through, and `from`/`to` bound `createdat` inclusively as RFC3339. The response becomes a page object, `{events, total, limit, offset}`, echoing the values actually applied so a client pager never re-derives the defaulting rules. No frontend consumer exists yet, so nothing breaks; the page in CMS-Enterprise/ztmf-ui#711 builds on this contract.

Two deviations from the ticket text, worked out in scoping. Offset pagination instead of a cursor, because the consumer is MUI DataGrid server-side paging, which wants page and pageSize with arbitrary jumps. And a migration (0058) the ticket didn't call for: the events table has no primary key and bulk writers stamp identical createdat values, so createdat alone is not a total order and tied rows could swap across page boundaries; 0058 adds an `eventid` identity column as tiebreaker plus a `(createdat DESC, eventid DESC)` index, since the 0037/0054 indexes both need their leading column in the predicate and cannot serve the unfiltered scan. Event rows now expose `eventid`, which the UI can use as a row key.

Also folded in: `sanitizeErr` now maps `ErrInvalidQueryParam` to 400 when a handler returns it directly. Previously only the schema-decode branch produced it; the new from/to parse path returns it, and unmapped it fell through to 500.

Gate: `make test-unit`, `make test-integration`, and `make test-e2e` all green locally (emberfall 217 ran, 0 failed, including seven new /events cases covering the page shape, echo and clamp behavior, and 400s for bad params). New unit tests pin the limit defaulting and clamping and the inverted-range 400; a new integration test writes tied timestamps and proves paging covers every row exactly once, with the range bounds inclusive. `openapi.yaml` regenerated via `make generate-openapi`; redocly lint is clean with the same 8 pre-existing warnings as main.

### DIFF
--- a/backend/cmd/api/internal/controller/controller.go
+++ b/backend/cmd/api/internal/controller/controller.go
@@ -144,6 +144,7 @@ func sanitizeErr(err error) (int, string, error) {
 		code = auth.CodeDelegateRequiresAdmin
 	case errors.Is(err, model.ErrNotUnique),
 		errors.Is(err, ErrMalformed),
+		errors.Is(err, ErrInvalidQueryParam),
 		errors.Is(err, model.ErrNotesTooLong),
 		errors.Is(err, model.ErrNoReference):
 		status = 400
--- a/backend/cmd/api/internal/controller/events.go
+++ b/backend/cmd/api/internal/controller/events.go
@@ -70,7 +70,8 @@ func RecordQuestionView(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNoContent)
 }
 
-//	@Summary	List audit-trail events
+//	@Summary		List audit-trail events, newest first, one page at a time
+//	@Description	Returns one page of the audit trail ordered by createdat descending (ties broken by eventid, so paging is stable). The response echoes the limit and offset actually applied and carries the total count matching the filters, for page math.
 //	@Tags		events
 //	@Produce	json
 //	@Security	bearerAuth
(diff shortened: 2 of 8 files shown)


jsos3-cms (approved):
Nice work on this. The ordering tiebreaker and the 500 cap are the right bones, and the migration comment explaining why `eventid` needs to exist will save the next person a lot of guessing. Worth saying too that this closes a real hole, since the endpoint was previously handing back the whole table.

Two things I want to note rather than raise, neither blocking.

The offset paging will drift a bit under live writes. Events get written on more or less every admin action, so a row can shift between someone loading page 1 and clicking through to page 2. You already flagged this as a deliberate trade-off for DataGrid, so mostly I am agreeing out loud. If we ever wanted to tighten it up, we could consider a keyset cursor on `(createdat, eventid)` behind the same response shape, but that feels like its own conversation rather than something to bolt onto this.

The other is the migration. Adding an identity column means Postgres rewrites the table instead of taking the cheap constant-default path, so it holds a brief exclusive lock and audit writes stall while it runs. At the table's current size that is likely seconds and nobody notices, and 0037 set the same precedent, so I am not asking for a change here. If we wanted to de-risk it before the table gets much bigger, we could consider splitting the column add from the index create, or seeing whether the index can be built outside the migration transaction.

Approving.

PR #568, opened 2026-08-21
ci: require migration-touching PRs to be current with main (misc#303)

voidspooks:
Requires PRs that touch the migrations package to be current with the head of main, and fixes the missing-migration fail-fast from #558 so it compares against the PR head instead of the merge-ref checkout.

Migrations are globally ordered state. A migration authored against a stale schema can collide on version number at merge, and an edited migration or schema change merged behind an open branch never shows up as a missing file, so the existing check could not see it. The new step fails the Check for Changes job with a rebase instruction whenever the PR diff touches backend/cmd/api/internal/migrations and the branch is behind main. PRs that do not touch migrations are unaffected.

The #558 fix: on pull_request events the default checkout is refs/pull/N/merge, whose main side diffs empty by construction, so the missing-migration step could only fire if main advanced in the seconds between the merge ref being computed and the job's fetch. Diffing from github.event.pull_request.head.sha (the merge commit's second parent, already present in the full-history checkout) makes it do what its comment says. Verified both behaviors in a scratch repo: the head-sha diff reports main's migration missing from a stale branch where the merge-ref diff returns nothing.

Known limit, recorded not solved: these steps run on PR events only, so a migration merging to main does not re-trigger open PRs. Making the rule binding at merge time is a branch-protection question (require branches up to date), not a workflow one.

Gate: unit, integration, and emberfall e2e suites pass locally; actionlint clean on the changed steps (one pre-existing SC2086 info on the untouched backend-gating step).

Closes CMS-Enterprise/ztmf-misc#303.

### DIFF
--- a/.github/workflows/orchestration-dev.yml
+++ b/.github/workflows/orchestration-dev.yml
@@ -49,9 +49,14 @@ jobs:
       # This is a git approximation: it cannot see a migration another open PR
       # deployed to dev without merging; that case is caught by the
       # deploy-failure diagnostics in infrastructure.yml. Context: ztmf-misc#298.
+      #
+      # The diff must run from the PR head, not HEAD: on pull_request events the
+      # checkout is refs/pull/N/merge, whose main-side diff is empty by
+      # construction. The head SHA is the merge commit's second parent, so the
+      # full-history checkout already has it.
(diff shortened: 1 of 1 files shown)


tarratsco (approved):
LGTM, approving

PR #569, opened 2026-08-24
feat(#565): resolve the initiating user to display name and email in the events response

### LINKED ISSUE #565: Backend: resolve the initiating user to display name and email in the events response
Part of epic CMS-Enterprise/ztmf-ui#183 (backend half).

Event rows carry only the initiating user's `userid` uuid. An admin reading the events page needs a name, so the events response should resolve the initiating user to display name and email server-side rather than making the client join against the users list.

Scope: include user display fields in the `GET /events` response (a join to users, shaped like the existing patterns in the model layer). Decide and document the rendering for soft-deleted users, who can still own historical events. OpenAPI regen and emberfall updates ride along with the contract change; coordinate with #564 so the contract only changes once if the two land together.

voidspooks:
Closes #565. Part of the admin events page epic (CMS-Enterprise/ztmf-ui#183, backend half).

Event rows carried only the initiating user's uuid, forcing the client to join against the users list to show who acted. GET /events now resolves the user server-side: each row carries userfullname, useremail, and userdeleted, populated by an inner join to users in FindEvents. The join cannot drop rows because events.userid is NOT NULL with an FK to users (migration 0006), which is also why the count query stays join-free and Total cannot drift from the page.

Soft-deleted users still own historical events, so the contract is: their identity is returned as it stands today, with userdeleted true as the client's cue to mark a retired account. Rendering that state is the frontend's call.

The change is additive; nothing in the existing response moves. openapi.yaml is regenerated (model.Event becomes model.EventWithUser in the page schema) and emberfall gains a deterministic assertion using the seeded archives service account, which is the only writer of action=imported. Integration coverage asserts identity resolution on every paged row and the soft-deleted case end to end.

Gate: make test-unit, make test-integration, and make test-e2e all pass locally.


danielbowne (approved):
Looks good. Checked the join can't drop rows (events.userid NOT NULL + FK), that users.email/fullname/deleted are all NOT NULL so the plain string/bool scans are safe on prod data, and that the only shared column name after the join (userid) is qualified. CI is green.

One optional nit, your call whether to fix now or just merge: the where closure and ORDER BY still reference createdat, action, resource, and payload unqualified. Nothing collides today since users has none of those, but qualifying them as events.createdat etc. would keep this query from breaking if a column with one of those names is ever added to users. Happy to approve as-is.

PR #559, opened 2026-08-19
chore(ci): bump the github-actions group with 9 updates

jsos3-cms:
Rehomes #556 onto an in-repo branch, with the original commit's authorship preserved.

Dependabot-triggered runs read the Dependabot secret store rather than the repo and environment secrets, so `SNYK_TOKEN` was empty on #556 and all three Snyk jobs failed with a 401 rather than on anything in the diff. That PR could never go green, and so could never satisfy the required dev deployment on main. Pushing the same commit from an in-repo branch gives those jobs real credentials. #556 is closed in favor of this one.

The bump itself is unchanged: nine action updates across seven workflows, moving `actions/checkout` and `actions/setup-go` onto majors that run on Node 24. That part is not cosmetic - current runs already warn that `actions/checkout@v4` and `actions/setup-go@v5` are being forced onto Node 24 because Node 20 is deprecated.

The second commit corrects two provenance comments. `snyk/actions` and `aquia-inc/emberfall` are pinned by SHA because neither publishes tags the workflow can track, and each carries a trailing comment naming the date of the pinned commit. Both SHAs moved in this bump while the comments still named the old date, so each described a commit that is no longer the one pinned. They now read 2026-08-03 and 2026-08-12 respectively.

Verified before opening: the first commit's tree is identical to #556's head, and the branch as a whole differs from it only in those two comment lines. The residual mix of bare tags in `analysis.yml` and SHA pins elsewhere is untouched and remains the subject of #317.

Expect the Go scan to be red here for the same reason it is red on #558 - a Go standard-library advisory against the version pinned in `backend/go.mod`, unrelated to this diff.

PR #571, opened 2026-08-26
chore(#317): SHA-pin every workflow action, retire the snyk @master refs, and replace staticcheck-action

### LINKED ISSUE #317: Bump GitHub Actions workflows to Node.js 24 and pin third-party actions to commit SHAs

voidspooks:
Stacked on #559: this branch sits on top of it, so the diff here includes #559's two commits until it merges. Once it does I will rebase onto main so only the commit below remains, and mark this ready for review. Draft until then.

Finishes ztmf#317 on top of #559. That PR moved the remaining actions to their Node.js 24 majors; this one converts every ref it left as a floating tag to a commit SHA with a version comment, matching the format #349 established in backend.yml, and closes out the rest of the ticket's acceptance criteria. Also closes CMS-Enterprise/ztmf-misc#173.

The two snyk/actions refs in analysis.yml were still on @master. All three snyk/actions uses now pin the v1.0.0 tag (9adf32b, 2025-10-03); comparing that tag to master head shows no changes under setup/, golang/, or docker/, so this is a provenance change only, and it retires the stale "no tagged releases" comment. aquia-inc/emberfall moves from a main commit to the v0.5.0 tag, whose action.yml is byte-identical to the previous pin; the binary stays at 0.4.1 via the version input. docker/setup-buildx-action goes to v4.3.0 for parity across backend.yml and ops-image.yml. The two setup-go blocks in analysis.yml gain cache-dependency-path: backend/go.sum, which #349 only did for backend.yml, so the openapi and snyk test jobs stop missing the cache on every run.

The lint go job no longer uses dominikh/staticcheck-action. The composite pins actions/cache to a Node.js 20 build (upstream dominikh/staticcheck-action#22 and #23, master unchanged since March 2026), which was the one deprecation warning left after every other bump. The job now installs staticcheck with go install after a setup-go on Go stable, mirroring the composite's install-go default, since staticcheck's latest release requires a newer Go than the go.mod floor and setup-go v6+ sets GOTOOLCHAIN=local. The action's problem matcher is checked in under .github/matchers/staticcheck.json so PR annotations behave as before. The same commands run clean locally against backend/ on Go 1.26.5 with staticcheck 2026.2.1.

Verification: a script resolved every pinned SHA back to its tag through the GitHub API and confirmed each action's runtime is node24 (or composite/docker); no @vN, @master, or @main refs remain under .github/workflows. actionlint reports the same 16 pre-existing shellcheck notes as main and nothing new. ops-image.yml and nightly-impl-sync.yml have no pull_request trigger, so their pins are verified by the resolver rather than a run. .github/dependabot.yml already carries the github-actions group from #349 and is unchanged.

Closes #317

The red Check for Changes is the #568 migration-currency guard: this branch inherits #559's base, which is five commits behind main and predates 0058eventspagination.go. It clears with the rebase once #559 merges; as a draft this PR does not deploy to dev, so the guard's crash-loop scenario cannot occur here.

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -24,7 +24,7 @@ jobs:
       backend: ${{ steps.backend.outputs.backend }}
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
@@ -46,7 +46,7 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: LINT backend/
         uses: dominikh/staticcheck-action@v1
@@ -61,10 +61,10 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Setup Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v7
         with:
           go-version-file: backend/go.mod
 
@@ -100,10 +100,10 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: TFLINT setup
-        uses: terraform-linters/setup-tflint@v4
+        uses: terraform-linters/setup-tflint@v6
 
       - name: TFLINT version
         run: tflint --version
@@ -123,10 +123,10 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Setup Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v7
         with:
           go-version-file: backend/go.mod
 
@@ -146,7 +146,7 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: code test
         uses: snyk/actions/golang@master
@@ -163,7 +163,7 @@ jobs:
       contents: read
     steps:
       - name: Check out repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
       
       - name: iac test
         uses: snyk/actions/golang@master
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -64,7 +64,7 @@ jobs:
         run: docker buildx build --platform linux/amd64 --tag ${{ secrets.ECR_REPO_URL }}:${{ steps.revParse.outputs.shaShort }} --load ./backend 
 
       - name: Snyk - Test Image
-        uses: snyk/actions/docker@12140f4059e244892ae643824a95459a102120dd # snyk/actions has no tagged releases; pinned master @ 2026-06-16
+        uses: snyk/actions/docker@12140f4059e244892ae643824a95459a102120dd # snyk/actions has no tagged releases; pinned master @ 2026-08-03
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
@@ -101,7 +101,7 @@ jobs:
           WORKSPACE: ${{ github.workspace }}
 
       - name: Emberfall Smoke Tests
-        uses: aquia-inc/emberfall@a0d970e3235fc42dd7b24c1b7dd18f924fcb5aed # main @ 2026-06-16 (composite; binary version set via `version` below)
+        uses: aquia-inc/emberfall@a0d970e3235fc42dd7b24c1b7dd18f924fcb5aed # main @ 2026-08-12 (composite; binary version set via `version` below)
         with:
           version: 0.4.1
           file: ./backend/emberfall_tests.yml
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -14,10 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repo
-        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Set up Go
-        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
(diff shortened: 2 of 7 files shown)


jsos3-cms:
Rebased this branch onto `main` and force-pushed (`da2ce34` -> `9dcd94b`). `#559` merged to main earlier today as squash `1e20bb2`, which put this branch six commits behind and was tripping the new migration-currency gate; that gate is satisfied now and the PR reads `MERGEABLE` again.

One note on what the rebase did to the commit list, since it is visible on the PR. This branch was stacked directly on the rehomed `#559` commits rather than sitting beside them, so two of its three commits — the dependabot bump and the follow-up that corrected the two pin provenance dates — were already upstream by content once `1e20bb2` landed. Git dropped both as redundant (`patch contents already upstream`), leaving your single commit as the whole PR. Nothing of yours was changed: the diff `9dcd94b` applies to `main` is byte-identical to the diff `da2ce34` applied to its old base, and the commit is still authored by you.

The only conflict was in `backend.yml`, on the two trailing provenance-date comments for `snyk/actions/docker` and `aquia-inc/emberfall` — both sides carried the identical SHA and differed only in the date string. It resolved to main's corrected dates, which is moot in the end since your commit replaces both of those lines with the tagged `v1.0.0` and `v0.5.0` pins anyway. No `uses:` line lost its SHA-pinned form; every external action in `.github/workflows/` is now a 40-character SHA with a version comment, and there are no `@v4`/`@v5`/`@master` refs left.

A review finding to pass on rather than something I have applied — your call how to handle it. `go install honnef.co/go/tools/cmd/staticcheck@latest` in `analysis.yml` is still a floating ref, and after this PR it is the only unpinned reference left anywhere in the CI surface. That cuts against the PR's own thesis: `@latest` is exactly the mutable-upstream exposure the SHA pins exist to close, and it is arguably a sharper one than a floating action tag, since it resolves at job time to whatever the module proxy is serving and executes as part of the lint step. Worth pinning to `@2026.2.1`, the version the PR body says was tested locally. Ideally as its own commit so the pin is reviewable on its own.

Also flagging that the force-push dismissed any prior approval on this PR — the repo ruleset has `dismiss_stale_reviews_on_push: true`, so this will need a fresh review pass before it can go in.

voidspooks:
Thanks for the rebase. Confirmed the commit is unchanged: the patch-id of my commit before and after (da2ce34, 9dcd94b) is identical.

Staticcheck pin is in as its own commit, 23143fe. One note on the target: 2026.2.1 is the release name, the Go module tag is v0.8.1, so the install line is `go install honnef.co/go/tools/cmd/staticcheck@v0.8.1`, which resolves to `staticcheck 2026.2.1 (0.8.1)` and matches what the PR body says was tested locally. With that there are no floating refs left in .github/workflows. Marking ready for review.

jsos3-cms (approved):
LGTM! Nice work 👍

